PR #182, opened 2026-08-27
refactor(performance): keep timing cases canonical

paulfalgout:
Related #127

Follow-up to #181 and its late [CodeRabbit review](https://github.com/marionettejs/marionette/pull/181#discussion_r3873599903).

## What
- Remove the exported `caseFactory` override so hosted reports always use the pinned canonical timing cases.
- Preserve the case-construction cleanup regression through a temporary ESM runtime fixture that fails during the real `createCases` path.
- Regenerate the pinned timing-harness SHA-256.

## Why
The test seam added while addressing #181 feedback also allowed programmatic callers to substitute different workloads under canonical configured case IDs. This follow-up restores a single benchmark-definition path without weakening cleanup coverage.

## Scope
- Benchmark tooling, its focused timing test, and the pinned performance contract only.
- Timing case IDs, workloads, thresholds, and hosted-reporting-only policy are unchanged.
- Runtime source, package exports, dependencies, and built artifact bytes are unchanged.
- The separate suggestion to expose private allocation/retention shapes as a public runtime API is intentionally out of scope because those tests are repository-internal structural proxies and the Phase 0 contract requires zero runtime overhead.

## Deployment Impact
No application or form redeploy is required. No npm publication, Git tag, GitHub release, or website publication is involved.

## Testing
- `npm ci`
- `npm run test:performance` — 17/17 passed
- `npm run lint:ci`
- `npm run size` — 49,500 B / 51,975 B; production graphs 40/1/1
- `npm run performance:timing` — all seven canonical cases completed
- `npm run check:dist`
- Independent SHA-256 source/contract equality check
- `git diff --check`

---
## Summary by cubic
Removes the `caseFactory` override from `measure` so hosted benchmark reports always run the pinned canonical timing cases, and preserves cleanup coverage by failing inside the real `createCases` path through a temporary ESM runtime fixture. Regenerates the timing-harness SHA-256 to match.

The override seam allowed programmatic callers to substitute workloads under canonical case IDs; closing it restores a single benchmark-definition path without weakening cleanup coverage. Timing case IDs, workloads, thresholds, and the hosted-reporting-only policy are unchanged. No runtime, export, dependency, artifact, or rollout changes are involved.

<sup>Written for commit 77000606d5c7b9d10f833dcf4c70d83266db12cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/182?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/benchmarks/performance.mjs
+++ b/benchmarks/performance.mjs
@@ -212,7 +212,6 @@ function createCases({ Backbone, Marionette }) {
 export async function measure({
   root = '.',
   configPath = 'config/performance.json',
-  caseFactory = createCases,
   dependencyRoot = root,
 } = {}) {
   const resolvedRoot = resolve(root);
@@ -223,7 +222,7 @@ export async function measure({
   const results = [];
 
   try {
-    const cases = caseFactory(runtime);
+    const cases = createCases(runtime);
     for (const caseConfig of contract.timing.cases) {
       const run = cases.get(caseConfig.id);
       if (!run) {
--- a/config/performance.json
+++ b/config/performance.json
@@ -289,7 +289,7 @@
   },
   "timing": {
     "harnessSchemaVersion": 1,
-    "harnessRevision": "f22242e13a8af8ee11a57930a6913d77b2ab65f32fa109c917987cdfeadb9145",
+    "harnessRevision": "271128bee00c66b3168424d65220dbe6c5dbea1f336082c21b45a07a344bb1d9",
     "sampleCount": 20,
     "warmupBatches": 5,
     "cases": [
--- a/test/performance/timing.test.mjs
+++ b/test/performance/timing.test.mjs
@@ -1,5 +1,5 @@
 import assert from 'node:assert/strict';
-import { mkdtemp, readFile, rm, writeFile } from 'node:fs/promises';
+import { mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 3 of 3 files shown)


paulfalgout:
Reviewed at `f5443aee` (the SHA I fetched). **Approve — good to merge.**

Removing the `caseFactory` parameter is the right call. As written on `master`, any programmatic caller could pass their own factory and have arbitrary workloads execute under the canonical configured case IDs — so a timing report could carry the pinned IDs while measuring something else entirely. Since the whole point of #127 is that benchmark evidence be reproducible from a pinned harness, a seam that decouples the ID from the workload undermines the artifact rather than just the test. Collapsing it back to a single `createCases(runtime)` path is the correct fix, and moving the cleanup regression to a runtime fixture that fails inside the real path keeps the coverage without the seam.

**Verified the re-pin is honest:**

```
computed sha256(benchmarks/performance.mjs) : 271128bee00c66b3168424d65220dbe6c5dbea1f336082c21b45a07a344bb1d9
contract timing.harnessRevision             : 271128bee00c66b3168424d65220dbe6c5dbea1f336082c21b45a07a344bb1d9
```

(My first attempt at this check pulled the wrong `sha256` field out of the contract and reported a mismatch — `config/performance.json` has several. The value above reads `timing.harnessRevision` specifically.)

**I also checked whether this trips the base-authority gate**, since this is the first PR merging against a `master` that now carries a performance contract, and `master` still pins the previous harness revision `f22242e1…`. It doesn't, and the reason is a deliberate separation worth naming: `harnessRevision` isn't referenced anywhere in `config/bundle-size.mjs`, so it never enters the blocking size/graph path that runs under the *base's* contract. It's asserted in `test/performance/timing.test.mjs` against the PR's own contract instead.

That split is the right one. Size and module-graph limits are base-authoritative because a PR must not be able to relax them; the harness revision is self-consistent because a PR legitimately needs to change the harness and re-pin it in the same change. Had `harnessRevision` been in the blocking path, every intentional harness edit would have been unmergeable — worth knowing the boundary was drawn deliberately rather than by luck.

Timing case IDs, workloads, thresholds, and the reporting-only policy are untouched, and the change is confined to benchmark tooling plus its contract pin — no runtime source, no artifacts.

One thing to glance at before merging: `Release package smoke (Windows x64)` was still pending when I looked; everything else was green.

paulfalgout:
Exact-head Claude review and the external Claude monitor both report no blockers / good to merge. Follow-up spot checks confirm no `caseFactory` references remain repo-wide, the pinned SHA-256 is exactly 64 characters and byte-equal to the harness source digest, and the regression retains both descriptor assertions plus temp-directory cleanup. No further code changes were needed. Cubic remains the final pending gate.

paulfalgout:
Updated onto `master` after #183. The benchmark/config/test-only PR now produces the required `Build and validate` check, which is the definitive regression proof that the ruleset/workflow deadlock is fixed. Exact head is `7700060`; all checks and review gates have reset.

paulfalgout:
Final exact-head composition check for 7700060:\n\n- PR diff SHA-256 before updating from master: 1ca3ba647c567cbbcfb92ec4422145df595e7615187464a2b8ef2e2f1e3f19fd\n- PR diff SHA-256 at the current head: 1ca3ba647c567cbbcfb92ec4422145df595e7615187464a2b8ef2e2f1e3f19fd\n- The hashes are identical; f5443ae..7700060 contains only the already-reviewed #183 workflow change.\n- Exact-head CI, Documentation, Node current, and release-promotion dry run all passed; publication jobs skipped.\n- Cubic and Greptile are clean, and there are zero unresolved threads.\n\nMerging normally with --match-head-commit 7700060; this protects the reviewed PR head while GitHub creates the normal merge commit on master.

Issue #174, opened 2026-08-27
Phase 0: decide and enforce the default-branch merge policy


paulfalgout:
## Goal

Choose and enforce an explicit default-branch merge policy for the agent-ready v5 repository.

## Dependencies

Builds on the GitHub Actions migration in #66 and the CODEOWNERS alignment in #172.

## Agent-development failure addressed

Agents can now rely on live repository enforcement instead of an informal convention for first-party CI freshness and review-thread completion.

## Public behavior contract

Repository governance only. This issue does not change package or runtime behavior.

## Runtime cost category

External repository settings and documentation only; zero production cost.

## Chosen policy

- Pull requests are required for `master`.
- Human approval count is zero and CODEOWNER approval is not required. This preserves the solo-maintainer, agent-driven workflow without manufacturing self-approval.
- Review conversations must be resolved.
- Required branches must be current with `master`.
- The six required first-party checks are:
  - `Node 24`
  - `Package smoke (macOS arm64)`
  - `Package smoke (Windows x64)`
  - `Bundle size`
  - `Build and validate`
  - `Analyze (javascript-typescript)`
- Deletion and non-fast-forward updates are blocked.
- Repository administrators retain an emergency bypass. Bypass is not a normal merge path and must be recorded in a public issue with the reason, exact action, affected pull request or commit, and restoration evidence.
- Claude remains an exact-head external-review convention rather than a native required check because the comment monitor does not expose a stable status-check context. Do not merge until the public review evidence identifies the current head SHA and reports approval or no blocker. Other third-party reviewers remain advisory unless a future decision promotes a stable, universally emitted check.

## Live enforcement

GitHub ruleset [21649624](https://github.com/marionettejs/marionette/settings/rules/21649624), `Default branch merge policy`, is active for `~DEFAULT_BRANCH` and enforces the policy above. The obsolete classic branch rule and disabled duplicate ruleset were removed, leaving one authoritative policy.

## Controlled pull-request evidence

PR #178 exercised both sides of the gate:

- GitHub reported `BLOCKED` while required exact-head checks were pending and while review threads were unresolved.
- After the branch was current, all six required checks passed, and every thread was resolved, GitHub reported `CLEAN` and `MERGEABLE`.
- The exact reviewed head `05a7e7b8fad3dcd033582070ec9a79a9fafa82cb` then merged normally as `988833151bcafc96f5048a5506940582b5c5a4b1` without a policy bypass.

## External-review and outage recovery

[...]

PR #172, opened 2026-08-27
docs(governance): align documentation code owners

paulfalgout:
## Linked issue

- Related #153

## What

- Add `@marionettejs/marionette-core` to all six existing documentation, documentation-tooling, diagnostic-catalog, and documentation-workflow ownership rules.
- Retain `@paulfalgout` on every existing rule.

## Why

The documentation publication contract names the Marionette core team as the repository owner. The current CODEOWNERS file names only one individual, so review ownership does not match that contract.

## Agent-development failure addressed

- Documentation and diagnostic changes now expose the durable team-level review owner directly in repository metadata instead of requiring agents to infer an individual maintainer.

## Public behavior contract

- Canonical behavior after this PR: matching pull requests request review from both the existing individual owner and `@marionettejs/marionette-core`.
- Superseded behavior removed, if any: individual-only ownership for these paths.
- Compatibility exception and removal condition, if any: N/A; retaining the individual owner provides explicit additional coverage and does not preserve obsolete behavior.

## Runtime cost

- [x] Static or documentation only
- [ ] Development or test only; excluded from production entrypoints
- [ ] Changes an existing production path
- [ ] Adds an explicitly opt-in runtime path

Measured impact or explanation:

- Bundle: 0 B across every reported bundle in CI.
- Startup/hot path: none; CODEOWNERS is GitHub metadata.
- Allocations/retention: none.

## Scope

- Changes only `.github/CODEOWNERS`.
- Does not change documentation content, diagnostics, workflows, Pages settings, DNS, HTTPS, runtime code, package contents, or generated artifacts.

## Deployment Impact

No application, package, documentation, or form deployment is required.

## Validation

- [x] Tests cover the acceptance criteria and relevant edge cases.
- [x] Docs, examples, declarations, diagnostics, and rule metadata agree where relevant.
- [x] No private consumer, private fixture, or undocumented context is required.
- [x] I ran the commands listed below and am reporting their actual results.

Commands:

```sh
git diff --check
# passed

# Validated all six non-comment rules contain one path and both expected owner tokens.
# Verified @marionettejs/marionette-core exists and has repository admin permission through the GitHub API.
```

CI also passed Node 24, Node 26 advisory, macOS arm64 and Windows x64 package smoke, CodeQL, and bundle-size checks at exact head `9ac5cbcd6dd95c045a3ffe4bd99018e5c45bcaed`.

## Package and browser impact

- Entry points or install fixtures affected: none.
- Browser contracts affected: none.
[...]

PR #183, opened 2026-08-27
fix(ci): run required docs check for every PR

paulfalgout:
## What
- Run the `Build and validate` documentation job for every pull request.
- Keep the existing path filter for `master` pushes, so documentation publishing work is not broadened.

## Why
The default-branch ruleset requires `Build and validate`, but the workflow previously existed only for pull requests touching documentation-related paths. Non-documentation PRs therefore could pass every available check while remaining permanently blocked because the required check was never created.

## Scope
- Documentation workflow trigger behavior only.
- Documentation build commands, deployment guards, runtime source, package artifacts, and release behavior are unchanged.
- Website deployment remains disabled unless the existing `DOCS_PAGES_ENABLED` variable and `master` push conditions are satisfied.

## Deployment Impact
No application, package, form, or website deployment is performed by this PR. Pull requests will run one additional required documentation validation job.

## Testing
- Parsed `.github/workflows/docs-pages.yml` with Ruby YAML
- `npm ci`
- `npm run docs:check` — 29 documentation pages, 30 HTML files, 288 internal links
- `git diff --check`
- The PR itself must produce the required `Build and validate` check

---
## Summary by cubic
Runs the required `Build and validate` docs check for every pull request. Previously, the check only ran when PRs touched documentation-related paths, so non-documentation PRs could fail to meet the branch ruleset because the required check never ran. The path filter is removed only for `pull_request` events; `master` pushes keep the existing filter, so deployment behavior is unchanged.

<sup>Written for commit 9359bbe6b6de78d18f431a539e615ad3638c1a68. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/docs-pages.yml
+++ b/.github/workflows/docs-pages.yml
@@ -18,19 +18,6 @@ on:
       - 'package-lock.json'
       - 'package.json'
   pull_request:
-    paths:
-      - '.github/workflows/**'
-      - '.nvmrc'
-      - 'config/check-release-profile.mjs'
-      - 'config/check-diagnostic-catalog.mjs'
-      - 'config/diagnostics/**'
-      - 'config/docs/**'
-      - 'config/release-profile.json'
-      - 'docs/**'
-      - 'docs-site/**'
-      - 'upgradeGuide.md'
-      - 'package-lock.json'
-      - 'package.json'
   workflow_dispatch:
 
 permissions:


paulfalgout:
Bounded Claude review found no blockers. It correctly noted that this workflow-file PR matched the old filter, so its own check is not the general regression proof. The operational proof will be PR #182 after #183 merges: #182 changes only benchmark/config/test files, is currently blocked solely because `Build and validate` is absent, and will be updated onto the fixed base before merge. The full workflow was rechecked: deploy still requires `master`, a non-PR event, and `DOCS_PAGES_ENABLED == true`. Running docs validation on every PR is the intentional ~12-second CI cost of making the required check reliable.

paulfalgout:
Reviewed at `9359bbe6` (the SHA I fetched). **Approve — merge this first.** The bug it fixes is live right now, and I can point at the PR it's blocking.

The "Default branch merge policy" ruleset is active and requires six checks, `Build and validate` among them:

```
Node 24 · Package smoke (macOS arm64) · Package smoke (Windows x64)
Bundle size · Build and validate · Analyze (javascript-typescript)
```

But `Build and validate` lives in `docs-pages.yml`, whose `pull_request` trigger was path-filtered to documentation paths. A PR touching none of them never creates the check, and GitHub can't satisfy a required check that was never reported.

**This is not hypothetical — #182 is stuck in it:**

```
#182  Build and validate: ABSENT — required check never created
      mergeable=MERGEABLE  state=BLOCKED

#183  Build and validate: SUCCESS
      mergeable=MERGEABLE  state=CLEAN
```

#182 touches `benchmarks/`, `config/performance.json`, and `test/performance/` — none on the filter list — so it is green on everything that ran and still unmergeable. #183 only reports the check because it happens to touch `.github/workflows/**`, which is on the list. That's also why this PR can merge itself out of the problem.

The fix is the right shape: drop the filter on `pull_request` so the required check always reports, keep it on `push` to `master` so documentation publishing isn't broadened. Deployment stays behind `DOCS_PAGES_ENABLED` and the `master` push conditions, untouched.

### One correction to my own earlier review

In my [#172 review](https://github.com/marionettejs/marionette/pull/172#issuecomment-5438179692) I said `master` had no required-review protection, based on the legacy branch-protection endpoint returning null. That was accurate then but is now **out of date** — #174 landed as a *ruleset*, which that endpoint doesn't report. Rulesets are queried separately, and this one is active.

Worth noting what it does and doesn't do, since it changes the picture from my #178 note: it requires six status checks but sets `required_approving_review_count=0`. So merges are now gated on CI but still not on human review — meaning flipping `publicationEnabled` to `true` remains a self-mergeable change, just one that has to be green first. If the intent was that the final publication authorization gets a second pair of eyes, that still wants a nonzero approval count.

Issue #131, opened 2026-08-26
v5: finish lifecycle contract and transition coverage mapping


paulfalgout:
## Current status — September 9, 2026

Repository readiness work is merged in [#500](https://github.com/marionettejs/marionette/pull/500) at [master `70eaeaf5`](https://github.com/marionettejs/marionette/commit/70eaeaf582a3ae062b5ff48650cfda12cfa3bf9f). It is available in source and verified tarballs; no new npm version or tag was published. The registry remains beta.1 on `latest`.

The lifecycle inventory, public test anchors and bounded Application command model are integrated and pass the exact-head candidate checks. Future behavior changes require their own concrete contract; this bounded mapping work is complete.

## Goal
Close the documentation-to-test mapping for the accepted View, Region, CollectionView, Behavior and Application lifecycle contracts. Prefer documentation and targeted tests over additional runtime state or guards.

## Delivered
- [x] Public View state vectors and operation/repeated-call tables in [docs/view.lifecycle.md](https://github.com/marionettejs/marionette/blob/8ea3708841e5ccb9a90f98d078cc4ca53c04ea1d/docs/view.lifecycle.md).
- [x] Application operation, readiness, invalidation and child propagation tables in [docs/marionette.application.md](https://github.com/marionettejs/marionette/blob/8ea3708841e5ccb9a90f98d078cc4ca53c04ea1d/docs/marionette.application.md).
- [x] Dedicated View, Region, CollectionView, Behavior and Application lifecycle suites, including supported construction/teardown and selected reentrant regressions. Historical constructor-recovery work is superseded by the supported synchronous failure boundary in #470.

## Local implementation awaiting integration
- [x] Map accepted lifecycle semantics to exact canonical documentation headings and public tests in `config/api-contracts/semantics.json` and the generated inventory.
- [x] Record signatures, results, timing, ownership, mutation, repeated calls, destruction boundaries and diagnostics; fail checks on source/test/document drift.
- [x] Add bounded Application command-model coverage for held readiness, shared starts, stop/restart/destroy supersession, late settlement and child ownership. Existing explicit tests remain valid evidence.

## Remaining acceptance
- [x] Integrate the local inventory, model and CI checks, and link their public commit/PR evidence.
- [x] Recheck the mapping and applicable public tests at the integration/release commit; add code only for a concrete supported-contract defect.

[...]

PR #195, opened 2026-08-28
test(lifecycle): specify Region state transitions

paulfalgout:
Related #131

## What
- Document the public Region lifecycle states and operation transitions.
- Define repeated show, detach, empty, replacement, external View destruction, and Region destruction behavior.
- Add six executable transition-contract tests covering state, ordering, idempotence, swapping, replacement, and teardown.

## Why
Agents currently have to infer Region lifecycle behavior from distributed implementation details and individual tests. This gives issue #131 a compact public transition table backed by one focused executable contract.

## Scope
- Documentation and tests for Region lifecycle behavior.
- No runtime, public API, Application lifecycle, package, release, or website publication changes.
- Post-destroy operations other than repeated destroy remain intentionally outside this slice.

## Deployment Impact
No application, package, documentation-site, or form redeploy is needed. This PR does not publish the website.

## Testing
- npx vitest run test/unit/region.spec.js test/unit/region-lifecycle.spec.js test/unit/region-el-validation.spec.js --coverage.enabled=false (128 passed)
- npx vitest run test/unit/region-lifecycle.spec.js --coverage.enabled=false (6 passed after final edits)
- npm run lint:ci
- npm run docs:check (29 pages, 30 HTML files, 288 internal links)
- git diff --check

---
## Summary by cubic
Documents the Region lifecycle contract and backs it with executable tests, so agents no longer have to infer state transitions from scattered implementation details. Addresses the documentation gap from #131.

- Adds a lifecycle transition table and operation semantics for `show`, `detachView`, `empty`, external View destruction, missing-element handling, and `destroy` to `docs/marionette.region.md`.
- Adds seven contract tests in `test/unit/region-lifecycle.spec.js` covering state ordering, idempotence, swapping, replacement, missing-element handling, and teardown.
- No runtime, API, package, or release changes. Post-destroy operations other than repeated `destroy()` remain unsupported.

<sup>Written for commit 2736849441aff603e48c630ba1e53fc612d84df4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/195?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added a lifecycle guide for regions, covering empty, occupied, detached, and destroyed states.
  * Documented supported operations, view detachment responsibilities, missing-element handling, and temporary view-swapping behavior.

* **Tests**
  * Added coverage for region state transitions, repeated operations, view replacement, event ordering, and external view destruction.

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -84,8 +84,7 @@ change which lifecycle operations are valid.
 
 Successful `show`, `empty`, and `destroy` calls return the Region when their
 operation completes. With `allowMissingEl: true`, `show` instead returns `undefined`
-and leaves the Region empty when its element does not resolve; `empty` returns the
-Region without changing unresolved DOM contents. A View returned
+and leaves the Region empty when its element does not resolve. A View returned
 by `detachView()` remains the caller's responsibility until another Region shows it
 or it is destroyed. Operations other than repeated `destroy()` after Region
 destruction are unsupported; this contract does not make a destroyed Region reusable.
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -16,6 +16,7 @@ Regions maintain the [View's lifecycle](./view.lifecycle.md) while showing or em
 ## Documentation Index
 
 * [Instantiating a Region](#instantiating-a-region)
+* [Lifecycle transition contract](#lifecycle-transition-contract)
 * [Defining the Application Region](#defining-the-application-region)
 * [Defining Regions](#defining-regions)
   * [String Selector](#string-selector)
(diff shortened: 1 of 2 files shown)


paulfalgout:
Reviewed at `e8b33a66` (the SHA I fetched). **Approve.** Docs and tests only, and the tests pin the parts of Region behavior that genuinely can't be inferred by reading the source.

**The destroy-ordering test is the most valuable thing here**, because it captures `isDestroyed()` *at each event* rather than only at the end:

```
region:before:destroy:false
region:before:empty:true
view:before:destroy
view:destroy
region:empty:true
region:destroy:true
```

I checked that against the implementation, and it follows from `destroy()` setting the flag between the two triggers:

```js
this.triggerMethod('before:destroy', this, options);   // false
this._isDestroyed = true;
this.reset(options);                                    // → empty lifecycle, now true
```

The consequence is worth having written down: a `before:empty` handler can distinguish "this Region is being torn down" from "someone called `empty()`" by reading `isDestroyed()`. That's a usable behavior nobody would guess, and the doc row — *"Emits `before:destroy`, enters destroyed, destroys and empties the current View, then emits `destroy`"* — states the same ordering, so table and test agree.

Interleaving `isSwappingView()` into each event label in the swap test does the same job for the flag's *duration* rather than just its final value, which is the part the docs make a specific claim about.

**One row isn't executable.** The prose states that with `allowMissingEl: true`, `show` "returns `undefined` and leaves the Region empty," and `empty` "returns the Region without changing unresolved DOM contents." The existing coverage in `test/unit/region.spec.js` asserts only that `show` *does not throw* in that configuration — which is equally satisfied by returning the Region, returning `undefined`, or leaving it occupied — and `_ensureElement({allowMissingEl: true}) === false` pins the internal mechanism, not the public return value. Nothing in the new contract file covers it either.

Given the table's purpose is to be read as authoritative by agents, a row whose specific claims no test can contradict is the one most likely to drift. Two assertions would close it: `expect(region.show(view)).to.be.undefined` plus `expect(region.hasView()).to.be.false` under `allowMissingEl: true`. Non-blocking — the claims look correct, they're just unpinned.

Scoping post-destroy operations out is consistent with the equivalent decision in #189, and the "Reuse is unsupported" cells say so in the table rather than leaving it implicit.

paulfalgout:
Addressed in `2736849`: the contract suite now asserts that `show` returns `undefined`, leaves the Region empty, and does not render the View when `allowMissingEl: true` and the element cannot resolve. Focused tests now pass 7/7, with docs and lint checks green locally.

paulfalgout:
Re-reviewed at `27368494` (the SHA I fetched). **Approve — nothing open from me.**

The two claims got opposite, and I think correct, treatments. The `show` claim is now pinned:

```js
expect(missingRegion.show(view)).to.be.undefined;
expect(missingRegion.hasView()).to.be.false;
expect(view.isRendered()).to.be.false;
```

That third assertion goes past what I suggested and is the one I'd have missed: "leaves the Region empty" would still hold if the View had been rendered and then discarded, so checking `isRendered()` pins that the work is skipped rather than done and thrown away.

The `empty` claim was deleted rather than tested, which is the right call for a sentence nobody was going to cover. An unverified assertion sitting in a table presented as the authoritative contract is worse than an absent one — it reads with the same weight as the rows that are executable, and there's nothing to catch it drifting.

PR #197, opened 2026-08-28
test(lifecycle): specify CollectionView state transitions

paulfalgout:
Related #131

## What

- Define the observable CollectionView lifecycle state table for the parent and its managed children.
- Add executable transition coverage for construction, render and rerender, Region detach and re-show, collection reset, explicit child detach/removal, external child destruction, and repeated parent destruction.
- Document default monitoring behavior plus the preventRender and monitorViewEvents exceptions.

## Why

View and Region now have explicit lifecycle contracts, but CollectionView still combined parent state, child ownership, and child lifecycle implicitly. Agents need a compact contract and executable examples before runtime ownership work proceeds.

## Scope

- Changes only the CollectionView reference documentation and one new lifecycle contract spec.
- Records current public behavior; no runtime, distribution, diagnostic, package, or API implementation changes.
- Leaves post-destroy operations other than repeated destroy, reentrant behavior, filters, Behavior ownership, and Application semantics for later #131 work.

## Deployment Impact

No application, website, documentation deployment, npm publication, or release tag is performed. This PR changes repository documentation and tests only.

## Testing

- npx vitest run test/unit/view-lifecycle.spec.js test/unit/region-lifecycle.spec.js test/unit/collection-view --coverage.enabled=false — 11 files, 307 tests passed.
- npm run docs:check — 29 pages built; 30 HTML files and 288 internal links validated.
- npm run lint:ci
- git diff --check

---
## Summary by cubic
Documents and tests the CollectionView lifecycle contract, specifying state transitions for the parent view and its managed children for #131. This is a test and documentation-only change; no runtime behavior changes.

Covers construction, render, Region detach/re-show, collection reset, child add/detach/removal, external child destruction, and repeated parent destruction, plus the `preventRender` and `monitorViewEvents: false` exceptions. Leaves post-destroy operations, reentrant behavior, filters, Behavior ownership, and Application semantics for later #131 work.

<sup>Written for commit 5e29a822d1b5fd915b2daa815d63dd1206629a81. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/197?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -123,9 +123,34 @@ the child views that may have previously been added.
 
 ## View Lifecycle and Events
 
-An instantiated `CollectionView` is aware of its lifecycle state and will throw events related
-to when that state changes. The view states indicate whether the view is rendered, attached to
-the DOM, or destroyed.
+Like `View`, a `CollectionView` exposes its lifecycle as the independent
+`isRendered()`, `isAttached()`, and `isDestroyed()` state values. Its managed
+children have their own View lifecycle state. Existing contents in the
+`CollectionView` element do not make the `CollectionView` rendered; rendering
+means its child set has been built and inserted into its element.
+
+The table describes the default rendered and monitored path. Passing
+`{ preventRender: true }` to `addChildView` still renders the parent when
+needed, but manages the supplied child without rendering it; detaching that
+child returns it in its current lifecycle state. Setting
+`monitorViewEvents: false` intentionally disables child attachment events and
+automatic child `isAttached()` updates.
+
+| Operation | CollectionView state | Managed child state |
+| --- | --- | --- |
+| Construct | Starts not rendered and not destroyed. It is attached only when its element is already in the document. | No children have been built. |
+| `render()` | Enters rendered and preserves its attached state. Repeated render stays rendered. | Builds and renders the current children. Repeated render destroys the previous children before building replacements. |
+| A rendered collection resets | Remains rendered and preserves its attached state. | Destroys the previous children and builds replacements for the reset collection. |
+| `addChildView(view)` | Renders first when needed, then remains rendered. | Renders and manages the added View. |
+| `detachChildView(view)` | State is unchanged. | Removes and returns the live View in a detached state. The caller becomes responsible for it. |
+| `removeChildView(view)` or external child destruction | State is unchanged. | Removes the child from the managed set. `removeChildView` destroys it; an externally destroyed child is removed once. |
+| The owning Region detaches and re-shows the CollectionView | Remains rendered while attached changes to `false`, then back to `true`. | Live children follow the parent's detached and attached state. |
+| `destroy()` | Detaches, becomes not rendered, and enters destroyed. Repeated destroy returns the CollectionView without repeating lifecycle events. | Detaches and destroys every still-managed child after the parent element is removed. |
+
+A View returned by `detachChildView()` is no longer managed by the
+`CollectionView`; another owner may show it, or the caller must destroy it.
+Operations on an already destroyed `CollectionView`, other than repeated
+`destroy()`, are not part of this lifecycle contract.
 
 Read More:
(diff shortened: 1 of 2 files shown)


paulfalgout:
Reviewed at `97b55867` (the SHA I fetched). **Approve.** Docs and tests only, and the spec carries the same discipline as the View and Region contracts.

The destroy test is the strongest piece, for a reason worth naming: it asserts *inside* the `before:destroy:children` handler that both `collectionView.isAttached()` and `child.isAttached()` are already `false`. That's what actually verifies the table's claim that children are destroyed "after the parent element is removed" — an end-state assertion would hold whichever order those two steps ran in. Interleaving parent and child events in one sequence array does the same job for the detach-before-destroy ordering.

**One thing the `monitorViewEvents` sentence doesn't say: whose flag it is.**

> Setting `monitorViewEvents: false` intentionally disables child attachment events and automatic child `isAttached()` updates.

I traced it rather than assuming, and it is the **CollectionView's** own flag in every case — never the child's:

```js
// modules/collection-view.js
516: const shouldTriggerDetach = view._isAttached && this.monitorViewEvents !== false;
577: const shouldTriggerAttach = this._isAttached  && this.monitorViewEvents !== false;
770: const shouldDisableEvents = this.monitorViewEvents === false;
789: if (this.monitorViewEvents === false) { this.Dom.detachContents(this.el); }
```

`this` is the CollectionView throughout, and the child's `view._isAttached = true/false` assignments sit inside those gates. The existing tests agree — `collection-view-children.spec.js:765` and `:887` both set it on `myCollectionView`. So a reader who sets `monitorViewEvents: false` on a *child* View to opt that child out gets nothing; its own flag is never consulted on these paths.

This is the same ambiguity #189 went three rounds on for the Region equivalent, and the resolution there was to name the owner explicitly — that doc now reads "on a Region's owning view." Worth matching here: "Setting `monitorViewEvents: false` **on the CollectionView** …". I'd rather flag it now than have the sibling docs disagree about a distinction that already cost a revert to get right once. (For the record, I was the one who got it wrong on #189 — which is why I traced this one before writing.)

**Minor, same sentence.** The claim has two halves, and only one is pinned. `children.spec.js:768-775` asserts `before:attach` and `attach` are not triggered on the child; nothing asserts the child's `isAttached()` stays `false`. The claim is true — line 590's `view._isAttached = true` is inside the `shouldTriggerAttach` gate — it just isn't executable anywhere, which is the same shape as the `allowMissingEl` row on #195. One assertion would close it.

`preventRender` is covered in the sorting, filtering, and children specs, so that half of the exceptions paragraph has backing.

paulfalgout:
Re-reviewed at `5e29a822` (the SHA I fetched). **Approve — nothing open from me.** Both points closed, and the test is better than what I suggested.

The doc now names the owner, matching how #189 settled the Region equivalent, so the two lifecycle documents no longer disagree about whose flag governs a contained view.

I'd asked only for an assertion that the child's `isAttached()` stays `false`. The test asserts the full state vectors of both objects at once:

```js
expect(state(collectionView)).to.deep.equal({ rendered: true,  attached: true,  destroyed: false });
expect(state(child)).to.deep.equal(         { rendered: true,  attached: false, destroyed: false });
```

That contrast is what makes it evidence rather than a coincidence. A lone `child.isAttached() === false` would also pass if the child had simply never reached the document; showing the parent as `attached: true` in the same breath establishes that the child *is* in an attached parent and reports `false` because monitoring is off. Asserting `rendered: true` on the child covers the other misreading — that disabling monitoring might suppress child rendering too, rather than attachment tracking specifically.

Issue #133, opened 2026-08-26
v5: specify Behavior scope, dependencies, and teardown


paulfalgout:
## Goal
Define the complete Behavior contract for host ownership, UI resolution, delegation, dependencies, communication, removal, and teardown.

## Dependencies
Requires lifecycle states (#131) and diagnostic semantics (#134).

## Agent-development failure addressed
Behavior boundaries are difficult to infer, which leads agents to cross DOM ownership or leak delegated work.

## Public behavior contract
A Behavior is scoped to its host View. Preserve the current host-ui merge unless separate benchmark evidence justifies changing it.

## Runtime cost category
Existing production path. No additional work for Views without Behaviors.

## Acceptance criteria
- [x] Document host, DOM, UI, event, and lifecycle ownership.
- [x] Specify dependency access and removal semantics.
- [x] Specify when host and Behavior `ui` values are captured, including construction before versus after the host View binds UI elements.
- [ ] Specify host-wins collision behavior and selector-versus-bound-element representations.
- [ ] Cover repeated delegation, dynamic removal, destroy, invalid references, and construction timing.
- [ ] Align diagnostics, tests, types, and canonical examples.

## Rollback or deprecation condition
A breaking UI-scope or merge change remains a candidate until public benchmark evidence supports it.

paulfalgout:
PRs #198 and #200–203 complete the host/lifecycle, UI timing, DOM/delegation, dependency/removal, and communication ownership slices.

Keep the remaining criteria open: the pre-rendered-host mixed UI representation is explicitly unresolved; repeated direct cleanup, dynamic replacement/removal edge cases, and invalid references remain; diagnostics, types, and canonical-example alignment are incomplete.

paulfalgout:
Progress update: #217 specifies that nested Behaviors are flattened into host ownership: directly destroying the declaring Behavior removes only that instance, the nested Behavior remains active on the same host, and host teardown later destroys the nested instance exactly once. The slice is test/docs-only with zero runtime delta. #133 remains open for the explicitly unresolved pre-rendered-host UI representation, repeated direct cleanup/reuse and dynamic replacement semantics, and final root declaration alignment under #137.

paulfalgout:
Progress: #223 merged as 5e961756. The Behavior UI example now executes against the packed package and verifies host-over-Behavior UI resolution, fresh UI lookup after rerender, public host coordination, and destroy cleanup.

PR #198, opened 2026-08-28
test(lifecycle): specify host-managed Behavior transitions

paulfalgout:
Related #131

## What

- Define the host-managed Behavior lifecycle for View and CollectionView hosts.
- Add executable ordering coverage for construction, render and rerender, Region attachment transitions, direct cleanup, nested Behaviors, host destruction, and repeated host destruction.
- Document lifecycle-monitoring behavior and explicitly defer ambiguous direct-repeat and reuse semantics.

## Why

Behavior lifecycle is currently distributed across host construction, proxied lifecycle events, entity and DOM delegation, nested Behavior flattening, and host teardown. Agents need one compact contract before later Behavior diagnostics and dependency work in #133.

## Scope

- Changes only the Behavior reference documentation and one new lifecycle contract spec.
- Records current public behavior; no runtime, distribution, diagnostic, package, or API implementation changes.
- Does not add Behavior isDestroyed state or define repeated direct destroy, post-cleanup reuse, UI collisions, dependency access, invalid references, dynamic replacement, Application, or State semantics.

## Deployment Impact

No application, website, documentation deployment, npm publication, or release tag is performed. This PR changes repository documentation and tests only.

## Testing

- npx vitest run test/unit/behavior-lifecycle.spec.js test/unit/behavior.spec.js test/unit/mixins/behaviors.spec.js --coverage.enabled=false — 3 files, 82 tests passed.
- npm run docs:check — 29 pages built; 30 HTML files and 288 internal links validated.
- npm run lint:ci
- git diff --check

---
## Summary by cubic
Documents and tests the host-managed Behavior lifecycle for View and CollectionView hosts, giving agents a single contract to rely on before Behavior diagnostics work in #133. Relates to #131.

- Records the current lifecycle ordering for construction, render, Region attachment and its `monitorViewEvents` interaction, direct cleanup, nested Behaviors, and repeated host destruction.
- Adds a new spec covering those transitions; no runtime, API, or distribution behavior changes.
- Deliberately defers direct-repeat destroy, reuse after cleanup, UI collisions, dependency access, dynamic replacement, and Application/State semantics.

<sup>Written for commit f5e5fe91fbb09fef2b122328edcca1274d6bb569. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/198?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -423,7 +423,10 @@ State lifecycle to Behavior.
 
 If a Region's owning view sets `monitorViewEvents: false`, the shown host does not
 receive attachment lifecycle notifications, so its Behaviors do not receive them
-either.
+either. Separately, setting `monitorViewEvents: false` on the host itself does not
+by itself suppress Region attachment lifecycle. It suppresses the host's
+`dom:refresh` and `dom:remove` notifications, so its Behaviors do not receive those
+notifications.
 
 ## Destroying a Behavior
 
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -29,6 +29,7 @@ Instead a `Behavior` should be instantiated by the view it is related to by
   * [Events / Initialize Order](#events--initialize-order)
   * [Using `ui`](#using-ui)
   * [View DOM proxies](#view-dom-proxies)
+* [Behavior Lifecycle](#behavior-lifecycle)
 * [Destroying a Behavior](#destroying-a-behavior)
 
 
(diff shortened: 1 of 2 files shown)


paulfalgout:
Reviewed at `d4131d89` (the SHA I fetched). **Approve.** Docs and tests only, and the destroy test is the most rigorous in this series so far.

Two things in it are worth naming. The `isDestroyed()` assertions live *inside* the Behavior handlers, so the alive-then-destroyed straddle the table claims — *"receives `before:destroy` while the View is alive … then receives `destroy` after the View's callback"* — is actually verified rather than restated. And:

```js
expect(parentBehavior.stopListening).to.have.been.calledOnce;
expect(nestedBehavior.stopListening).to.have.been.calledOnce;
```

across two `view.destroy()` calls. That's a stronger idempotence claim than the event array gives you: a second cleanup pass could be silent event-wise (the destroyed guard suppresses re-emission) while still tearing down listeners twice. Checking the cleanup call rather than its observable side effect is the assertion that would catch it. Asserting `nestedBehavior.view === view` also pins the flattening claim directly.

**The closing `monitorViewEvents` paragraph names the owner, which I'm glad to see** — consistent with how #189 and #197 settled it. I verified the causal chain rather than taking it: `Region._shouldDisableMonitoring()` reads `this._parentView.monitorViewEvents` ([region.js:163](modules/region.js:163)), which gates the host's `attach`/`detach` triggers ([region.js:185](modules/region.js:185)); Behaviors receive host events through `this.listenTo(view, 'all', this.triggerMethod)` ([behavior.js:49](modules/behavior.js:49)), so an event the host never emits is one its Behaviors never see. The sentence is correct.

**But there are two flags here, and the paragraph covers one.** The *host's own* `monitorViewEvents: false` has a different and narrower effect, which I traced:

- It does **not** suppress `attach`/`detach` for the host — Region triggers those directly, gated only on the owning view's flag.
- It **does** suppress `dom:refresh` and `dom:remove`, because those come from the handlers `monitorViewEvents(view)` registers, and that function returns early on `view.monitorViewEvents === false` ([monitor-view-events.js:75](modules/common/monitor-view-events.js:75), called from [region.js:143](modules/region.js:143)).

So a host with its own flag disabled still delivers attachment lifecycle to its Behaviors, but never delivers `dom:refresh`/`dom:remove` to them. That's a case an agent reading this table would get wrong in either direction — assuming the host's flag behaves like the owner's, or assuming it does nothing. One sentence distinguishing them would finish the convention this series has been converging on. Non-blocking.

paulfalgout:
Re-reviewed at `f5e5fe91` (the SHA I fetched). **Approve — nothing open from me.**

The added sentences draw the distinction precisely, and *"does not by itself suppress"* is the right hedge: it states what the host's own flag doesn't do without overclaiming about the case where both flags are set.

I also checked whether the new claim has executable backing before calling this closed, since I've been holding the series to that. It does, one level down: [`test/unit/common/monitor-view-events.spec.js:18`](test/unit/common/monitor-view-events.spec.js:18) asserts that with `monitorViewEvents: false` the handlers are never attached, and those handlers are the only thing that emits `dom:refresh`/`dom:remove`. So "handlers not registered" entails "those notifications don't fire," which is stronger backing than the `allowMissingEl` row had on #195 — there, the existing test asserted only that `show` didn't throw, which didn't constrain the documented return value at all. I'd have flagged this the same way if it were in that shape; it isn't.

That closes the last of my notes on this one. The four-part #131 series — View, Region, CollectionView, Behavior — now uses the same convention for naming which object's flag governs, which was the recurring ambiguity across all of them.

Issue #128, opened 2026-08-26
v5: publish a neutral reference app and pinned agent benchmark


paulfalgout:
## Current status — September 6, 2026

Deferred intentionally to the real-world examples and agent-evidence stage. Benchmark infrastructure exists; empirical agent results and the reference application remain outstanding.

Reconciled against [master at `8ea37088`](https://github.com/marionettejs/marionette/commit/8ea3708841e5ccb9a90f98d078cc4ca53c04ea1d). This is a source candidate for the next alpha, not a declaration of stable-v5 readiness.

## Goal
Create an original public reference application, a frozen paired-comparable task corpus, the candidate-only task contract, a frozen agent harness and evaluator, a powered sample-size plan, and the Phase 0 paired baseline.

## Dependencies
Requires clean installation (#150), governance cleanup (#152), the supported profile (#153), and the shared rule catalog (#129).

## Agent-development failure addressed
Agent-readiness claims are currently anecdotal and cannot be reproduced from this repository. A fixed but underpowered run count would also make a statistical non-regression gate look rigorous without being able to detect meaningful regressions.

## Public behavior contract
The fixture uses only public APIs and defines no new runtime behavior.

## Runtime cost category
Test/benchmark only. None of the corpus or harness ships in production artifacts.

## Acceptance criteria
- [ ] Cover representative nested Region and Application outcomes, Behaviors, resources, async startup and teardown, communication, CollectionView work, invalid cases, and role selection through paired-comparable tasks whenever public APIs in both revisions can complete the objective.
- [ ] Cover accepted candidate-only contracts such as Application-owned and View-owned state sources, correct state-source versus domain-data ownership, and nested Application lifecycle only when acceptance necessarily exercises a public API absent from Phase 0; a poor or failed baseline result never justifies candidate-only classification.
- [ ] Cover a shared-overlay-host pattern through a paired-comparable task that displays only Views through a real Region and keeps positioning and exclusivity outside core.
- [ ] Publish at least 10 paired-comparable tasks with hidden acceptance tests; candidate-only tasks do not count toward that floor. Across the full candidate corpus, exercise every named capability area through at least two independently scored tasks; one task may cover multiple areas, but no single task certifies an area.
- [ ] Predeclare every task's stratum and run count before its pilot or scored runs, with every count at least 10. Run a Phase 0 pilot that gives each paired-comparable task at least 80 percent power, after the chosen multiple-comparison correction, to detect a 15-percentage-point absolute regression.
[...]

PR #206, opened 2026-08-28
docs(benchmark): define cross-revision comparability

paulfalgout:
Related #128

## What

- Split the agent benchmark into frozen `paired-comparable` and `candidate-only` evaluation strata.
- Apply absolute candidate-quality gates to the full corpus while limiting improvement, McNemar, and relative architecture-violation claims to the paired stratum.
- Define preclassification, minimum task/run counts, and the changes that invalidate a benchmark series.

New v5 capabilities absent from Phase 0 cannot be scored as though their baseline failure were a comparable regression, while candidate-only successes must not inflate claims that Marionette improved on work both revisions could perform.

## Public behavior contract

- Canonical behavior after this PR: every task is preclassified; at least 10 tasks are genuinely pairable; candidate-only work supports absolute conformance evidence but never comparative claims.
- Superseded behavior removed, if any: the roadmap no longer treats every future benchmark task as implicitly pairable with Phase 0.
- Compatibility exception and removal condition, if any: none; runtime and package APIs are unchanged.

## Runtime cost

- Bundle: unchanged at 49.82 kB aggregate against the 51.98 kB ceiling.
- Startup/hot path: unchanged; no runtime source is modified.
- Allocations/retention: unchanged; no runtime source is modified.

## Scope

- `ROADMAP.md` measured-agent-outcomes policy.
- `benchmarks/agent/README.md` corpus and evaluation-strata guidance.
- No task schema, capability vocabulary, validator, runtime, workflow, issue status, package, release, website, or repository-settings change.

## Deployment Impact

No package, application, documentation-site, or form deployment is required. Pages publication remains disabled.

## Validation

Commands:

```sh
npm ci
npm run docs:check
npm run test:agent-benchmark-contract
npm run size
git diff --check
```

Results: 357 locked packages installed with zero vulnerabilities; 29 pages built and 288 internal links validated; 19/19 benchmark-contract tests passed; production size remained 49.82 kB with graphs 40/1/1; the diff check passed.

## Testing

The documentation build/link validator, agent benchmark contract tests, deterministic production size/graph/resource checks, and diff check all passed locally. No browser behavior changed.

## Package and browser impact

## Rollback or deprecation

Revert this policy before collecting pilot results if the two-strata model cannot be evaluated consistently; never mix results from the reverted and replacement series.

## Review notes

The future benchmark profile should make classification and run counts machine-readable. This PR intentionally does not change `task.schema.json` because comparability depends on the frozen revision pair rather than task metadata alone.

---
## Summary by cubic
Splits the agent benchmark into paired-comparable and candidate-only task strata so new-revision-only capabilities aren't scored as regressions against Phase 0 and candidate-only successes can't inflate comparative claims. Relates to #128.

- Every task is now preclassified before pilot or scored runs; at least 10 tasks must be genuinely pairable with Phase 0.
- Relative claims (improvement, McNemar non-regression, architecture-violation reduction) use only paired-comparable runs; candidate-only results are reported separately.
- Absolute quality gates (Wilson lower bound, per-task floor) still apply across the full corpus.
- Changes to the pinned model, harness, permissions, paired task artifacts, evaluator, pairing, classification, or statistical procedure start a new series; changing a frozen candidate-only task invalidates the full-corpus result.
- Documentation-only change; runtime and package APIs are unchanged.

<sup>Written for commit 316279ea289cffc32632d7fa0503e52322991e04. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -138,31 +138,48 @@ budget.
   composing Regions and Applications, repairing lifecycle bugs, adding Behaviors,
   implementing an overlay through a shared host Region, using communication
   boundaries, and proving cleanup.
-- The model version, agent harness, prompt, repository revision, commands, evaluator,
-  and expected outcomes are pinned for each benchmark series.
-- The stable-v5 benchmark uses at least 10 tasks. A Phase 0 pilot predeclares the
-  paired-run count for each task; each count is at least 10 and provides at least 80
+- The model version, agent harness, permissions, commands, evaluator, expected
+  outcomes, repository revisions, and task classification are pinned for each
+  benchmark series.
+- Every task is assigned to one of two strata before its pilot or scored runs. A
+  paired-comparable task uses byte-identical prompts, visible workspaces, hidden
+  acceptance tests, commands, and evaluator rules for the Phase 0 revision and release
+  candidate; only the installed Marionette revision differs, and the objective is
+  achievable through public APIs in both revisions. A candidate-only task is allowed
+  only when its acceptance criteria necessarily exercise an accepted public contract
+  absent from Phase 0. A poor or failed Phase 0 result never justifies reclassification.
+- The stable-v5 benchmark uses at least 10 paired-comparable tasks. Candidate-only
+  tasks do not count toward that floor. Across the full candidate corpus, every named
+  capability area is exercised by at least two independently scored tasks. One task
+  may cover multiple areas, but no single task certifies an area.
+- The classification and run count for every task are predeclared, and every count is
+  at least 10. A Phase 0 pilot sets each paired-comparable task's count with at least 80
   percent power to detect an absolute regression of 15 percentage points. Release
   decisions use one-sided exact McNemar tests and control family-wise error at 0.05
   with the Holm correction. Sample-size planning uses the conservative Bonferroni
-  level of 0.05 divided by the task count and the pilot's one-sided 95 percent upper
-  confidence bound for discordant pairs under that regression alternative. The
-  executable power calculation and inputs are published before candidate runs.
-- Every named capability area is exercised by at least two independently scored
-  tasks. One task may cover multiple areas, but no single task certifies an area.
-- The aggregate fully-correct rate has a 95 percent Wilson lower bound of at least
-  80 percent, and no individual task has a fully-correct point estimate below 60
-  percent. Aborted runs count as not fully correct.
-- Relative to the Phase 0 baseline on the same pinned harness, the fully-correct point
-  estimate does not regress and either improves by at least 20 percentage points or
-  reaches at least 95 percent. Cataloged framework-architecture violations per 100
-  attempted runs fall by at least 50 percent. Violations found before an aborted run
-  still count.
-- No individual task has a statistically significant paired regression after the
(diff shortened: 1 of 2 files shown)


paulfalgout:
Reviewed at `316279ea` (the SHA I fetched). **Approve.** The stratification is the right correction, and the clause that makes it work is *"A poor or failed Phase 0 result never justifies reclassification."* Without that, the split is an invitation: any paired task that unexpectedly tanks on Phase 0 gets quietly moved to candidate-only, and the improvement claim survives by attrition. Naming and forbidding it up front is what turns preclassification into a real control rather than a stated intention.

Two other details land correctly. Bonferroni now divides by the *paired-comparable* count, which is the right family — the candidate-only tasks aren't in the comparative test set, and dividing by the total would have been needlessly conservative. And *"candidate-only results and violations never enter a comparative denominator"* closes the matching hole in the violations-per-100-runs metric.

**One asymmetry worth pricing, because it lands on the newest code.** Run counts split unevenly:

- Paired-comparable: *"A Phase 0 pilot sets each paired-comparable task's count with at least 80 percent power…"* — the pilot will push these well above the floor.
- Candidate-only: no power or precision requirement. `at least 10` is the only constraint.

But both strata are subject to the absolute gates, including *"no individual task has a fully-correct point estimate below 60 percent."* At n=10 that gate is noisy in both directions:

```
per-task gate (>= 6/10 at n=10)
  true rate 50%  ->  passes 37.7% of the time
  true rate 60%  ->  passes 63.3%
  true rate 70%  ->  fails  15.0%
  true rate 80%  ->  fails   3.3%

same gate at n=50 (>= 30/50)
  true rate 50%  ->  passes 10.1%
  true rate 70%  ->  passes 95.2%
```

So a candidate-only task measuring a genuinely 70%-capable feature blocks release about one run-set in seven, and one measuring a genuinely 50%-capable feature ships about three times in eight. Candidate-only tasks are exactly the ones covering v5-only contracts — the newest code with the least-exercised docs — so the least-proven capabilities end up gated by the noisiest measurements, with no comparative check to contextualize the result. Giving candidate-only counts a precision requirement (a target CI half-width, say) rather than a bare floor would fix it, and the number can be far smaller than the paired stratum's since no McNemar test is involved.

**Incidentally this relieves the coverage-arithmetic concern I raised on #199.** Since v5-only capabilities can't satisfy *"achievable through public APIs in both revisions,"* they must be candidate-only, and candidate-only tasks don't count toward the ≥10 paired floor. The corpus therefore has to exceed 10 tasks outright, so the two-independent-tasks-per-capability rule is much less cramped than when 14 capabilities had to fit inside a 10-task minimum. Worth noting that the fix arrived as a side effect rather than by design.

Issue #134, opened 2026-08-26
v5: add shared Marionette diagnostics with stable codes


paulfalgout:
## Goal
Use the existing MarionetteError taxonomy with stable rule codes for framework invariant failures.

## Dependencies
Requires the catalog and repository-local generated diagnostic contract in #129. Public website publication is deferred to post-v5 issue #177.

## Agent-development failure addressed
Incidental TypeErrors and prose-only messages cannot be classified reliably by agents, tests, lint rules, or docs.

## Public behavior contract
Preserve and formalize existing names such as `ViewError`, `CollectionViewError`, `NoChildViewError`, and `InvalidChildViewError`. Codes are stable at v5 final; messages may improve. Stable codes and repository-generated docs do not depend on a live website. Do not standardize accidental JavaScript exceptions.

## Runtime cost category
Exceptional production paths. Compact codes are allowed; the full rule catalog stays out of core.

## Acceptance criteria
- [x] Define and export the public error shape and code lookup contract.
- [ ] Seed conversion with unknown Region operations, unknown literal `@ui` references, missing literal string handlers, unknown Radio/reset operations, and other verified raw TypeError or silent-failure paths.
- [x] Preserve meaningful existing error names while giving each invariant a stable code.
- [x] Validate codes and repository-generated documentation anchors against #129 without requiring a live website.
- [x] Do not point runtime messages at unpublished `docs.marionettejs.com` routes; #177 owns public link cutover after v5 launch.
- [x] Test code stability separately from message prose.
- [x] Verify stack behavior on engines with and without `Error.captureStackTrace`.
- [x] Document SemVer and deprecation handling.

## Rollback or deprecation condition
Before stable, revise codes through the catalog. After stable, deprecate rather than reuse or silently change them.

paulfalgout:
Phase 0 catalog groundwork is now merged via #170.

`npm run check:diagnostics` currently derives the shipped Rollup source graph and rejects every literal emitted `code` that is uncataloged, still `defined`, or missing the `runtime` surface. It intentionally does not reject the existing code-less `MarionetteError` construction sites because compact runtime emission belongs to this issue.

When implementing #134, please make the completeness boundary explicit:

- Give every deliberate runtime diagnostic construction site a literal catalog `code`.
- Change each adopted entry from `defined` to `active` in the same change.
- Add a static completeness check so a new code-less `MarionetteError` site cannot pass CI after migration.
[...]

PR #211, opened 2026-08-28
fix(diagnostics): reject unknown Radio resets

### LINKED ISSUE #82: v5: specify and harden Radio lifecycle and diagnostics
## Goal

Specify Radio channel, request/reply, reset, debug, ownership, and cleanup behavior with stable diagnostics.

Agents cannot reliably distinguish intended Radio behavior from adapter behavior or incidental JavaScript errors.

## Public behavior contract

Define known and unknown channel semantics, reply replacement, reset behavior, cleanup ownership, debug policy, and error codes. Supersede accidental raw exceptions rather than standardizing them.

## Runtime cost category

Existing production path and exceptional diagnostics. Preserve clean successful-path performance.

## Acceptance criteria

- [ ] Publish the Radio lifecycle and operation table.
- [ ] Define reset and unknown-channel semantics.
- [ ] Map invariant failures to #129/#134 codes.
- [x] Complete core coverage through #69.
- [ ] Measure successful-path and bundle impact.

## Rollback or deprecation condition

Remove pre-release behavior and tests that conflict with the final specified contract.

paulfalgout:
Closes #82
Related #134

## What

- Define zero-argument, named, falsy-name, and unknown-channel behavior for Radio.reset with stable MN0017 and MN0021 diagnostics.
- Use null-prototype channel and log registries so prototype-like names remain ordinary channels without changing successful channel identity or handler cleanup.
- Document the Radio lifecycle operation table, migration behavior, and diagnostic catalog entry with focused regression coverage.

## Why

Unknown named resets currently fall through to incidental TypeErrors, while inherited object properties can be mistaken for channels. Agents and developers need deterministic reset semantics and stable framework error codes without adding cost to successful event and request paths.

## Scope

- Affects Radio channel/log storage, Radio.reset behavior, Radio documentation, migration guidance, diagnostics, and generated main distribution artifacts.
- Preserves Radio.channel(name) creation, top-level event/request/tuning channel creation, channel identity, and zero-argument reset-all behavior.
- A supplied falsy name remains invalid and now consistently reaches MN0017 without resetting existing channels.
- Does not publish documentation, change the website, create a tag, publish npm, or release v5.

## Deployment Impact

No application or website deployment is required. Normal library integration after a future release is sufficient.

## Testing

- npm run coverage — 70 files, 1,212 tests, 100% statements, branches, functions, and lines; 19 agent benchmark contract tests.
- npm run lint:ci
- npm run check:diagnostics — 21 catalog entries.
- npm run docs:check — 33 pages, 34 HTML files, and 320 internal links.
- npm run test:source
- npm run test:dist
- npm run test:fixtures
- npm run test:performance — 89 tests.
- npm run check:release-profile
- npm run check:browser-profile
- npm run check:release-promotion
- npm run check:dist
- npm run size — exact-base cumulative growth +160 B, total 51,878 B / 51,975 B, 97 B remaining; every artifact remains below 1% and no approval is required.
- git diff --check

---
## Summary by cubic
Rejects unknown channel names in `Radio.reset` with a stable diagnostic and allows prototype-like channel names like `toString` to behave as ordinary channels.

Previously, resetting an unknown channel fell through to an incidental TypeError, and inherited object properties were mistaken for channels. Now `Radio.reset(name)` throws `MarionetteError` with code MN0021 for unknown names, and falsy names always produce MN0017 without resetting anything. Channel and log registries use null-prototype objects, so names such as `toString` and `constructor` are treated as normal channel names. Zero-argument `Radio.reset()` behavior is unchanged.

- Closes #82 and addresses #134.

<sup>Written for commit d57a39ff3abc5ac523142c8ea2b869b716d31a11. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/211?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
Exact-head Claude review requested for d57a39ff3abc5ac523142c8ea2b869b716d31a11. Please review Radio reset semantics, null-prototype channel/log registries, one-time hostile-name coercion, mutation-free unknown resets, prototype-name identity and cleanup, MN0017 versus MN0021 behavior, documentation accuracy, and runtime cost. Local exact-head validation: 70 files / 1,212 tests at 100% coverage, 19 agent benchmark tests, 89 performance tests, lint, 21 diagnostics, docs, source, dist, fixtures, reproducibility, release preflight, size, and diff checks. Exact-base cumulative growth is +160 B; total 51,878 B / 51,975 B with every artifact below 1%.

paulfalgout:
Reviewed at `d57a39ff` (the SHA I fetched). **Approve.** It fits — but only just, and that's the thing I'd want on the record.

```
AGGREGATE  51718 -> 51878   +160 (+0.31%)
min-build  +57 bytes
ceiling    51975  ->  97 bytes remaining  (~19 min-build bytes)
```

Under the 1% per-artifact threshold, so no growth approval. But 97 bytes is not headroom for anything: the next runtime change of any size cannot land without the #207 amendment protocol, which has still never run end to end. Given #209 showed that the first live exercise of a comparable chain surfaces real defects — the required-workflow lookup silently returned nothing — I'd rather see the amendment PR land deliberately, before it's the thing blocking a change someone needs.

**The bug this fixes is worse than "incidental TypeErrors" undersells.** The old line was:

```js
const channels = !channelName ? this._channels : [this._channels[channelName]];
```

So `Radio.reset('')`, `Radio.reset(null)`, `Radio.reset(0)` — any falsy name — took the reset-*all* branch. A caller doing `Radio.reset(nameFromConfig)` with an empty value silently tore down every channel in the application, with no error. Unknown-name resets threw a TypeError, which is at least visible; the falsy case failed silently and destructively. Splitting on `arguments.length` so only a genuine zero-argument call resets everything is the right fix.

**The test pins the destructive half, not just the throw**, which is what makes it worth having:

```js
['', null, false, 0, undefined].forEach(channelName => {
  expect(() => Radio.reset(channelName)).to.throw().with.property('code', 'MN0017');
});
Radio.trigger('existing', 'event');
expect(handler).to.have.been.calledOnce;
```

Asserting the surviving handler still fires is the assertion that would have caught the old behavior; the throw alone wouldn't. Including `undefined` in that list also pins the `arguments.length` distinction — `reset(undefined)` throws while `reset()` resets all — which is the subtle part of the change.

**You applied both #210 lessons here without being asked**: the lookup is wrapped so a hostile property key yields MN0021 rather than a raw TypeError, and the message omits the name entirely rather than coercing it a second time. `Object.create(null)` for `_channels` and `_logs` means `reset('toString')` no longer finds `Object.prototype.toString` and calls `.reset()` on it, and the prototype-named-channel test proves reset-all still reaches channels literally named `toString` and `__proto__`.

One small note: `if (!channelName) { Radio.channel(channelName); }` is called purely for its throw, which reads as a discarded no-op — a future reader could delete it and silently convert MN0017 into MN0021. A comment would cost nothing and it's clearly a deliberate byte trade given where the ceiling is.

Issue #137, opened 2026-08-26
v5: finish generated public API metadata


paulfalgout:
## Current status — September 9, 2026

Repository readiness work is merged in [#500](https://github.com/marionettejs/marionette/pull/500) at [master `70eaeaf5`](https://github.com/marionettejs/marionette/commit/70eaeaf582a3ae062b5ff48650cfda12cfa3bf9f). It is available in source and verified tarballs; no new npm version or tag was published. The registry remains beta.1 on `latest`.

Generated public API metadata and drift checks are integrated. The compact reference is generated from the same semantic records. This closes static metadata delivery; measured reference usefulness remains #140/#128/#146.

## Goal
Keep public declarations truthful and add the compact, drift-checked API manifest needed by documentation and agent tooling. Metadata and type generation add no production JavaScript.

## Delivered
- [x] Cover root runtime exports and optional data/adapters entrypoints with public declarations (#429).
- [x] Check native classes, `.extend`, constructor/options contracts, UI, Regions, bindings, child events, fluent methods and lifecycle overrides.
- [x] Compile ESM/CommonJS type fixtures and actual installed-package consumers.
- [x] Verify package export conditions, including separate generated root ESM/CommonJS declaration trees.

Evidence: #429, [test/types](https://github.com/marionettejs/marionette/tree/8ea3708841e5ccb9a90f98d078cc4ca53c04ea1d/test/types), [test/fixtures](https://github.com/marionettejs/marionette/tree/8ea3708841e5ccb9a90f98d078cc4ca53c04ea1d/test/fixtures), [build/declarations.mjs](https://github.com/marionettejs/marionette/blob/8ea3708841e5ccb9a90f98d078cc4ca53c04ea1d/build/declarations.mjs), and the package manifests. Canonical optional-package declaration generation is delivered by the closed source-migration issue #403.

## Local implementation awaiting integration
- [x] Inventory nine runtime entrypoints and 131 exports with compiler-derived signatures and 56 explicit semantic profiles, linked to canonical guide headings and actual public test declarations.
- [x] Check 62 source fingerprints plus cited fixtures/docs; generated inventory drift fails local verification, CI and candidate certification.
- [x] Package the metadata and inventory guide with documentation while excluding checker and consumer-lint tooling from ordinary production graphs.
- [x] Explicitly classify the separate `marionette/eslint` development export; it does not expand runtime APIs.

[...]

Issue #71, opened 2026-05-21
v5: Add Backbone shim interop tests


paulfalgout:
## Goal
Verify the optional Backbone shim through public package and runtime contracts.

## Release-blocking rationale
`marionette/backbone` is a published v5 entrypoint. It must either pass its documented package/runtime contract or be removed before stable.

## Dependencies
Requires artifact integrity (#151), declarations/package conditions (#137/#144), and the supported profile (#153).

## Public behavior contract
The shim remains optional. Tests describe final supported behavior without making an external application a gate.

## Acceptance criteria
- [ ] Verify supported import order and package entrypoint behavior.
- [ ] Cover model, collection, events, and identity interop promised by the shim.
- [ ] Document unsupported behavior explicitly.
- [ ] Keep the root production entrypoint independent of the shim.

## Runtime cost
Test and package-contract work. Applications that do not import the shim pay no cost.

paulfalgout:
The acceptance criterion `Backbone.on(...) is no longer a thing` conflicts with the current merged shim contract.

`backbone.js` extends only `Backbone.Model`, `Collection`, `View`, and `Router` prototypes. Backbone itself already exposes namespace event methods, and `test/unit/backbone-shim.spec.js` explicitly verifies that importing the shim leaves `Backbone.on`, `off`, and `trigger` equal to their pre-shim values.

Before taking this test-only issue, can we clarify the intended assertion? The source-consistent version appears to be: "the shim does not add or replace Marionette helpers on the Backbone namespace." Removing Backbone's own `on` would be a new public behavior change outside this issue's current boundary.

paulfalgout:
Progress after #212: the runtime/package interop portion now has isolated ESM and CommonJS processes for both Backbone-first and shim-first loading, exact module/constructor/prototype identity checks, Model/Collection/listener behavior coverage, and explicit unsupported-boundary documentation. Runtime and every shipped JavaScript artifact remain byte-identical. This issue stays open while declaration/package-condition dependencies #137 and #144 are unfinished; the next active slice targets the ./backbone conditional declarations under #137.

PR #213, opened 2026-08-28
feat(backbone): add conditional shim declarations

paulfalgout:
Related #137
Related #71

## What

- Add conditional ESM and CommonJS declarations for the `marionette/backbone` entrypoint.
- Preserve the existing runtime export targets while binding each module condition to its matching declaration format.
- Add isolated strict NodeNext installed-package fixtures for both module modes, including supported and intentionally unsupported augmentation boundaries.

## Why

The runtime shim contract is now covered, but the public subpath had no TypeScript declaration contract. Consumers should receive the canonical Backbone type plus the four runtime-added `triggerMethod` surfaces without exposing unsupported namespace or `History` members.

## Scope

This changes only the `./backbone` package export metadata, its two declaration files, and installed-package fixture coverage. The root entrypoint and other subpaths are intentionally unchanged. No JavaScript runtime or generated JavaScript artifact changed.

## Deployment Impact

## Testing

- `npm run validate`
- `npm run test:fixtures`
- `npm run lint:ci`
- `npm run check:dist`
- `npm run size`
- `git diff --check`
- Separate `tsc -p tsconfig.esm.json` and `tsc -p tsconfig.cjs.json` fixture programs verify conditional declaration selection without cross-program augmentation leakage.
- Full validation passed with 1,214 tests at 100% coverage, 19 agent benchmark tests, 89 performance tests, docs/source/dist/release checks, and unchanged runtime size at 51,878 / 51,975 bytes.

---
## Summary by cubic
Adds conditional ESM and CommonJS type declarations for the `marionette/backbone` entrypoint, so TypeScript consumers get `triggerMethod` typed on Backbone models, collections, views, and routers. Previously this subpath had no type contract; the runtime shim and all other entrypoints are unchanged, and no deployment or release is required.

**Testing**
- Adds strict NodeNext installed-package fixtures for both module modes that verify the declarations resolve correctly and that unsupported namespace and `History` members still error.

<sup>Written for commit 4356854045a9512a410b34d9be9ac37357ccad82. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/213?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/dist/backbone.d.cts
+++ b/dist/backbone.d.cts
@@ -1,0 +1,21 @@
+import Backbone = require('backbone');
+
+declare module 'backbone' {
+  interface Model<T extends ObjectHash = any, S = ModelSetOptions, E = any> {
+    triggerMethod(event: string, ...args: unknown[]): unknown;
+  }
+
+  interface Collection<TModel extends Model = Model> {
+    triggerMethod(event: string, ...args: unknown[]): unknown;
+  }
+
+  interface View<TModel extends Model | undefined = Model, TElement extends Element = HTMLElement> {
+    triggerMethod(event: string, ...args: unknown[]): unknown;
+  }
+
+  interface Router {
+    triggerMethod(event: string, ...args: unknown[]): unknown;
+  }
+}
+
+export = Backbone;
--- a/dist/backbone.d.mts
+++ b/dist/backbone.d.mts
@@ -1,0 +1,21 @@
+import Backbone from 'backbone';
+
+declare module 'backbone' {
+  interface Model<T extends ObjectHash = any, S = ModelSetOptions, E = any> {
+    triggerMethod(event: string, ...args: unknown[]): unknown;
+  }
+
+  interface Collection<TModel extends Model = Model> {
+    triggerMethod(event: string, ...args: unknown[]): unknown;
+  }
+
+  interface View<TModel extends Model | undefined = Model, TElement extends Element = HTMLElement> {
+    triggerMethod(event: string, ...args: unknown[]): unknown;
+  }
+
+  interface Router {
+    triggerMethod(event: string, ...args: unknown[]): unknown;
+  }
+}
+
+export default Backbone;
--- a/package.json
+++ b/package.json
@@ -24,8 +24,14 @@
       "default": "./dist/marionette.js"
     },
     "./backbone": {
-      "import": "./dist/backbone.js",
-      "require": "./dist/backbone.cjs",
+      "import": {
+        "types": "./dist/backbone.d.mts",
+        "default": "./dist/backbone.js"
+      },
+      "require": {
+        "types": "./dist/backbone.d.cts",
+        "default": "./dist/backbone.cjs"
+      },
       "default": "./dist/backbone.js"
     },
     "./jquery-dom-api": {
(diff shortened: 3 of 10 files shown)


paulfalgout:
Exact-head Claude review requested for 4356854045a9512a410b34d9be9ac37357ccad82. Please verify conditional export-map declaration selection, ESM default versus CommonJS export-assignment fidelity, supported triggerMethod augmentation boundaries, isolation of the two TypeScript programs, installed-package coverage, and zero JavaScript/runtime drift. Local exact-head validation passed: 1,214 tests at 100% coverage, 19 agent benchmark tests, full fixture suite including separate strict NodeNext ESM/CommonJS compiles, 89 performance tests, lint, docs, source/dist/reproducibility/release checks, and unchanged runtime size at 51,878 / 51,975 bytes.

paulfalgout:
Reviewed at `43568540` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, so the 97-byte margin is intact.

**The exports wiring is correct in the way that usually goes wrong.** `types` is first within each of `import` and `require`, which TypeScript requires — a `types` condition listed after `default` is silently ignored, and the failure mode is "types just don't resolve" with no error. Separate `.d.mts` and `.d.cts` with matching module syntax (`export default` versus `export =`) is the right dual-package shape rather than duplication, and `NodeNext` in the fixture tsconfigs is what makes the conditional resolution actually exercised. `skipLibCheck: false` is stricter than most setups and means the declaration files are themselves type-checked instead of trusted.

**I want to note explicitly that declaring only `triggerMethod` is correct, not an omission**, because it looks like one at first. `backbone.js` does `extend(proto, Events)` for all four prototypes, which grafts eight members — `on`, `off`, `once`, `listenTo`, `listenToOnce`, `stopListening`, `trigger`, `triggerMethod`. Seven of those already exist in `@types/backbone`, so redeclaring them in a `declare module` block would risk conflicting with the upstream types rather than improving them. `triggerMethod` is the only genuinely new name, and adding exactly that is the right scope. `docs/optional-backbone.md:222` documents the wholesale swap in prose, which is where that belongs — the behavioral difference between Backbone's `Events` and Marionette's isn't expressible in a signature anyway.

**One drift risk worth deciding about.** The four-prototype list now exists in three hand-maintained places:

```
backbone.js         extend(Backbone.Model/Collection/View/Router .prototype, Events)
dist/backbone.d.mts declare module 'backbone' { Model, Collection, View, Router }
dist/backbone.d.cts same four
```

Nothing links them. No build step generates the declarations — I checked `package.json` and the config directory for a `.d.mts`/`.d.cts` producer and found none — so if a fifth prototype is augmented, or one is dropped, the types keep asserting the old set and every check here still passes. The type fixtures validate that the declarations *compile and resolve*, not that they match what `backbone.js` does.

The cheap fix is that #212 already pins the runtime list: `assert-interop.cjs` iterates `Object.values(constructors)` asserting each has `triggerMethod`. A type-level counterpart — asserting `triggerMethod` is visible on exactly those four and erroring on a fifth — would close the loop, or generating the declarations from the shim source would remove the duplication outright. Non-blocking either way; it's a small surface and it's correct today.

PR #214, opened 2026-08-28
test(lifecycle): specify setElement transitions

paulfalgout:
Related #131

## What

- Add table-driven lifecycle coverage for alive View and CollectionView element replacement across empty/populated and attached/detached roots.
- Characterize repeated replacement calls and preservation of nested Region and CollectionView child ownership without moving child DOM.
- Document the distinct View and CollectionView rendered-state contracts and correct the previous View lifecycle overstatement.

## Why

Agents should not need to infer setElement lifecycle behavior from implementation details. The existing lifecycle tables omitted the replace-element transition, and the View guide incorrectly implied that an alive rendered View could never become unrendered.

## Scope

This changes only View and CollectionView lifecycle tests and documentation. Post-destroy replacement behavior, Application/State design, runtime modules, generated artifacts, package metadata, and performance contracts are intentionally unchanged.

## Deployment Impact

## Testing

- Full Vitest coverage: 70 files, 1,218 tests, 100% statements/branches/functions/lines.
- `npm run lint:ci`
- `npm run check:dist`
- `npm run test:release-promotion`
- Browser profile, release profile, release-promotion dry-run, and bundle-size checks.

---
## Summary by cubic
Adds lifecycle coverage for `setElement()` element replacement on `View` and `CollectionView`, and corrects the docs to match the actual behavior.

Previously the View lifecycle docs claimed a rendered View could never become unrendered while alive; `setElement()` can change that based on the replacement element's contents. The new tests specify that rendered and attached state are recomputed from the replacement element, repeated calls with the same element recompute the same state, and managed children and Region contents are preserved without moving child DOM. Runtime behavior is unchanged.

**Testing**
- Full Vitest suite: 70 files, 1,218 tests, 100% coverage.
- Performance, lint, docs, and bundle-size checks all pass; runtime size unchanged.

<sup>Written for commit 231c0d60194e81b95f1a355e87fb6e6fd98df3ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/214?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Clarified how replacing a live view’s element recalculates rendered and attached state.
  - Documented event redelegation and clarified that managed children and Region contents remain in their original DOM location.
  - Added guidance favoring reconstruction of a new view when appropriate.

- **Tests**
  - Added coverage for element replacement across detached, attached, empty, and populated states.
  - Verified repeated replacements are consistent and preserve child ownership and DOM placement.

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -140,6 +140,7 @@ attachment events and automatic child `isAttached()` updates.
 | --- | --- | --- |
 | Construct | Starts not rendered and not destroyed. It is attached only when its element is already in the document. | No children have been built. |
 | `render()` | Enters rendered and preserves its attached state. Repeated render stays rendered. | Builds and renders the current children. Repeated render destroys the previous children before building replacements. |
+| `setElement(el)` while alive | Preserves rendered state and recomputes attached state from whether the replacement element is in the document. Repeating the call with the same element recomputes the same state. | Existing managed child state and ownership are unchanged; child elements are not moved to the replacement element. |
 | A rendered collection resets | Remains rendered and preserves its attached state. | Destroys the previous children and builds replacements for the reset collection. |
 | `addChildView(view)` | Renders first when needed, then remains rendered. | Renders and manages the added View. |
 | `detachChildView(view)` | State is unchanged. | Removes and returns the live View in a detached state. The caller becomes responsible for it. |
--- a/docs/view.lifecycle.md
+++ b/docs/view.lifecycle.md
@@ -40,7 +40,7 @@ Returns a boolean value reflecting if the view has been destroyed.
 ### State vectors
 
 The three lifecycle methods are independent observations, not one linear state enum.
-Construction can therefore produce any of the four alive render/attachment vectors:
+`View` construction can therefore produce any of the four alive render/attachment vectors:
 
 | Initial `el` | `isRendered()` | `isAttached()` | `isDestroyed()` |
 | --- | --- | --- | --- |
@@ -49,12 +49,16 @@ Construction can therefore produce any of the four alive render/attachment vecto
 | Populated and detached | `true` | `false` | `false` |
 | Populated and in the document | `true` | `true` | `false` |
 
+`CollectionView` starts unrendered regardless of its initial contents and has its own
+[lifecycle transition table](./marionette.collectionview.md#view-lifecycle-and-events).
+
 With lifecycle monitoring enabled, Marionette-managed operations preserve the
 following observable transitions:
 
 | Operation | Result | Repeated call |
 | --- | --- | --- |
 | `view.render()` with a renderable template while alive | Rendered becomes `true`; attachment is unchanged | Renders again and runs the render lifecycle again |
+| `View#setElement(el)` while alive | Rendered reflects whether the replacement element has contents; attached reflects whether it is in the document | Recomputes the same state from the current element |
 | `region.show(view)` | Ensures the view is rendered; attached becomes `true` only when the Region is in the document | Showing the current view is a no-op |
 | `region.detachView()` | Rendered stays `true`; attached becomes `false`; destroyed stays `false` | Returns `undefined` with no transition |
 | Re-show a detached view | Rendered stays `true`; attachment reflects the Region | Does not render the view again |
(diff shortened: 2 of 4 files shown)


paulfalgout:
Exact-head Claude review requested for 69464e42e3ea17bb2534d2d14a98be57345e4e3b. Please verify the alive View versus CollectionView setElement state distinction, attached/detached and empty/populated transition coverage, repeated-call semantics, nested Region/managed-child ownership and DOM-location claims, documentation accuracy, and zero runtime/dist drift. Local exact-head validation passed: 1,218 tests at 100% coverage, 19 agent benchmark tests, 89 performance tests, lint, docs, reproducible dist, release dry-run checks, and unchanged runtime size at 51,878 / 51,975 bytes.

paulfalgout:
Reviewed at `69464e42` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, margin still 97 bytes.

**The most valuable thing here is that it falsifies an absolute claim the docs have been making.** The old note read *"Once a view is considered 'rendered' it cannot be unrendered until it is destroyed."* `setElement` recomputes `_isRendered` from `this.Dom.hasContents(this.el)`, so replacing a populated element with an empty one takes a live View from rendered back to not-rendered. The new matrix test covers all four combinations — `{false,false}`, `{true,false}`, `{false,true}`, `{true,true}` — which is the right shape, since rendered and attached are computed from independent properties (contents versus in-document) and only the full matrix shows they aren't coupled. Calling `setElement` twice and asserting the same state pins the idempotence claim as well.

**The child-DOM test is the other strong one**, for its negative half:

```js
expect(oldRoot.contains(child.el)).to.be.true;
expect(replacementRoot.contains(child.el)).to.be.false;
```

"Ownership preserved" alone would pass if children had been moved; asserting they are *not* in the replacement is what proves it. And it pins a genuinely surprising state — the parent reporting `attached: false` while its child still reports `attached: true`, because the child's element remains in the document inside the now-orphaned old root. That's a real inconsistency in the object graph, and characterizing it is exactly what this series is for.

**One asymmetry I'd raise.** The prose also now says *"View and Behavior DOM events are redelegated to the replacement element,"* correcting the old text's claim that *"the view's `behaviors` will also be unaware of the change."* I verified the correction is right rather than taking it — `View#setElement` at `modules/view-region.js:762` runs `_undelegateViewEvents()`, then `_setBehaviorElements()` (which maps `behavior.setElement()` across `this._behaviors`), then `_delegateViewEvents()`, and behavior events are aggregated into that delegation through `_getEvents()`. So the statement is accurate.

But no test in this PR or in `behavior-lifecycle.spec.js` exercises a Behavior across `setElement`. I've accepted mechanism-level backing elsewhere in this series and would normally leave it — the difference here is that the previous documentation asserted the *opposite*, and it was wrong for a long time without anything catching it. A regression would quietly restore that falsehood. A Behavior with a DOM trigger, a `setElement` to a fresh root, and an assertion that the trigger fires from the replacement would close it cheaply, and it's the claim in this PR most worth pinning precisely because it reverses a standing error.

paulfalgout:
Exact-head Claude re-review requested for 231c0d60194e81b95f1a355e87fb6e6fd98df3ef. The duplicate Cubic documentation blockers are fixed: the four construction vectors are explicitly View-specific, CollectionView links to its separate lifecycle table, and the contents-based row is now named View#setElement. Exact-head docs validation rebuilt 33 pages and checked 34 HTML files / 320 internal links; lint and diff checks pass. Runtime, tests, dist, and size are unchanged from the previously approved head.

paulfalgout:
Re-reviewed at `231c0d60` (the SHA I fetched). **Approve.** Docs-only, and it corrects an over-generalisation rather than just rewording.

The construction table said *"Construction can therefore produce any of the four alive render/attachment vectors"* without qualification, but that holds for `View` only. `CollectionView` treats rendering as "the child set has been built and inserted", so a `CollectionView` constructed on a populated element is *not* rendered — two of those four vectors are unreachable for it at construction. Scoping the table to `View` and pointing at the CollectionView contract for the difference is the right fix, and it's consistent with what that table already said in #197.

I checked the new cross-reference resolves rather than assuming `docs:check` had it covered: `## View Lifecycle and Events` is at line 124 of `marionette.collectionview.md`, which is exactly the `#view-lifecycle-and-events` anchor. Renaming the row label to `View#setElement(el)` is consistent with the same scoping, since `CollectionView#setElement` has its own row in the other table with different child semantics.

My note from the previous head still stands and is unaffected by this delta: the Behavior redelegation claim is accurate but untested, and it's the sentence in this PR that reverses a long-standing documented falsehood.

PR #215, opened 2026-08-28
feat(types): complete optional subpath declarations

paulfalgout:
Related #137

## What

- Add format-correct ESM and CommonJS declarations for `marionette/jquery-dom-api`, including a precise generic `JQuery<TElement>` return for `findEl`.
- Bind each runtime export condition to its matching declaration and add separately compiled strict NodeNext installed-package fixtures.
- Declare `@types/backbone` and `@types/jquery` as optional peers, align fixture versions with the supported runtime majors, and document the explicit TypeScript install contract.

## Why

The jQuery adapter had one unverified declaration whose `findEl` result was `any`. The new Backbone and jQuery subpath declarations also rely on external peer types that their runtime packages do not bundle; making those type peers explicit prevents fixtures from masking an undeclared consumer requirement while keeping JavaScript users unaffected.

## Scope

This changes optional-subpath declarations, export metadata, optional peer metadata, installation documentation, and installed-package type fixtures. Root Marionette declarations, Application/State contracts, runtime source, generated JavaScript, Rollup configuration, and performance contracts are intentionally unchanged.

## Deployment Impact

No application or website deployment is required. Optional type peers are not installed unless a consumer opts into the typed subpath. This does not publish npm, create a tag, or release v5.

## Testing

- Full Vitest coverage: 70 files, 1,218 tests, 100% statements/branches/functions/lines.
- `npm run test:fixtures`, including separate ESM and CommonJS TypeScript programs for both optional subpaths.
- `npm run lint:ci`
- `npm run check:dist`
- `npm audit --omit=dev` reported 0 vulnerabilities.
- Runtime artifacts stayed unchanged; cumulative size remains 51,878 / 51,975 bytes.

---
## Summary by cubic
Adds complete TypeScript declarations for the `marionette/jquery-dom-api` subpath, replacing the previous untyped `findEl` (which returned `any`) with a precise generic `JQuery<TElement>` return and format-correct ESM/CommonJS declaration files. Declares `@types/backbone` and `@types/jquery` as optional peers so TypeScript consumers know they must install those packages; JavaScript consumers are unaffected.

**Migration**

- TypeScript consumers importing `marionette/backbone` or `marionette/jquery-dom-api` must install the matching `@types/*` package as a dev dependency, since npm does not install optional peers.
- Adds strict NodeNext installed-package fixtures for both ESM and CommonJS to verify the new declarations resolve correctly.

<sup>Written for commit 839de3f6b3f8a677e567b3c63082646a820a0718. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/215?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/dist/jquery-dom-api.d.cts
+++ b/dist/jquery-dom-api.d.cts
@@ -1,0 +1,17 @@
+import 'jquery';
+
+declare const JQueryDomApi: {
+  findEl<TElement extends Element = HTMLElement>(
+    el: Element | Document,
+    selector: string,
+  ): JQuery<TElement>;
+  detachEl(el: Element): void;
+  setContents(el: Element, html: string): void;
+  appendContents(
+    el: Element | DocumentFragment,
+    contents: Element | DocumentFragment | string,
+  ): void;
+  detachContents(el: Element): void;
+};
+
+export = JQueryDomApi;
--- a/dist/jquery-dom-api.d.mts
+++ b/dist/jquery-dom-api.d.mts
@@ -1,0 +1,17 @@
+import 'jquery';
+
+declare const JQueryDomApi: {
+  findEl<TElement extends Element = HTMLElement>(
+    el: Element | Document,
+    selector: string,
+  ): JQuery<TElement>;
+  detachEl(el: Element): void;
+  setContents(el: Element, html: string): void;
+  appendContents(
+    el: Element | DocumentFragment,
+    contents: Element | DocumentFragment | string,
+  ): void;
+  detachContents(el: Element): void;
+};
+
+export default JQueryDomApi;
--- a/dist/jquery-dom-api.d.ts
+++ b/dist/jquery-dom-api.d.ts
@@ -1,9 +1,0 @@
-declare const JQueryDomApi: {
-  findEl(el: Element | Document, selector: string): any;
-  detachEl(el: Element): void;
-  setContents(el: Element, html: string): void;
-  appendContents(el: Element | DocumentFragment, contents: Element | DocumentFragment | string): void;
-  detachContents(el: Element): void;
-};
-
-export default JQueryDomApi;
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -39,7 +39,9 @@ required; the rest are optional.
 |---|---|---|
 | `underscore` `^1.13.0` | Required | Always. Marionette publishes named ESM imports from `underscore`, and versions before 1.13 are CJS-only with no package `exports` map, so modern Node and bundler ESM resolution cannot satisfy the imports. |
 | `backbone` `^1.4.0` | Optional | Only if your app uses Backbone models/collections, or imports the bundled `marionette/backbone` shim. See [Backbone is optional](#backbone-is-optional). |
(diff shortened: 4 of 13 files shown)


paulfalgout:
Exact-head Claude review requested for 839de3f6b3f8a677e567b3c63082646a820a0718. Please verify conditional ESM/CommonJS declaration fidelity, precise jQuery adapter types, isolated installed-package compilers, optional @types peer semantics and ranges, manifest/lock/docs consistency, JavaScript install neutrality, and zero runtime/dist-JS drift. Local exact-head validation passed: 1,218 tests at 100% coverage, 19 agent benchmark tests, full fixtures, 89 performance tests, lint, docs, reproducible dist, release dry-run checks, audit, and unchanged runtime size at 51,878 / 51,975 bytes.

paulfalgout:
Reviewed at `839de3f6` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, margin still 97 bytes.

**The declared surface matches the runtime exactly.** I compared them rather than assuming: `jquery-dom-api.js` defines `findEl`, `detachEl`, `setContents`, `appendContents`, `detachContents`, and the declaration lists those five and nothing else. That's the drift risk I raised on #213 not materialising here, though it's still hand-maintained.

**The fixture's negative assertions are the strongest part**, and `@ts-expect-error` is the right instrument:

```ts
// @ts-expect-error findEl returns a jQuery collection, not a DOM element.
const element: Element = JQueryDomApi.findEl(host, '.child');
// @ts-expect-error The adapter does not expose arbitrary jQuery methods.
JQueryDomApi.addClass('active');
```

These fail the compile if the error *stops* occurring, so loosening the declaration to `any` — the usual way type coverage rots — would turn all three into unused-expect-error failures. That's the type-level equivalent of the rejection tests worth having, and it's the property that makes this fixture able to fail rather than merely able to pass.

**Replacing the top-level `"types"` with per-condition entries is a real fix, not just consistency.** The old form pointed both ESM and CJS consumers at a single `.d.ts`, which under `NodeNext` means CommonJS consumers were typed by a file whose module shape didn't match how they resolve it. Splitting into `.d.mts`/`.d.cts` fixes that, and the old `dist/jquery-dom-api.d.ts` is properly removed rather than left orphaned.

Aligning `@types/jquery` to `3.5.34` from `4.0.1` is also right — the `jquery` peer range is `^3.5.0`, so typing against the 4.x line described a different runtime than the one the package supports.

**One thing worth being deliberate about, not a defect here.** After this PR the two *optional* subpaths are fully typed while the root `.` export has no `types` condition at all — `import { View, CollectionView, Region } from 'marionette'` is untyped. That's an inversion a consumer won't expect: seeing `.d.mts` files in `dist/` reads as "this package ships types", and then the main API is `any`. The roadmap puts complete declarations in Phase 2, so this is sequencing rather than omission — but until the root is typed, a line in the docs saying so would save someone the discovery.

PR #216, opened 2026-08-28
test(lifecycle): specify nested attachment propagation

paulfalgout:
Related #131

## What

- Add one canonical public-API lifecycle sequence for a detached rendered parent View and its existing Region child.
- Characterize attachment propagation, a second child shown synchronously during parent `onAttach`, repeated same-parent show, detach/re-show, and repeated final teardown.
- Document the corresponding nested Region-managed attachment and detachment contract.

## Why

The behavior existed in implementation and scattered legacy callback tests, but agents still had to infer the complete nested state transition sequence. This makes the detached-parent, selected reentrant, repeated-call, and teardown behavior executable in the canonical lifecycle contract.

## Scope

This changes only the View lifecycle test and documentation. It intentionally excludes pre-render `showChildView`, post-destroy operations, Application/State lifecycle design, runtime modules, generated artifacts, package metadata, and performance contracts.

## Deployment Impact

## Testing

- Full Vitest coverage: 70 files, 1,219 tests, 100% statements/branches/functions/lines.
- `npm run lint:ci`
- `npm run check:dist`
- `npm run test:fixtures`

---
## Summary by cubic
Specifies the nested attachment propagation lifecycle for rendered-but-detached parent Views and their existing Region children, covering reentrant `showChildView` during parent `onAttach`, repeated same-parent shows, detach/re-show, and repeated teardown. This makes the previously inferred behavior executable and documented, satisfying the contract requirements in #131.

No behavior, runtime, or deployment changes; the diff adds only a canonical lifecycle test and two doc paragraphs.

<sup>Written for commit 88964c114811d85ea6509ceb2eba4be3db8f2fd6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/view.lifecycle.md
+++ b/docs/view.lifecycle.md
@@ -221,6 +221,12 @@ The attached state is maintained when attaching a view with a `Region` or as a c
 or during [view instantiation](#instantiating-a-view).
 If a view is attached by other means like `$.append` [`isAttached`] may not reflect the actual state of attachment.
 
+A child shown in a rendered but detached parent View's Region is rendered and remains
+detached. When the parent is later shown in an attached Region, attachment propagates
+to its existing children. A child shown during the parent's `onAttach` is attached
+immediately. Showing the same attached parent again is a no-op for both parent and child
+attachment lifecycles.
+
 ## Detaching a View
 
 A view is detached when its `el` is removed from the DOM.
@@ -229,6 +235,11 @@ The best time to clean up any listeners added to the `el` is in the [`before:det
 While the `el` of the view may remain attached, its contents will be removed on render.
 If you have added listeners to the contents of the view rather than `before:detach` the
 [`dom:remove` event](./events.class.md#domremove-event) would be best.
+
+Detaching a parent View propagates detachment to its managed Region children while
+preserving their rendered state and ownership. Re-showing that parent attaches the same
+children again. Emptying the parent-owning Region then detaches and destroys the parent
+and its still-managed children once.
 
 ## Destroying a View
 
--- a/test/unit/view-lifecycle.spec.js
+++ b/test/unit/view-lifecycle.spec.js
@@ -136,6 +136,96 @@ describe('View lifecycle contract', function() {
     view.destroy();
   });
 
+  it('propagates nested attachment through detached, reentrant, and repeated transitions', function() {
+    this.setFixtures('<div id="nested-lifecycle-region"></div>');
+    const existingChild = new View({ template: () => '<span>Existing child</span>' });
+    const reentrantChild = new View({ template: () => '<span>Reentrant child</span>' });
+    let firstReentrantStateDuringAttach;
+    const ParentView = View.extend({
+      template: () => `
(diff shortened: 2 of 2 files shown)


paulfalgout:
Exact-head Claude review requested for 88964c114811d85ea6509ceb2eba4be3db8f2fd6. Please verify public-API-only lifecycle assertions, detached-parent and nested child state transitions, synchronous onAttach child attachment, repeated show/detach/re-show/empty event counts, documentation scope, and zero runtime/dist drift. Local exact-head validation passed: 1,219 tests at 100% coverage, 19 agent benchmark tests, full fixtures, 89 performance tests, lint, docs, reproducible dist, release dry-run checks, and unchanged runtime size at 51,878 / 51,975 bytes.

paulfalgout:
Reviewed at `88964c11` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, margin still 97 bytes.

**The reentrancy assertion is the best thing in this series so far.** Capturing the child's state *inside* the parent's `onAttach` rather than after it:

```js
onAttach() {
  this.showChildView('reentrant', reentrantChild);
  if (!firstReentrantStateDuringAttach) {
    firstReentrantStateDuringAttach = state(reentrantChild);
  }
}
…
expect(firstReentrantStateDuringAttach).to.deep.equal({ rendered: true, attached: true, destroyed: false });
```

That's what distinguishes "attached immediately" from "attached by the time we checked." I traced why it holds: `Region._attachView` sets `view._isAttached = true` *before* `triggerMethod('attach')`, so a child shown during the parent's attach handler finds the parent's element already in the document and attaches in the same pass. If that ordering were ever inverted, this assertion records `attached: false` and fails — which is exactly the property worth pinning, and it's the same before-or-after-the-flag question that `isDestroyed()` during `before:empty` raised on #195.

**Every propagation claim is pinned by call count rather than end state**, which is what makes the no-op and once-only claims meaningful:

- re-showing an attached parent leaves `attach` at `calledOnce` — a state assertion alone would pass whether or not the lifecycle re-ran
- `detach` reaches `calledOnce` on all three, including the reentrant child, so propagation covers a child added during a previous attach
- two `empty()` calls yield `detach` twice and `destroy` **once**, so the second genuinely did nothing

Worth noting the spies also do double duty as an ownership check. After `detachView()` the assertions cover state but not `getChildView` identity — that would normally be a small gap, except the subsequent re-show driving `attach` to `calledTwice` on *those same spy objects* can only happen if the original child instances are still the ones being managed. Fresh children would carry fresh spies and leave those counts unchanged.

Nothing open from me.

PR #217, opened 2026-08-28
test(behavior): specify nested removal ownership

paulfalgout:
Related #133

## What

- Add a lifecycle contract proving that directly destroying a declaring Behavior removes only that instance.
- Verify its nested Behavior remains active on the shared host and is later destroyed exactly once by host teardown.
- Document Behavior nesting as declaration grouping rather than parent-owned cleanup.

## Why

Nested Behavior instances are flattened into the host-owned Behavior collection, but the public docs and tests did not state what happens when the declaring Behavior is removed directly. Agents should not infer cleanup ownership from declaration nesting.

## Scope

This changes only Behavior lifecycle tests and documentation. It intentionally leaves pre-rendered-host UI representation, repeated direct cleanup/reuse, dynamic Behavior replacement, root declarations, runtime modules, generated artifacts, package metadata, and performance contracts unchanged.

## Deployment Impact

## Testing

- Full Vitest coverage: 70 files, 1,220 tests, 100% statements/branches/functions/lines.
- `npm run lint:ci`
- `npm run check:dist`
- `npm run test:fixtures`

---
## Summary by cubic
Addresses #133 by clarifying that nested Behaviors belong to the host view, not to the declaring Behavior. Previously cleanup ownership was unspecified; now destroying the declaring Behavior directly leaves nested Behaviors active until removed directly or the host is destroyed.

**Changes**
- Updates docs to state that nesting groups declarations rather than transferring cleanup ownership.
- Adds a lifecycle test proving the nested Behavior survives direct declarer removal and is destroyed exactly once at host teardown.
- Limits changes to tests and documentation; no production behavior or bundle size changes.

<sup>Written for commit 1b9791adeaa734f547deda2eae7d875bd74aa2ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -236,8 +236,10 @@ const Modal = Behavior.extend({
 
 [Live example](https://jsfiddle.net/marionettejs/7ffnqff3/)
 
-Nested Behaviors act as if they were direct Behaviors of the parent `Behavior`'s
-view instance.
+Nesting groups Behavior declarations; it does not transfer cleanup ownership to
+the declaring Behavior. Nested Behaviors act as direct Behaviors of the same host
+view, so destroying the declarer leaves them active until they are removed
+directly or the host is destroyed.
 
 ## The Behavior's `view`
 The `view` is a reference to the `View` instance that the `Behavior` is attached to.
--- a/test/unit/behavior-lifecycle.spec.js
+++ b/test/unit/behavior-lifecycle.spec.js
@@ -186,6 +186,58 @@ describe('Behavior lifecycle contract', function() {
     expect(behavior.stopListening).to.have.been.calledOnce;
   });
 
+  it('keeps a nested Behavior host-owned after directly removing its declarer', function() {
+    const parentHostEvent = this.sinon.spy();
+    const nestedHostEvent = this.sinon.spy();
+    const parentBeforeDestroy = this.sinon.spy();
+    const nestedBeforeDestroy = this.sinon.spy();
+    const parentDestroy = this.sinon.spy();
+    const nestedDestroy = this.sinon.spy();
+    let parentBehavior;
+    let nestedBehavior;
+
+    const NestedBehavior = Behavior.extend({
+      initialize() {
+        nestedBehavior = this;
+      },
+      onHostEvent: nestedHostEvent,
+      onBeforeDestroy: nestedBeforeDestroy,
+      onDestroy: nestedDestroy,
+    });
+    const ParentBehavior = Behavior.extend({
+      behaviors: [NestedBehavior],
+      initialize() {
+        parentBehavior = this;
+      },
+      onHostEvent: parentHostEvent,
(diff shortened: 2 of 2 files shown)


paulfalgout:
Exact-head Claude review requested for 1b9791adeaa734f547deda2eae7d875bd74aa2ab. Please verify flattened host ownership semantics, direct declarer removal, continued nested host-event participation, exact host teardown counts, public observability, documentation scope, and zero runtime/dist drift. Local exact-head validation passed: 1,220 tests at 100% coverage, 19 agent benchmark tests, full fixtures, 89 performance tests, lint, docs, reproducible dist, release dry-run checks, and unchanged runtime size at 51,878 / 51,975 bytes.

paulfalgout:
Reviewed at `1b9791ad` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, margin still 97 bytes.

**This corrects a genuinely misleading sentence.** The old text — *"Nested Behaviors act as if they were direct Behaviors of the parent `Behavior`'s view instance"* — reads as though nesting establishes a parent relationship, which invites exactly the wrong inference: that destroying the declarer takes its nested Behaviors with it. Saying plainly that nesting *groups declarations* and does not transfer cleanup ownership, then naming the consequence, is the difference between a sentence someone can act on and one that quietly misleads.

**The assertion that carries it is the count after the declarer is destroyed:**

```js
parentBehavior.destroy();
view.triggerMethod('host:event');
expect(parentHostEvent).to.have.been.calledOnce;    // stopped
expect(nestedHostEvent).to.have.been.calledTwice;   // still live
```

A state-oriented test would show the declarer destroyed and stop there; showing the nested Behavior still *receiving host events afterwards* is what makes the footgun concrete.

**The test also pins something a little beyond the doc sentence, which I think is worth keeping.** After `view.destroy()`, the parent's `onBeforeDestroy` and `onDestroy` are asserted never to have been called at all. So a direct `destroy()` doesn't merely deactivate a Behavior — it removes it from the host's cleanup set entirely, and the host later destroys only what it still owns. Combined with `nestedDestroy` at `calledOnce` across two `view.destroy()` calls, that covers both the removal and the idempotence in one pass.

`calledOn(nestedBehavior)` throughout keeps the receiver pinned rather than just the call, which matters here because both Behaviors declare identically-named handlers and a mis-bound context would otherwise look like a pass. Consistent with what #198 established about nested Behaviors being flattened onto the same host — `nestedBehavior.view === view` is re-asserted here too.

Nothing open from me.

Issue #139, opened 2026-08-26
v5: make canonical documentation examples executable


paulfalgout:
## Goal
Choose one canonical pattern for common Marionette tasks, publish counterexamples, and verify examples against the shipped package.

## Dependencies
Depends on the settled public lifecycle, ownership, Behavior, diagnostic, and package-entrypoint contracts exercised by each example. Complete declarations and generated API metadata in #137, architecture lint rules in #138, and the final reference sweep in #147 are separate downstream gates; when they change a documented contract, update the canonical example and fixture in the same change.

## Agent-development failure addressed
Conflicting or stale examples cause agents to reproduce obsolete patterns.

## Public behavior contract
Examples must match Phase 1 contracts and supported entrypoints.

## Runtime cost category
Documentation/test only.

## Acceptance criteria
- [x] Inventory duplicated or conflicting patterns.
- [x] Define View, Behavior, child, Region, and external-shell DOM ownership boundaries, absorbing the scope of #76.
- [x] Select canonical lifecycle, Region, Behavior, communication, and cleanup examples.
- [x] Identify cross-boundary DOM traversal and other common mistakes as counterexamples.
- [x] Remove or deprecate superseded guidance.
- [x] Execute examples in CI or validate them through an equivalent drift check.
- [x] Link relevant diagnostic and lint rule codes.

## Rollback or deprecation condition
When a pattern changes before stable, update the canonical example and remove the superseded path in the same change.

paulfalgout:
Progress: #218 merged as cc05bf38e2855fb39fbb5d54b9837065b509d1fa.

This adds an executable packed-package contract for the documented Behavior host-communication pattern. The fixture extracts the marked documentation example verbatim and verifies host-scoped events, a single host request with the View payload, and event cleanup after destroy. The documentation also links invalid host communication and invalid Behavior UI configuration to MN0018/MN0019.

This is a zero-runtime-delta slice. Issue #139 remains open for the broader executable documentation and diagnostic coverage still outstanding.

paulfalgout:
Further progress: #220 merged as 56a794bbf42f7b800e1d6a25da9b7b8cb7979729 and #221 merged as f37830d8c4e42a5d9d1a30b0c9e71a46f7da5c28.

[...]

PR #218, opened 2026-08-28
test(docs): execute Behavior host example

paulfalgout:
Related #139

## What

- Add a marked Behavior host-communication example that is extracted verbatim from the canonical documentation.
- Execute that example against the packed Marionette package with pinned jsdom and Underscore dependencies.
- Verify host DOM scope, exactly-once host requests, request payload, teardown cleanup, and link invalid UI/handler literals to MN0018/MN0019.

## Why

Canonical examples should be executable rather than prose agents must trust. This turns one already-settled Behavior composition pattern into an installed-package contract without duplicating the example in test code.

## Scope

This changes only Behavior documentation and package-fixture coverage. It intentionally avoids the broader DOM interactions rewrite, root declarations, runtime modules, distribution artifacts, export metadata, workflows, and performance contracts.

## Deployment Impact

## Testing

- Full Vitest coverage: 70 files, 1,220 tests, 100% statements/branches/functions/lines.
- `npm run test:fixtures`, including extraction and execution of the exact documented fence against the packed package.
- `npm run lint:ci`
- `npm run check:dist`

---
## Summary by cubic
Turns the documented Behavior host-communication example into an executable fixture so the canonical docs example is verified against the packed package rather than trusted as prose.

- Extracts the code fence verbatim from `docs/marionette.behavior.md` and runs it in `test/fixtures/docs-behavior-host`.
- Verifies host-scoped DOM delegation, exactly-once requests, the host View as the request argument, and teardown cleanup.
- Pins `jsdom` 30.0.1 and `underscore` 1.13.8 in the fixture.
- Invalid `@ui` references and non-callable handlers now link to MN0018 and MN0019 in the docs.

<sup>Written for commit 2173f8689dd7c4d943b3c47229e1b102a7f793cd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
Exact-head Claude review requested for 128b614790359b867e7e5ca45b28995c5f3c9ec3. Please verify verbatim fence extraction, installed-package resolution and cleanup, jsdom global isolation, host-scoped DOM behavior, exactly-once semantic request/payload, destroy undelegation, diagnostic links, pinned fixture dependencies, and zero runtime/dist drift. Local exact-head validation passed: 1,220 tests at 100% coverage, 19 agent benchmark tests, full fixtures, 89 performance tests, lint, docs, reproducible dist, release dry-run checks, and unchanged runtime size at 51,878 / 51,975 bytes.

paulfalgout:
Reviewed at `128b6147` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, margin still 97 bytes.

**Extracting the code from the markdown is what makes this real**, and it's worth saying why the obvious alternative fails. The usual "executable example" is a copy of the snippet living in a test file, which proves the *copy* works while the documented version drifts away unnoticed. Here the doc is the single source:

```js
const codeFence = markedContent.match(/^\s*```javascript\n([\s\S]*?)\n```/);
await writeFile(examplePath, codeFence[1], 'utf8');
const { FormView } = await import(pathToFileURL(examplePath));
```

Edit the fence and the test runs the edit. Drift isn't guarded against, it's impossible.

The two guards around the extraction matter as much as the extraction. Asserting exactly one marker means adding a second executable example fails loudly instead of silently testing the first, and requiring the fence to immediately follow the marker means a doc reshuffle that separates them is caught rather than quietly matching some later block.

**The assertions pin the claims the prose actually makes.** The JSDOM deliberately places a matching `.save` button *outside* the host, and `requestCount === 0` after clicking it is what turns "delegation stays scoped to the host View" into something falsifiable. And the post-destroy check clicks the **retained** `hostButton` reference rather than re-querying:

```js
view.destroy();
hostButton.click();
assert.equal(requestCount, 1, 'destroy must undelegate the retained host button');
```

Re-querying after destroy could return null and pass vacuously; holding the reference proves the listener was removed rather than the element merely being gone.

Worth noting this runs against the packed tarball installed by `run.mjs`, so it exercises the shipped `exports` map and built artifacts rather than source — which is what the roadmap's "examples are run in CI or otherwise verified against the shipped package" actually asks for.

**One housekeeping item.** `validate.mjs` writes `test/fixtures/docs-behavior-host/dist/example.mjs`, and `git check-ignore` confirms nothing ignores it — `.gitignore` has no `dist` rule, correctly, since built `dist/` is tracked. So after a local test run that generated file sits untracked in the tree. I checked whether it could simply be written to a temp directory instead, and it can't: the fixture's `package.json` doesn't depend on `marionette`, which `run.mjs` installs into the fixture's own `node_modules`, so the file has to live there for `import { Behavior, View } from 'marionette'` to resolve. An ignore entry for that path is the fix. The risk is specific rather than cosmetic — a copy committed by an `add -A` becomes a second, stale, authoritative-looking version of the example, which is exactly what this design eliminates.

**Smaller, and take it or leave it.** The prose says an undeclared `@ui` reference throws MN0018 and an unresolvable string handler throws MN0019, framed as properties of *this* example. The fixture runs only the valid variant. Those codes have their own unit coverage from #208 so the claims aren't unbacked, but the extraction machinery is already here — two marked invalid variants asserting the codes would make the whole paragraph executable rather than half of it.

paulfalgout:
Exact-head Claude re-review requested for 2173f8689dd7c4d943b3c47229e1b102a7f793cd. The housekeeping finding is fixed with a fixture-local .gitignore for only the generated dist directory, so direct validation cannot expose the extracted example as a stale second source. The full installed-package fixture suite, lint, and diff checks pass on this head; runtime, dist, and size remain unchanged.

paulfalgout:
Re-reviewed at `2173f868` (the SHA I fetched). **Approve — nothing open from me.**

Scoping the rule to a `.gitignore` inside the fixture directory is better than a path entry in the root file: `dist/` there can only ever match `test/fixtures/docs-behavior-host/dist/`, so there's no way for it to interact with the tracked root `dist/` that holds the shipped artifacts. A root-level `test/fixtures/**/dist/` rule would have worked too but reads as something that needs checking against the artifact directory every time someone edits it; this one doesn't.

That closes the drift risk I raised — the generated `example.mjs` can no longer be picked up by an `add -A` and become a second copy of the example competing with the doc.

Issue #132, opened 2026-08-26
v5: expose ownership and make topology queries side-effect free


paulfalgout:
## Goal
Expose ownership facts Marionette already maintains, add public nested Application topology from #190, and make Region and topology queries pure without breaking mutating convenience operations.

## Dependencies
Requires lifecycle states (#131), diagnostic semantics (#134), and the support profile (#153).

## Agent-development failure addressed
Agents must currently read private fields, and Region lookup may render a View as a query side effect.

## Public behavior contract
Define owner, name, current child, Application parent/name/children/root/host/current root View, unknown Region, unrendered Region, empty Region, missing child, and destroyed-object semantics. `getRegion`, `getRegions`, `hasRegion`, and topology reads must not render or mutate.

## Runtime cost category
Existing production path. Reuse stored relationships without new per-instance data when possible.

## Acceptance criteria
- [x] Add minimal public read-only ownership accessors for rendered topology and Application composition without duplicating mutable registries.
- [x] Make `getRegion`, `getRegions`, `hasRegion`, and topology reads side-effect free.
- [x] Explicitly specify `showChildView`, `detachChildView`, `getChildView`, and `emptyRegions` when called before render: each must render intentionally or emit the documented diagnostic rather than relying on query side effects.
- [x] Update the current documentation that describes query-triggered rendering.
- [x] Cover unknown, invalid, unrendered, empty, and destroyed states with stable diagnostics.
- [x] Add contract tests and canonical docs.
- [x] Measure bundle and hot-path impact.

## Rollback or deprecation condition
Do not retain both mutating and pure query paths. Any compatibility exception needs a verified consumer and removal condition.

paulfalgout:
Progress: #219 merged as e46f9af008a9a0e0f1f22787f3537a7f4970577b.

This pins the existing View-owned Region teardown topology using only public APIs: direct Region destruction destroys and clears its current View once, removes the Region from the owning View's named lookups, and prevents repeated Region or later owner destruction from repeating cleanup. The documentation now states the same contract.

This is a zero-runtime-delta characterization slice. Issue #132 remains open for pure unrendered Region queries, public ownership accessors, Application topology, and the associated diagnostics/migration decisions.

paulfalgout:
Completed across focused merged slices:

- #219 characterized owned Region teardown through public topology.
- #257 replaced incidental missing-Region failures with stable diagnostics.
- #302 defined destroyed Region behavior.
[...]

PR #219, opened 2026-08-28
test(region): cover owned region teardown

paulfalgout:
Related #132

## What

- Add a public-API contract test for directly destroying a View-owned Region.
- Document removal from the owning View lookup and exactly-once child/Region teardown.

## Why

Marionette already removes a directly destroyed named Region from its owner, but that topology transition was split across implementation details and separate tests. Agents need the observable ownership and teardown contract pinned before the broader pure-query work in #132.

A caller could not determine from one public contract whether direct Region destruction leaves stale owner topology or whether later owner destruction repeats cleanup.

## Public behavior contract

- Canonical behavior after this PR: directly destroying an owned Region destroys its current View once, clears Region state, removes the named Region from public owner lookups, and makes repeated Region or owner destruction a no-op for that teardown.
- Superseded behavior removed: none; this characterizes existing behavior.

## Scope

- Affects Region lifecycle tests and Region documentation only.
- Does not change runtime code, Application topology, unrendered lookup behavior, or destroyed-Region reuse semantics.

## Runtime cost

- Static or documentation only.
- Bundle: zero runtime artifact delta.
- Allocations/retention: unchanged.

## Deployment Impact

No application redeploy or package publication is needed. This PR changes repository tests and documentation only.

## Testing

- Narrow Region lifecycle test: 8 tests passed.
- npm run coverage: 70 files, 1,221 tests, 100% statements/branches/functions/lines; agent benchmark contract 19/19.
- npm run lint:ci.
- npm run check:dist.
- git diff --check.
- Confirmed no changes under runtime, build, dist, package, or configuration paths.

## Package and browser impact

- Browser contracts affected: none; the test characterizes existing public lifecycle behavior.

## Rollback or deprecation

Revert the test/documentation commit if the broader #132 design intentionally changes this ownership contract; no runtime rollback is required.

## Review notes

This is deliberately a partial #132 slice. Pure unrendered Region queries and Application topology remain out of scope.

---
## Summary by cubic
Pins the public contract for directly destroying a View-owned Region, supporting the pure-query work in #132. Directly destroying an owned Region destroys its current View once, clears Region state, unlinks the named Region from the owner's lookups, and makes repeated Region or owner destruction a no-op for that teardown.

**Scope**
- Changes tests and `docs/marionette.region.md` only; no runtime code changes.
- Does not touch Application topology, unrendered lookup behavior, or destroyed-Region reuse semantics.

<sup>Written for commit 36ba9b2b6d88f81c78c4060276c0cc42a2088087. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/219?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Clarified that destroying a Region also destroys its current View and removes it from its parent View’s Region lookups.
  * Documented that destruction is safe to repeat and that destroyed Regions should not be reused.

* **Tests**
  * Added coverage for Region and child View teardown, owner unlinking, cleared state, and one-time lifecycle callbacks.

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -507,8 +507,10 @@ This can be useful in unit testing your views.
 
 ## `destroy` A Region
 
-A region can be destroyed which will `reset` the region, remove it from any parent view,
-and stop any internal region listeners. A destroyed region should not be reused.
+A region can be destroyed which will `reset` the region, destroy its current View,
+remove it from any parent View's Region lookups, and stop any internal Region listeners.
+Repeated Region destruction or later destruction of the parent View does not repeat
+the child or Region teardown. A destroyed Region should not be reused.
 
 ```javascript
 import { View } from 'backbone.marionette';
--- a/test/unit/region-lifecycle.spec.js
+++ b/test/unit/region-lifecycle.spec.js
@@ -211,4 +211,60 @@ describe('Region lifecycle contract', function() {
     expect(region.hasView()).to.be.false;
     expect(region.currentView).to.be.undefined;
   });
+
+  it('unlinks a directly destroyed owned Region without repeating teardown', function() {
+    const owner = new View({
+      regions: {
+        content: '.content',
+      },
+      template() {
+        return '<div class="content"></div>';
+      },
+    });
+    const child = new TestView();
+    const regionLifecycle = {
+      beforeDestroy: this.sinon.spy(),
+      beforeEmpty: this.sinon.spy(),
+      empty: this.sinon.spy(),
+      destroy: this.sinon.spy(),
+    };
+    const childLifecycle = {
+      beforeDestroy: this.sinon.spy(),
+      destroy: this.sinon.spy(),
+    };
+
+    owner.render();
+    const ownedRegion = owner.getRegion('content');
+    ownedRegion.on('before:destroy', regionLifecycle.beforeDestroy);
+    ownedRegion.on('before:empty', regionLifecycle.beforeEmpty);
+    ownedRegion.on('empty', regionLifecycle.empty);
+    ownedRegion.on('destroy', regionLifecycle.destroy);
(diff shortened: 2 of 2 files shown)


paulfalgout:
Reviewed at `36ba9b2b` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, margin still 97 bytes.

**The doc correction adds a step that was genuinely missing.** The old text said destroy would *"`reset` the region, remove it from any parent view, and stop any internal region listeners"* — `reset` is vague about the child's fate, and a reader could reasonably conclude the current View is merely detached. Stating that it destroys the current View, and narrowing "remove it from any parent view" to "remove it from any parent View's Region lookups", both describe what actually happens rather than gesturing at it.

**The final assertion is the efficient part.** Four teardown calls across two objects, then every one of six lifecycle spies at `calledOnce`:

```js
ownedRegion.destroy(); ownedRegion.destroy();
owner.destroy();       owner.destroy();
// → region before:destroy / before:empty / empty / destroy, child before:destroy / destroy — all once
```

That covers both repetition axes at the same time: repeated Region destroy is idempotent, *and* the owner's later destroy doesn't re-tear-down a Region it no longer holds. Testing those separately would take two tests and still miss the interaction between them.

`expect(owner.getRegions()).not.to.have.own.property('content')` is the right assertion shape rather than a plain property check, given the snapshot is built through `defineProperty` since #210 — a `not.to.have.property` would also pass for a prototype-inherited name, which is exactly the distinction that work introduced.

Nothing open from me.

PR #220, opened 2026-08-28
test(docs): execute Radio owner example

paulfalgout:
Related #139

## What

- Make the canonical Radio-backed MnObject example executable and correct its message-count behavior.
- Add a packed-package fixture for shared channel identity, event/request ownership cleanup, and named reset semantics.
- Link the MN0021 and MN0017 reset diagnostics from the canonical documentation.

## Why

The existing example triggered a message without recording it, so its request still returned zero, and its cleanup claim was not executed against the shipped package. Agents need one verified Radio ownership pattern whose example, diagnostics, and runtime behavior agree.

The documentation could teach a superficially complete integration that neither changed state nor proved which event and request handlers were removed when the owning Marionette object was destroyed.

## Public behavior contract

- Canonical behavior after this PR: a named Radio channel is shared with its MnObject owner; owner destruction removes that owner event and request handling; a known named reset clears handlers without replacing the channel; invalid named resets fail with MN0021 or MN0017 without changing existing channels.
- Superseded behavior removed: the no-op showMessage example is replaced with the working canonical pattern.

## Scope

- Affects Radio documentation and a packed-package documentation fixture only.
- Does not change Radio, MnObject, runtime entrypoints, package exports, or the broader #139 documentation inventory.

## Runtime cost

- Development or test only; excluded from production entrypoints.
- Bundle: unchanged at 51.88/51.98 kB totals.
- Allocations/retention: unchanged in production.

## Deployment Impact

No application redeploy or package publication is needed. This PR changes repository documentation and tests only.

## Testing

- Test-first validator failed on the missing executable marker before the documentation change.
- Focused fresh packed-package fixture passed.
- npm run test:fixtures.
- npm run lint:ci.
- npm run check:dist.
- npm run size: unchanged 51.88/51.98 kB totals.
- git diff --check.

## Package and browser impact

- Entry points or install fixtures affected: adds one package-consumer fixture using the public marionette entrypoint.
- Browser contracts affected: none; no DOM environment is required.
- Production/dev/test module-graph impact: test fixture only.

## Rollback or deprecation

Revert the documentation fixture and example together if the canonical Radio ownership pattern changes before stable v5.

## Review notes

This is a partial #139 slice and does not close the issue. The fixture uses exact Backbone 1.4.0 and Underscore 1.13.8 dependencies and no jsdom.

---
## Summary by cubic
Fixes the Radio owner doc example so it records and counts messages instead of being a no-op, and links the MN0021/MN0017 reset diagnostics. Adds a packed-package fixture that executes the example to verify destroy and named reset behavior.

- The fixture asserts owner destroy removes event and request handlers, named reset clears handlers while preserving channel identity, and invalid resets throw MN0021/MN0017 without changing existing channels.
- This is a partial slice of #139 and does not close it.
- No production code changes; the example is now exported and marked executable and runs via the fixture runner.

<sup>Written for commit 7da5bb6788bff0a335d15b4b8a7fe83232ae936d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/220?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
Reviewed at `a73b7825` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, margin still 97 bytes. The #218 pattern is applied faithfully, `.gitignore` included from the start.

**This is the first time #211's Radio semantics run against the shipped package.** The falsy-name loop is the assertion I'd most want carried out of unit tests and into a packed-tarball fixture:

```js
for (const name of ['', null, false, 0, undefined]) {
  assert.throws(() => Radio.reset(name), error => error.code === 'MN0017');
}
channel.trigger('unrelated');
assert.equal(unrelatedCount, 1, 'invalid reset calls must not change existing channels');
```

The trailing check is what pins the destructive half — before #211 those falsy values took the reset-*all* branch, and only the surviving handler count would have caught it. Verifying that through the built artifact rather than source closes the gap between "the module behaves" and "the package behaves."

The named-reset pair is precise in the same way: channel identity preserved *and* handlers cleared, asserted separately, which is exactly what the table claims.

**One tension worth naming now that this is the second executable example, rather than after a dozen more.** Making the snippet importable required adding five `export`s, and two of them exist only for the test:

```js
export const message = { text: 'Hello' };
export const count = channel.request('message:count');
```

Exporting a fixture object and a module-scope computed value isn't something a reader would write, so the doc has drifted slightly toward test-shaped code. The other three (`Notifications`, `notifications`, `channel`) read naturally enough.

Both could come out at no cost to coverage: the trigger line still executes on import, so the fixture can read `notifications.messages[0]` instead of importing `message`, and call `channel.request('message:count')` itself instead of importing `count`. The doc keeps `const message` / `const count`, the assertions stay identical, and the example reads like application code again.

Not blocking, and I'd rather flag the pattern's cost than the instance — the extraction approach is clearly worth keeping, and it's cheaper to settle the convention on example two than on example twelve.

paulfalgout:
Re-reviewed at `bdf9b5e9` (the SHA I fetched). **Approve.** The example reads as application code again — the three remaining exports are all things someone would plausibly write, and `message` and `count` are ordinary locals.

**One consequence of my own suggestion that I should point out**, since I checked what it did rather than just that the diff matched:

```js
const message = notifications.messages[0];
…
assert.deepEqual(notifications.messages, [message], 'the event must reach its owner once');
```

That comparison is now partly self-referential — it establishes that exactly one message arrived, but no longer that the arrived object is the one the example triggered with, since `message` is derived from the array being checked.

In practice the coverage holds up, for two reasons worth stating rather than assuming. `assert.equal(count, 1)` independently proves one message was recorded, via a completely different path (the request handler counting `messages.length`). And the *second* use is unaffected — `message` is captured before `notifications.destroy()`, so comparing after the post-destroy trigger still proves no new entry was appended and the original is untouched. That assertion carries the same weight it did before.

If you want the length claim stated rather than implied, `assert.equal(notifications.messages.length, 1)` alongside would do it. I'd avoid comparing against a literal `{ text: 'Hello' }` — that would put a copy of the doc's content back in the fixture, which is the drift the extraction design exists to prevent.

Worth noting #221 arrived at a cleaner form of the same idea by exporting a function instead of values, which sidesteps this trade entirely; that's the shape I'd reach for on the next one.

paulfalgout:
Re-reviewed at `7da5bb67` (the SHA I fetched). **Approval stands — no new review substance.** This head is the previously-approved work plus a rebase onto `e46f9af0`, which carries #219.

Verified rather than assumed: `git diff bdf9b5e9 7da5bb67 -- docs/radio.md test/fixtures/docs-radio-owner/` is empty, so this PR's own content is byte-identical across the rebase, and its delta against master is only its five files.

PR #222, opened 2026-08-29
test(view): cover reentrant destruction

paulfalgout:
Related #131

## What

- Add a public lifecycle contract test for reentrant View destruction.
- Document that destroy calls from before:destroy, destroy, and after completion return the same View without restarting teardown.
- Verify parent, managed child, and owning Region cleanup each complete exactly once.

## Why

The runtime already guards reentrant destruction, but the behavior was only visible in implementation. #131 requires selected reentrant lifecycle cases to be explicit so agents do not infer whether nested destroy calls repeat detach, child destruction, or Region emptying.

A lifecycle handler could reasonably call destroy again without any public test or documentation explaining whether that restarts teardown, duplicates events, or corrupts the owning Region.

## Public behavior contract

- Canonical behavior after this PR: once View destruction starts, reentrant and later repeated destroy calls return the same View and do not restart teardown; parent, child, and owning Region lifecycles each complete once.
- Superseded behavior removed: none; this characterizes the existing guard.

## Scope

- Affects View lifecycle tests and documentation only.
- Does not change runtime code, Application lifecycle, State, Region lookup behavior, or post-destroy operations other than repeated destroy.

## Runtime cost

- Static or documentation only.
- Bundle: zero runtime artifact delta.
- Allocations/retention: unchanged.

## Deployment Impact

No application redeploy or package publication is needed. This PR changes repository tests and documentation only.

## Testing

- Narrow View lifecycle suite: 9 tests passed.
- npm run coverage: 70 files, 1,222 tests, 100% statements/branches/functions/lines; agent benchmark contract 19/19.
- npm run lint:ci.
- npm run check:dist.
- git diff --check.
- Confirmed no changes under runtime, build, dist, entrypoint, or manifest paths.

## Package and browser impact

- Browser contracts affected: none; the unit test uses the existing DOM harness.

## Rollback or deprecation

Revert the test/documentation commit if the broader #131 lifecycle design intentionally changes reentrant destruction before stable v5; no runtime rollback is required.

## Review notes

This is a partial #131 slice. Application invalidation and other unsupported post-destroy operations remain out of scope.

---
## Summary by cubic
Adds a lifecycle contract test and documentation for reentrant View destruction, covering part of #131 by making the existing guard against repeated teardown explicit.

- Calling `destroy()` from `before:destroy`, `destroy`, or after completion returns the same View without restarting teardown.
- Parent detach, managed child destruction, and owning Region empty each fire exactly once.
- No runtime code changes; the PR only updates `test/unit/view-lifecycle.spec.js` and `docs/view.lifecycle.md`.

<sup>Written for commit 3fd05165d919fa1db281238c3af0df45167aef03. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/222?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/view.lifecycle.md
+++ b/docs/view.lifecycle.md
@@ -249,6 +249,11 @@ a view's instance is no longer referenced the view can be cleaned up by the brow
 The [`before:destroy` event](./events.class.md#destroy-and-beforedestroy-events) is the best place to clean
 up any added listeners not related to the view's DOM.
 
+Once destruction begins, reentrant `destroy()` calls from `before:destroy` or
+`destroy`, and later repeated calls, return the same View without restarting
+teardown. An attached parent, its managed children, and its owning Region therefore
+complete their detach, destroy, and empty lifecycles once.
+
 The state of the view after the destroy is not attached and not rendered although the `el` is not emptied.
 
 ## Destroying Children
--- a/test/unit/view-lifecycle.spec.js
+++ b/test/unit/view-lifecycle.spec.js
@@ -226,6 +226,70 @@ describe('View lifecycle contract', function() {
     region.destroy();
   });
 
+  it('ignores reentrant and repeated destroy calls while tearing down once', function() {
+    this.setFixtures('<div id="reentrant-destroy-region"></div>');
+    const parent = new View({
+      regions: { child: '.child-region' },
+      template: () => '<div class="child-region"></div>',
+    });
+    const child = new View({ template: () => '<span>Child</span>' });
+    const region = new Region({ el: '#reentrant-destroy-region' });
+    let beforeDestroyReturn;
+    let destroyReturn;
+    const lifecycle = {
+      parentBeforeDestroy: this.sinon.spy(currentView => {
+        beforeDestroyReturn = currentView.destroy();
+      }),
+      parentBeforeDetach: this.sinon.spy(),
+      parentDetach: this.sinon.spy(),
+      parentDestroy: this.sinon.spy(currentView => {
+        destroyReturn = currentView.destroy();
+      }),
+      childBeforeDetach: this.sinon.spy(),
+      childDetach: this.sinon.spy(),
+      childBeforeDestroy: this.sinon.spy(),
+      childDestroy: this.sinon.spy(),
+      regionBeforeEmpty: this.sinon.spy(),
+      regionEmpty: this.sinon.spy(),
+    };
+
+    parent.on({
+      'before:destroy': lifecycle.parentBeforeDestroy,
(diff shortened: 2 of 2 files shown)


paulfalgout:
Reviewed at `3fd05165` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878, margin still 97 bytes.

**Putting the reentrancy inside the spies is what makes this test hard to fake.** `destroy()` is invoked from within both `before:destroy` and `destroy`, plus twice externally — four calls from three points in the lifecycle — and then all ten spies across parent, managed child, and owning Region are asserted `calledOnce`. Asserting the returns as well (`beforeDestroyReturn === parent`) pins the documented "return the same View" rather than just "doesn't blow up".

**I traced why the `before:destroy` case is guarded at all, because it isn't obvious.** The base mixin sets the flag *after* the event:

```js
// mixins/destroy.js
if (this._isDestroyed) { return this; }
this.triggerMethod('before:destroy', this, options);
this._isDestroyed = true;
```

so `_isDestroyed` alone cannot catch a reentrant call from `before:destroy`. What saves View is a second, separate flag:

```js
// mixins/view.js:122
if (this._isDestroyed || this._isDestroying) { return this; }
this._isDestroying = true;
```

Two flags doing genuinely different jobs — `_isDestroying` as the reentrancy guard set immediately, `_isDestroyed` as observable state set at the point that keeps `isDestroyed()` reading `false` during `before:destroy`, which is what #198's Behavior test depends on. Worth appreciating rather than collapsing into one flag later.

**That asymmetry is worth knowing about, and this PR's wording is what surfaces it.** `MnObject` and `Application` compose the base `DestroyMixin` (`modules/object.js:7`, `modules/application.js:8`) and have no `_isDestroying`. Reproducing those exact semantics:

```
before:destroy handler calls destroy()
→ RECURSION: before:destroy fired 21 times   (stopped by my own counter)
```

So the same pattern this PR documents as safe recurses unbounded on an `MnObject` or `Application`. The new paragraph lives in `view.lifecycle.md` and says "View" throughout, so it's accurate as written — but it's now the only place in the docs describing reentrant destroy, and a reader is likely to generalise it. Either extending `_isDestroying` to the base mixin or noting the limitation where the object lifecycle is documented would close that gap; a follow-up rather than something for this PR.

PR #225, opened 2026-08-29
test(docs): execute CollectionView child ownership

paulfalgout:
Related #139

## What

- Replace the broken `addChildView` example with a self-contained shipped-package example.
- Add an executable packed-package fixture for add, detach, re-add, remove, and owner teardown.
- Link the documented already-owned failure to `MN0003`.

## Why

The previous example contained invalid variable usage and did not prove the ownership or lifecycle contract it described. Agents need a copyable example whose ownership transfer, DOM containment, render reuse, and exactly-once cleanup behavior execute in CI.

## Scope

- Documentation for `CollectionView` child management.
- Packed-package fixture coverage and fixture registration.
- No runtime, distribution artifact, configuration, package contract, or website publication changes.

## Deployment Impact

No deployment or redeploy is required. This PR does not publish the website, package, tag, or release.

## Testing

- `npm run test:fixtures`
- `npm run docs:check`
- `npm run lint:ci`
- `git diff --check`
- Verified no runtime, config, dist, or package diff against `origin/master`.

---
## Summary by cubic
Replaces the broken `addChildView` documentation example with an executable one that verifies the full child ownership lifecycle (add, detach, re-add, remove, and owner teardown) in CI.

- The old example referenced undefined variables and never ran; the new one is a self-contained packed-package fixture executed in CI.
- Documents that `detachChildView()` returns the same live view for re-adding without re-rendering, and `removeChildView()` destroys the removed view.
- The already-owned error now links to the `MN0003` error page.

No runtime, configuration, distribution, or package changes; no deployment needed.

<sup>Written for commit 92254d37262b6f17ec282bc8e4ac225e27252d50. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/225?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added an executable example explaining `CollectionView` child view ownership and lifecycle.
  * Clarified that detaching preserves a live view, removing destroys it, and destroying the collection view cleans up managed children.
  * Updated the error guidance to reference the `MN0003` error code.

* **Tests**
  * Added automated validation to ensure the documented child view lifecycle behaves as described.


paulfalgout:
Reviewed at `92254d37` (the SHA I fetched). **Approve.** No bundle impact — aggregate unchanged at 51878.

**The example this replaces could never have run**, and that's the strongest case yet for the executable-example programme:

```js
import ButtonView from './button-view';   // module that does not exist
...
View = new ButtonView();                  // assigns to `View`
this.addChildView(buttonView, …);         // reads `buttonView` — never defined
```

A name mismatch guaranteeing a `ReferenceError`, a relative import to nothing, and `from 'backbone.marionette'` — the v4 package name — all in a snippet a reader was invited to copy. Three defects that survived indefinitely because nothing executed it. #223 turned up the same stale package name; this one adds code that was simply wrong. That reframes the pattern from documentation hygiene to defect detection.

**The fixture applies the conventions this series has converged on**, and adds one worth noting. The `this === collectionView` pins on every wrapped method matter more here than on #221, because `addChildView` is called three times with two different Views — asserting identity per call is what keeps a reordered or duplicated example from satisfying the counters. And:

```js
assert.equal(renderCounts.get(view), 1, 'the detached child must already be rendered before re-add');
…
assert.equal(renderCounts.get(view), 1, 're-adding a rendered child must not render it again');
```

Checked before *and* after the re-add, which is what proves the documented "may be added again without rendering it a second time" rather than just that the count ended at one.

Checking `children.hasView(view)` on both sides of each operation is the right ownership assertion too — it distinguishes "released ownership" from "removed from the DOM", which `el.contains` alone would conflate.

**One light note on the sharpened error text.** Naming [`MN0003`](/errors/MN0003/) instead of "an error will be thrown" is a real improvement, and I verified the code is right — catalog slug `view-already-owned`, thrown at `collection-view.js:684` and `view-region.js:116`, with remediation matching the doc's. It's backed by unit coverage at `collection-view-children.spec.js:437` and `region.spec.js:626`, so this isn't an unbacked claim by the standard I've applied elsewhere.

Worth considering anyway: unlike the MN0018/MN0019 mentions on #218, this error is the direct counterpart of what this example teaches. The example demonstrates the *legal* transfer — detach, then re-add — and the note describes the *illegal* one. Those are two halves of one rule, and the fixture already has a `collectionView` and live Views in hand, so asserting the throw is a couple of lines. Your call; the claim is covered either way.

paulfalgout:
Keeping the `MN0003` throw out of this packed fixture intentionally. The diagnostic and both runtime throw sites already have focused unit coverage; this fixture stays bounded to executing the documented legal transfer lifecycle through the shipped package. The note links that separate failure contract without duplicating it here.

PR #228, opened 2026-08-29
test(docs): execute View-owned Region lifecycle

paulfalgout:
Related #139

## What

- Replace stale View child-Region snippets with one self-contained public-API lifecycle example.
- Demonstrate show-before-parent-render, exact child lookup, detach, re-show without rerender, and owner-directed Region emptying.
- Execute the published example against the packed package, including lifecycle counts, DOM containment, ownership, and repeated-run cleanup.

## Why

The previous reference examples imported the v4 package name, depended on a nonexistent `./subview`, duplicated setup across sections, and showed child cleanup calls on the wrong owner. Agents need one copyable example whose documented identities and lifecycle behavior are proven by the released package shape.

## Scope

- Affects `docs/marionette.view.md` and one packed-package documentation fixture.
- Removes obsolete external fiddle links for the replaced snippets.
- Does not change runtime source, configuration, exports, dependencies, distribution artifacts, or package metadata.

## Deployment Impact

No website, npm package, tag, or release is published by this PR. No consumer redeploy is required.

## Testing

- Focused `docs-view-child-region` packed-package fixture.
- `npm run test:fixtures`
- `npm run docs:check`: 33 pages, 34 HTML files, 320 links.
- `npm run lint:ci`
- `npm run check:dist`
- `npm run size`: runtime remains 51,878/51,975 bytes on this branch base.
- `git diff --check`

The fixture proves the parent auto-renders, `getChildView` returns the exact shown instance, detach preserves a live/rendered child, re-show does not rerender, Region empty destroys and releases the child, DOM containment follows ownership, and a second run starts from fresh instances.

---
## Summary by cubic
Replaces the stale View child-Region snippets in `docs/marionette.view.md` with one self-contained executable lifecycle example, closing the documentation gaps from #139.

- Old examples imported the v4 package name, depended on a nonexistent `./subview`, and showed cleanup on the wrong owner.
- The new example covers show-before-parent-render, exact child lookup, detach without rerender, re-show, and owner-directed Region emptying.
- Adds a `docs-view-child-region` packed-package fixture that runs the published example and asserts lifecycle counts, DOM containment, ownership, and fresh instances on repeated runs.
- Removes the obsolete external fiddle links.
- No runtime source, configuration, exports, or package metadata changes.

<sup>Written for commit ccb467c0c83717d726ec54e8dae9822747d47d11. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approve** at `ccb467c0` (the SHA I fetched). This is the strongest of the executable-docs fixtures so far, and the removals are all corrections rather than losses.

The mechanism is right where it counts: `validate.mjs` reads `docs/marionette.view.md`, asserts **exactly one** occurrence of the marker, and extracts the fence immediately following it. The published example is the thing under test, so the doc cannot drift from the behavior — and the single-marker assertion means a second copy of the marker fails loudly instead of silently testing the wrong block. `run.mjs` installs the packed tarball with `npm install --no-save`, so this exercises the released package shape rather than the source tree.

I checked the removals individually, since a net −77 lines is where content usually goes missing. Every deleted snippet was actually broken: `backbone.marionette` is the v4 package name, `./subview` never existed, and `myChildView.getChildView('regionName').destroy()` called the accessor on the child rather than the owner. The `MN0020` sentence survives and is upgraded from a bare code to `[MN0020](/errors/MN0020/)`, matching the convention already used across four docs files. The only real subtraction is the "directly destroy the child" alternative; steering to owner-directed `region.empty()` is the safer idiom to hand an agent, so I read that as deliberate.

The assertions earn their keep rather than restating the example. `assert.deepEqual(run.childEvents, {beforeRender: 1, render: 1, beforeDestroy: 1, destroy: 1})` is what actually proves the interesting claim — re-showing a detached child does **not** rerender it — and it's checked against the same counter that already read 1 after the first show. Ownership is verified on both sides (`currentView` identity *and* `el.contains`), the second run asserts fresh instances rather than reusing state, and teardown lives in a `finally` so a mid-run failure can't leak a JSDOM window into the next fixture.

CI is green at this SHA, and the fixture genuinely executes — `test:fixtures` runs in `ci.yml` (both jobs) and `node-current.yml`, and the new directory is registered in the runner's list.

One nit, preventive only. `run.mjs` carries a hardcoded `fixtures` array with no cross-check against what's on disk. A listed-but-missing fixture throws, which is good, but a fixture directory that exists and is *not* listed is silently never run — and since `run.mjs` prints nothing per fixture, the CI log can't distinguish "ran and passed" from "never ran." I diffed the two and they're currently in exact agreement (15/15, including this PR's addition), so nothing is inert today. Given that the whole point of this program is that documented examples stay verified, deriving the list from the directory — or asserting the two match — would close the gap permanently for about three lines.

PR #229, opened 2026-08-29
test(fixtures): discover package fixtures

paulfalgout:
Related #139

## What

- Discover immediate packed-package fixture directories from their `package.json` files.
- Sort discovered fixture names before running them.
- Remove the obsolete manually maintained fixture registration list.

## Why

The executable-documentation program can currently add a fixture directory without registering it in `test/fixtures/run.mjs`, causing CI to silently skip it. Discovery makes every package fixture part of the validation run by construction.

## Scope

- Affects only `test/fixtures/run.mjs`.
- Ignores non-directory entries and directories without a `package.json`.
- Does not change runtime source, distribution artifacts, package metadata, exports, dependencies, workflows, performance contracts, or repository rules.

## Deployment Impact

No website, npm package, tag, or release is published by this PR. No consumer redeploy is required.

## Testing

- `npm run test:fixtures`: all 15 packed-package fixtures passed.
- `npm run lint:ci`
- `npm run check:dist`
- `npm run size`: 51.88 kB / 51.98 kB.
- `git diff --check`
- Verified discovery returns the same 15 current fixtures in deterministic order.

---
## Summary by cubic
Discovers packed-package test fixtures automatically from their `package.json` files instead of relying on a manually maintained list, so new fixture directories can no longer be silently skipped by CI.

- Reads `test/fixtures/` for directories containing a `package.json` and runs them in sorted order.
- Removes the hardcoded fixture registration list from `test/fixtures/run.mjs`.
- Excludes non-directory entries and directories without a `package.json`; fails the run if no fixtures are discovered.
- Affects only the fixture runner; no runtime, package, or consumer impact.
- All 15 existing fixtures pass with the same deterministic order.

<sup>Written for commit 7ce815a6fbcca1d7801bbcca88961edb827f979b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/229?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/fixtures/run.mjs
+++ b/test/fixtures/run.mjs
@@ -8,23 +8,15 @@ const rootDir = resolve(__dirname, '../..');
 const packDir = resolve(rootDir, 'test/tmp/pack-fixtures');
 const npmCli = process.env.npm_execpath;
 const cliArgs = process.argv.slice(2);
-const fixtures = [
-  'cjs-node',
-  'cjs-adapters',
-  'esm-node',
-  'no-default-export',
-  'shim',
-  'shim-types',
-  'jquery-dom-api',
-  'jquery-dom-api-types',
-  'docs-behavior-host',
-  'docs-collectionview-child-ownership',
-  'docs-radio-owner',
-  'docs-region-lifecycle',
-  'docs-view-child-region',
-  'vite',
-  'peer-underscore-min',
-];
+const fixtures = readdirSync(__dirname, { withFileTypes: true })
+  .filter(entry => entry.isDirectory() &&
+    existsSync(resolve(__dirname, entry.name, 'package.json')))
+  .map(entry => entry.name)
+  .sort();
+
+if (fixtures.length === 0) {
+  throw new Error('No packed-package fixtures discovered under test/fixtures');
+}
 
 function run(command, args, options = {}) {
   execFileSync(command, args, {


paulfalgout:
**Approve** at `b72ecc8a` (the SHA I fetched). This closes the gap I raised on #228, and the implementation is the right shape — the predicate for "is a fixture" is now the same fact the runner already depended on, so the two can't disagree.

I verified the equivalence claim rather than taking it: enumerating `test/fixtures/*/package.json` at this head yields exactly the 15 names the deleted array listed, no additions and no drops. `diff` of the two sets is empty.

The alphabetical reordering is safe. Each fixture is `cleanFixture`d before and after its run — `dist`, `node_modules`, and `package-lock.json` all removed — with its own `npm install` and tarball install inside a `try`/`finally`, so no fixture can observe another's state and order carries no meaning.

**One thing worth fixing before merge, because it re-creates the same failure mode one level up.** Discovery has no lower bound:

```js
const fixtures = readdirSync(__dirname, { withFileTypes: true })
  .filter(entry => entry.isDirectory() && existsSync(…'package.json'))
  .map(entry => entry.name)
  .sort();
```

If that expression ever yields `[]`, the loop body never runs and the script exits 0. CI reports `test:fixtures` green having validated nothing. The hardcoded list had exactly one virtue this loses: it asserted that 15 specific fixtures *must* exist, and the `Fixture is missing package.json` throw fired if any went absent.

This isn't hypothetical. The plausible trigger is a directory restructure — group the fixtures into `test/fixtures/docs/*` and `test/fixtures/pkg/*` and the immediate children stop carrying `package.json`. Discovery quietly returns zero, the whole packed-package suite stops running, and every check stays green. Under the old code that same refactor failed loudly fifteen times. Given that this PR exists precisely because silent skips are the enemy, a floor seems worth three lines:

```js
if (fixtures.length === 0) {
  throw new Error('No packed-package fixtures discovered under test/fixtures');
}
```

Separately, the guard at `run.mjs:85-87` is now unreachable — discovery filters on `existsSync(…'package.json')`, so a discovered fixture always has one. Harmless, but it now reads as a live invariant when it can't fire. Since `run.mjs` prints nothing per fixture, an emitted count would also make the log self-evidencing, which is what let the #228 gap hide in the first place.

None of this blocks: the set is correct today and CI is green at this SHA.

paulfalgout:
Addressed the valid zero-discovery gap in `7ce815a`: the runner now fails explicitly when no packed-package fixtures are found.

I kept the later per-fixture `package.json` guard because it still protects against the file disappearing between discovery and execution. I did not add log output because the failure floor closes the silent-zero-run path without broadening this harness-only slice.

Exact updated-head validation passed: `npm run lint:ci`, `npm run test:fixtures` (all 15 fixtures), and `git diff --check`.

paulfalgout:
**Approve** at `7ce815a6` (the SHA I fetched). The floor is in and my finding is resolved.

The delta from `b72ecc8a` is exactly the guard, and its placement is right — top-level, immediately after discovery, so it fails before `npm pack` and before any fixture directory is touched. A restructure that empties discovery now aborts the run instead of reporting a green suite that validated nothing. Discovery still resolves to 15 at this head, and CI is green.

Not addressed, both explicitly optional and neither worth another round: the now-unreachable `Fixture is missing package.json` guard at `run.mjs:85-87`, and emitting the discovered count so the log is self-evidencing rather than silent.

Issue #147, opened 2026-08-26
v5: complete stable release evidence and reference documentation


paulfalgout:
## Current status — September 9, 2026

Repository readiness work is merged in [#500](https://github.com/marionettejs/marionette/pull/500) at [master `70eaeaf5`](https://github.com/marionettejs/marionette/commit/70eaeaf582a3ae062b5ff48650cfda12cfa3bf9f). It is available in source and verified tarballs; no new npm version or tag was published. The registry remains beta.1 on `latest`.

The finite repository closeout is integrated: authorship/browser inventories, compact reference, migration procedure, static tools, CI certification and [current browser evidence](https://github.com/marionettejs/marionette/actions/runs/34333478421). Stable publication, real application/corpus experience and all-package live provenance on the next authorized version remain future gates. The channel policy is deliberately `latest` before first stable v5, then `next` for later prereleases.

## Goal
Collect final package, performance, declaration, browser, documentation, migration and warning-free evidence for an exact stable release commit. Reopened because the former completed status did not satisfy these broad criteria; earlier documentation work remains delivered.

## Delivered evidence to reuse
- [x] Current core/optional public typing and installed-package fixtures (#429).
- [x] Source cleanup and separately reviewed Behavior initialization correction (#430).
- [x] README/guide voice refresh, executable examples, generated documentation and link validation (#431).
- [x] Existing distribution, package, browser, diagnostic and byte/graph CI infrastructure.

These PRs document their actual validation. This issue reconciliation does not rerun tests or claim that a future release commit passed them. Public agent results and a dedicated stable-release timing baseline are still absent. The new portable workload scripts and desktop baseline exist locally under #127 and await integration/public availability. #376 records the completed CollectionView candidate audit; #451/#453 record regressions and matched local timing with explicit accessibility/lifecycle settings.

## Remaining stable-release acceptance criteria
- [ ] All applicable Phase 0, 1, 2 and 4 gates and evidence-dependent decisions pass on the exact release commit; #41 reflects their actual state. The maintainer explicitly deferred Phase 3 (#141–#143); those tools are not stable blockers. #144 retains only accepted static-tooling package verification.
- [ ] Final packed artifacts cover every supported export, module format and declaration mode and match the tested build.
- [ ] Pinned browser evidence (#145), controlled-runner timing and performance reporting (#127) are recorded for that candidate. Enforceable budgets require a separate evidence-backed decision after stabilization; no retired alpha budget is implicitly restored.
[...]

PR #234, opened 2026-08-29
docs(reference): repair package and source links

paulfalgout:
Related #147

## What

- Repair the malformed relative link from the common Events API note to the current DOM-interactions reference.
- Update both `docs/utils.md` examples to import the documented utilities from the published `marionette` root package.

## Why

The current-reference documentation still contains one path that source-level validation cannot resolve and two examples that use the obsolete v4 package name. This bounded slice keeps the examples aligned with the v5 package contract while the remaining source-link repairs wait for the lifecycle PR files they overlap.

## Scope

- Documentation only: `docs/common.md` and `docs/utils.md`.
- No runtime, distribution, package map, lockfile, workflow, migration, or website-publication behavior changes.
- This completes the `docs/utils.md` package-name correction, not the repository-wide obsolete-language sweep.

## Deployment Impact

No application redeploy or package publication is required. The public documentation website remains unpublished until the post-v5 publication phase.

## Testing

- `npm run docs:check`
- `npm run lint:ci`
- `git diff --check`
- Verified `docs/utils.md` contains no remaining `backbone.marionette` references.

---
## Summary by cubic
Fixes the malformed relative link in the Events API note so the view events reference resolves correctly, and updates `docs/utils.md` examples to import utilities from the published `marionette` root package instead of the obsolete `backbone.marionette`.

This is documentation-only; no runtime, package, or publication behavior changes.

<sup>Written for commit 9974635c55d7af3d73765908f9e34595e2b57329. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #235, opened 2026-08-29
docs(reference): update entity event package imports

paulfalgout:
Related #147

## What

- Update all five `docs/events.entity.md` examples to import `View` from the published `marionette` root package.

## Why

The entity-event examples still use the obsolete v4 package name even though v5 publishes `View` from `marionette`. Keeping canonical examples importable is part of the final reference-documentation evidence.

## Scope

- Documentation only: `docs/events.entity.md`.
- No entity-event behavior, runtime, distribution, package map, lockfile, workflow, migration, or website-publication changes.
- This completes the package-name correction for this reference page, not the repository-wide obsolete-language sweep.

## Deployment Impact

No application redeploy or package publication is required. The public documentation website remains unpublished until the post-v5 publication phase.

## Testing

- `npm run docs:check`
- `npm run lint:ci`
- Verified `docs/events.entity.md` contains no remaining `backbone.marionette` references.

---
## Summary by cubic
Updates `docs/events.entity.md` to import `View` from the published `marionette` package instead of the obsolete `backbone.marionette` in all five examples. This makes the reference examples importable under v5 and completes the package-name correction for this page. No runtime or build changes are included.

<sup>Written for commit b8f479bbc06bf495d00be7fe2fcc206ef3ec1add. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/events.entity.md
+++ b/docs/events.entity.md
@@ -11,7 +11,7 @@ and called with the context of the view instance.
 For example, to listen to a model's events:
 
 ```javascript
-import { View } from 'backbone.marionette';
+import { View } from 'marionette';
 
 const MyView = View.extend({
   modelEvents: {
@@ -37,7 +37,7 @@ The `modelEvents` attribute can also take a
 You can also bind a function callback directly in the `modelEvents` attribute:
 
 ```javascript
-import { View } from 'backbone.marionette';
+import { View } from 'marionette';
 
 const MyView = View.extend({
   modelEvents: {
@@ -56,7 +56,7 @@ Collection events work exactly the same way as [`modelEvents`](#model-events)
 with their own `collectionEvents` key:
 
 ```javascript
-import { View } from 'backbone.marionette';
+import { View } from 'marionette';
 
 const MyView = View.extend({
   collectionEvents: {
@@ -78,7 +78,7 @@ Just as in `modelEvents`, you can bind function callbacks directly inside the
 `collectionEvents` object:
 
 ```javascript
-import { View } from 'backbone.marionette';
+import { View } from 'marionette';
 
 const MyView = View.extend({
   collectionEvents: {
@@ -97,7 +97,7 @@ If your view has a `model` and `collection` attached, it will listen for events
 on both:
 
 ```javascript
-import { View } from 'backbone.marionette';
+import { View } from 'marionette';
 
 const MyView = View.extend({
 


paulfalgout:
**Approve** at `b8f479bb` (the SHA I fetched). Clean and complete for this page: all five imports converted, nothing but the specifier changed, and no `backbone.marionette` remains in the file. `View` comes off the root package, which I confirmed on #234. No CI failures.

Repo-wide the count moves 130 → 125.

One process question, not an objection. At roughly a page per PR, the remaining 125 is on the order of a dozen more of these, each needing a review round for a change that is mechanically identical and verifiable by grep. If the page-at-a-time cadence is there to stay clear of in-flight lifecycle work, that makes sense and it's the right call — but the docs-heavy PRs have largely landed now, and the files still holding stale imports don't obviously collide with #224:

```
marionette.collectionview.md  30    events.class.md   11
view.rendering.md             17    basics.md          9
marionette.region.md          16    upgrade-v3-v4.md   7  (historical — likely leave as-is)
```

A single sweep with `upgrade-v3-v4.md` deliberately excluded would be one review instead of a dozen, and the verification is the same grep either way. Worth considering if nothing structural is forcing the split.

Separately, and more durable than the rename: these five examples still aren't executable-verified, so nothing stops the *next* drift. `events.entity.md` has no `executable-example` marker, which means the imports are now correct but unproven — the same gap that let the v4 name persist here in the first place. That's the argument for the fixture program eventually reaching these pages rather than the sweep being the end state.

paulfalgout:
Agreed. After this page-level slice, I’ll batch the remaining mechanically valid, non-overlapping current-reference imports into one audited sweep, while keeping pages with broader invalid v5 examples or #224/#190 overlap in their own contract-aware work.

paulfalgout:
Follow-up audit changed that plan: the large non-overlapping pages are not mechanically equivalent after all. `common.md`, `marionette.collectionview.md`, and `view.rendering.md` each contain additional v5-invalid examples, while `features.md` is coupled to the pending unknown-flag diagnostic. I’m batching only where the whole example remains valid; the rest will be repaired page-by-page as complete contracts rather than renamed into still-broken snippets.

PR #236, opened 2026-08-29
docs(view): align reference with v5 contracts

paulfalgout:
Related #147

## What

- Update the View reference examples to import from the published `marionette` root package.
- Describe View construction, rendering, DOM interaction, and Region selectors using the v5 standalone/native contracts.
- Remove obsolete claims that the core View inherits Backbone.View behavior or requires jQuery selector syntax.

## Why

The canonical View page still mixed v4 package and dependency assumptions into otherwise current v5 guidance. Agent-authored code needs the reference to state the framework-owned contract directly so examples do not reintroduce Backbone or jQuery coupling accidentally.

## Scope

- Documentation only: `docs/marionette.view.md`.
- No View runtime, serialization, DOM API, package, lockfile, workflow, migration, or website-publication changes.
- The retained backbone-events reference describes the supported event protocol, not a core dependency.

## Deployment Impact

No application redeploy or package publication is required. The public documentation website remains unpublished until the post-v5 publication phase.

## Testing

- `npm run docs:check`
- `npm run lint:ci`
- Verified all 40 unique links in `docs/marionette.view.md` resolve against current headings and repository targets.
- Verified the removed Backbone/jQuery/package claims against current View source, package exports, optional-peer contracts, and native DOM behavior.

---
## Summary by cubic
Aligns the View reference docs with v5 contracts by updating imports and removing Backbone and jQuery references.

- Replace `backbone.marionette` imports with `marionette` in examples.
- Remove claims that View inherits from `Backbone.View` and that region selectors use jQuery syntax.
- Clarify that View properties and `render` are defined by Marionette's standalone `View`.
- Documentation-only changes; no runtime, package, or deployment impact.

<sup>Written for commit 0ef05076c4792cfc5c777945c61c2518aa0acc2c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.view.md
+++ b/docs/marionette.view.md
@@ -2,8 +2,8 @@
 
 A `View` is used for managing portions of the DOM via a single parent DOM element or `el`.
 It provides a consistent interface for managing the content of the `el` which is typically
-administered by serializing a `Backbone.Model` or `Backbone.Collection` and rendering
-a template with the serialized data into the `View`s `el`.
+administered by serializing attached model or collection data and rendering a template
+with that data into the `View`'s `el`.
 
 The `View` provides event delegation for capturing and handling DOM interactions as well as
 the ability to separate concerns into smaller, managed child views.
@@ -48,13 +48,12 @@ that will be attached directly to the instance:
 `tagName`, `template`, `templateContext`, `triggers`, `ui`
 
 ```javascript
-import { View } from 'backbone.marionette';
+import { View } from 'marionette';
 
 const myView = new View({ ... });
 ```
 
-Some of these properties come from Marionette, but many are inherited from
-[`Backbone.View`](http://backbonejs.org/#View-constructor).
+These properties are defined by Marionette's standalone `View` constructor.
 
 ## Rendering a View
 
@@ -63,8 +62,8 @@ The Marionette View implements a powerful render method which, given a
 HTML from that template, mixing in `model` or `collection` data and any
 extra [template context](./view.rendering.md#adding-context-data).
 
-Unlike `Backbone.View` Marionette defines `render` and this method should
-not be overridden. To add functionality to the render use the
+Marionette `View` defines `render`, and this method should not be overridden.
+To add functionality around rendering, use the
 [`render` and `before:render` events](./events.class.md#render-and-beforerender-events).
 
 [Live example](https://jsfiddle.net/marionettejs/dhsjcka4/)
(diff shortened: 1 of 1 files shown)


paulfalgout:
**Approve** at `0ef05076` (the SHA I fetched). Both substantive claims check out against the source, which matters more here than on the rename-only PRs since this one asserts contracts rather than just changing a specifier.

"These properties are defined by Marionette's standalone `View` constructor" is accurate — `modules/view-region.js` contains no reference to Backbone at all, and `const View = function(options)` at line 735 does its own setup (`_setOptions`, `_getEl`/`setElement`, `_initViewEvents`) with no `Backbone.View.apply`. The old "many are inherited from `Backbone.View`" was simply false in v5.

The Region selector claim holds too. `getEl` resolves through `this.Dom.findEl(context || document, el)[0]` with `context = result(this, 'parentEl')`, so for the `regions` hash those strings really are CSS selectors scoped to the View's `el`. Dropping "any valid jQuery syntax will suffice" is the right call — under the default native DomApi, jQuery-only extensions like `:first` or `:contains()` are not valid CSS and would throw at `querySelector`, so the old sentence actively invited broken code.

Keeping the backbone-events catalog link at line 94 is justified as you describe: `docs/events.md` already states the Events primitive implements that API without requiring Backbone, so it reads as a protocol reference rather than a dependency claim.

**One thing this PR surfaces but doesn't cover — the same obsolete claims still live in source comments.** Three remain:

```
modules/view-region.js:306   "a jQuery selector object scoped to a provided parent el or the document"
mixins/ui.js:95              "the associated jQuery selectors."
modules/behavior.js:64       "this is useful for doing jquery DOM lookups"
```

The first is the one I'd fix. It sits directly above `getEl`, whose preceding line is *"Override this method to change how the region finds the DOM element that it manages"* — so it's instructions for an extension point, and it now describes the wrong return type. `findEl(...)[0]` yields a DOM element; anyone following that comment and returning a jQuery object from a custom `getEl` would set `region.el` to a jQuery wrapper and break every native operation downstream. That's a comment that leads to broken code, not merely a stale word.

Worth folding into whichever PR does the source-comment pass, rather than expanding this one. All three imports in the page are converted, none remain, and CI is green at this SHA.

Issue #241, opened 2026-08-29
v5: remove the required Underscore runtime dependency


paulfalgout:
## Goal

Remove Underscore as a required Marionette v5 runtime peer before stable while preserving the documented, useful Backbone/Marionette collection ergonomics developers rely on.

This explicitly supersedes #48's 2020 instruction not to remove Underscore in v5. Backbone applications may continue using Underscore directly, and Marionette may retain it as a development dependency for Backbone/parity tests; the root Marionette runtime and its shipped subpaths must not import or require it.

Related to #41 and the runtime-cost contract in `ROADMAP.md`.

## Agent-development failure addressed

Generic Underscore iteratee shorthand makes ordinary method names accept multiple unrelated input shapes and hides whether a predicate reads a child View, its model, or model attributes. Agents are more reliable when native-named methods use ordinary JavaScript callback/value semantics and convenience methods have one explicit interpretation.

Removing the dependency must not erase Marionette's useful ergonomics or make Backbone-oriented development needlessly verbose.

### Preserve the documented ChildViewContainer vocabulary

Retain and own the 19 currently documented methods:

- `each`
- `map`
- `reduce`
- `find`
- `filter`
- `reject`
- `every`
- `some`
- `contains`
- `invoke`
- `toArray`
- `first`
- `initial`
- `rest`
- `last`
- `without`
- `isEmpty`
- `pluck`
- `partition`

Add `Symbol.iterator` on the prototype so children work with `for...of`, spread, destructuring, and `Array.from` without per-instance allocation.

Remove only the six undocumented aliases already proxied from Underscore:

- `forEach` (`each` is canonical)
- `detect` (`find` is canonical)
- `select` (`filter` is canonical)
- `all` (`every` is canonical)
- `any` (`some` is canonical)
- `include` (`contains` is canonical)

Do not add `where` or `findWhere` in this work.

### Make semantics explicit

- `each`, `map`, `find`, `filter`, `reject`, `every`, `some`, `reduce`, and `partition` accept functions, with individually documented callback arguments, context behavior, return values, and empty-container behavior.
- `contains` accepts a child View value.
- `pluck(key)` reads a property from each child View. It does not read model attributes. Existing `children.pluck('model')` remains valid.
- `invoke`, `first`, `initial`, `rest`, `last`, `without`, `isEmpty`, and `toArray` retain their documented useful semantics unless characterization proves a contract is unsafe or incoherent; any intentional change requires explicit migration guidance.
[...]

PR #242, opened 2026-08-29
docs(roadmap): require Underscore-free v5 runtime

paulfalgout:
Related #241

## What
- Establish that stable v5 has no required Underscore runtime peer or production import.
- Preserve the documented ChildViewContainer vocabulary and Backbone integration contract while allowing undocumented pure aliases to be removed.
- Require exact-base, dependency-inclusive evidence and the existing budget-amendment process for dependency-removal work.

## Why
The original Underscore decision predates the agent-first v5 strategy. Removing the dependency after stable would create a new breaking migration, while removing documented ergonomic APIs would make Marionette and Backbone development harder. This records the intended boundary before implementation begins.

## Scope
- ROADMAP only.
- Runtime implementation is tracked in #241.
- No source, package, dependency, workflow, GitHub setting, website, or publication behavior changes.

## Deployment Impact
No deployment or redeploy is required.

## Testing
- Ran `git diff --check`.
- Reviewed the complete one-file diff against current `master`.
- Runtime tests are not applicable to this documentation-only change; pull-request CI is pending.

---
## Summary by cubic
Adds a v5 release requirement that removes Underscore from Marionette's runtime production path while keeping the documented `CollectionView.children` and Backbone integration semantics. This records the dependency-removal boundary before implementation, addressing the gap raised in #241.

- ROADMAP now states that production entrypoints and shipped subpaths must not import or require Underscore, and that undocumented pure aliases may be removed but documented methods are kept.
- Dependency-removal work must provide exact-base, dependency-inclusive bundle evidence and follow the existing budget-amendment process.
- This is a docs-only change; no source, package, or behavioral impact.

<sup>Written for commit 3d254d7391a81686884e01b42df69c83eecdef9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/242?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -256,6 +256,17 @@ Appropriate core additions include public read-only ownership accessors, pure Re
 lookup, shared diagnostics, extension hooks with no registered listeners by default,
 and narrowly designed resource ownership.
 
+Stable v5 removes Underscore as a required Marionette runtime peer without removing
+the documented, useful Backbone-era ergonomics of `CollectionView.children`. The
+container owns its documented method vocabulary and gains prototype-level iteration;
+callback methods use explicit function semantics, `pluck` reads child View properties,
+and model attributes are accessed explicitly through the View. Undocumented pure
+aliases may be removed, but a method is not removed merely because one private
+application does not use it. Private consumers may provide real-world migration
+evidence, but never define the public contract or a release gate. Marionette may use
+Underscore in development-only Backbone parity coverage, while production entrypoints
+and shipped subpaths do not import or require it.
+
 Core does not absorb a statechart runtime, signals runtime, virtual DOM, query layer,
 schema system, router, agent protocol, or inspector UI. Optional integrations with
 those systems may adapt to the State or data-source contracts without becoming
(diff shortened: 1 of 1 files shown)


paulfalgout:
**Approve** at `3d254d73` (the SHA I fetched). ROADMAP-only, no CI failures, and the text is careful in the places that matter — particularly "a method is not removed merely because one private application does not use it" and "private consumers may provide real-world migration evidence, but never define the public contract or a release gate," which line up with the existing "no release criterion depends on a private consumer" gate rather than cutting across it.

Two things worth knowing before #241 starts, neither a reason to hold this PR.

**The new release gate is not currently mechanically enforceable.** The criterion reads "Production source and shipped subpaths contain no Underscore import or required Underscore peer." I went looking for the knob that would enforce it and the existing machinery doesn't cover this case:

- `forbiddenProductionModules` / `forbiddenProductionModulePrefixes` are the only lists that raise a violation (`bundle-size.mjs:425`), and both hold **repo-internal paths** — `test/`, `docs/`, `config/release/`, `config/performance.json`. They don't describe npm packages.
- `underscore` lives in `baselineExternalImports` instead. Drift there **is measured** — `phase0AddedExternalImports` / `phase0RemovedExternalImports` at lines 325-326 — but nothing pushes a violation for it.

So if #241 removes `underscore` from `baselineExternalImports` while a production entrypoint still imports it, the report records `phase0AddedExternalImports: ['underscore']` and the check **passes green**. That's the same reporting-only shape as the new-subpath window on #188. Making this gate real needs either a violation on `phase0AddedExternalImports`, or extending the forbidden mechanism to external specifiers. Also worth noting `underscore` is presently a hard `peerDependencies` entry with no `peerDependenciesMeta.optional`, so "no required peer" is a package.json change too, not only an import change.

**The dependency-removal evidence clause leans on machinery that has not yet succeeded once.** The new bullet requires the two-stage ceiling amendment when shipped artifacts grow, and growth is close to certain here: Underscore is external today, so its bytes sit outside the 51878 aggregate, and owning replacements moves them inside. But #224 — the first-ever consumption of BA0001 — is currently red precisely on that path, with the base authority enforcing the pre-amendment 51975 against the candidate's 51995. Until that's resolved, this ROADMAP clause points at a process with a zero success rate, and #241 would be the second PR to hit it. Sequencing #224's fix ahead of #241 seems worth making explicit.

PR #245, opened 2026-08-29
test(collection): characterize boundary helpers

paulfalgout:
Related #241

## What

- Characterize ChildViewContainer.toArray, first, last, and isEmpty.
- Pin child View order and identity, fresh-array behavior, count forms, and empty-container results.
- Document these contracts separately from CollectionView.isEmpty.

## Why

Issue #241 removes the required Underscore runtime dependency while retaining useful collection ergonomics. These tests make the current non-callback boundary-helper behavior executable before the proxy is replaced.

## Scope

This changes only ChildViewContainer unit tests and CollectionView reference documentation. It does not change runtime source, aliases, package metadata, performance configuration, distribution artifacts, or the remaining Underscore-backed helpers.

## Deployment Impact

No redeploy or package publication is needed. This is test and documentation coverage only.

## Testing

- npm test -- test/unit/child-view-container.spec.js --reporter=dot — 49 passed
- npm run coverage — 70 files, 1,236 tests, 100% statements/branches/functions/lines; 19 agent benchmark contract tests passed
- npm run lint:ci
- npm run check:dist
- git diff --check

---
## Summary by cubic
Pins down the current `ChildViewContainer` boundary helper behavior with tests and docs before the Underscore-backed helpers are replaced in #241. This is test and documentation coverage only; runtime source, package metadata, and distribution artifacts are unchanged.

**Tests**
- Characterizes `toArray`, `first`, `last`, and `isEmpty` on `ChildViewContainer`.
- Verifies child view order, identity, fresh-array returns, count forms, and empty-container results.

**Docs**
- Documents the boundary helper contracts in `docs/marionette.collectionview.md`.
- Clarifies that `children.isEmpty()` is distinct from the overridable `CollectionView#isEmpty()`.

<sup>Written for commit 26675166721cb225e22eba0929968c07ca17f7e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/245?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -650,6 +650,20 @@ for those values. An empty container returns `[]`.
 or another object with the same properties is not considered contained. An empty
 container returns `false`.
 
+`toArray()` returns a new array containing the current child Views in container
+order. Changing the returned array's membership or order does not change the
+container. An empty container returns `[]`.
+
+Without a count, `first()` and `last()` return the first or last child View. With
+a nonnegative integer count, they return a new ordered array containing up to
+that many Views from the corresponding end of the container. A count of `0`
+returns `[]`. For an empty container, the no-count forms return `undefined` and
+the count forms return `[]`.
+
+`children.isEmpty()` reports whether the child container currently has zero
+Views. It is distinct from the overridable `CollectionView#isEmpty()` method,
+which controls whether a CollectionView renders its `emptyView`.
+
 ```javascript
 import Backbone from 'backbone';
 import { CollectionView } from 'backbone.marionette';
--- a/test/unit/child-view-container.spec.js
+++ b/test/unit/child-view-container.spec.js
@@ -66,6 +66,104 @@ describe('#ChildViewContainer', function() {
     });
   });
 
+  describe('ordered collection helpers', function() {
+    let container;
+    let views;
+
+    beforeEach(function() {
+      views = [
+        new Backbone.View(),
+        new Backbone.View(),
+        new Backbone.View()
+      ];
+
+      container = new ChildViewContainer();
+      container._set(views, true);
+    });
+
+    describe('#toArray', function() {
+      it('returns a new ordered array of the child views', function() {
+        const snapshot = container.toArray();
+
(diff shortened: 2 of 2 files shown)


paulfalgout:
**Approve** at `26675166` (the SHA I fetched). No CI failures. I exercised all thirteen documented claims against the shipped package and every one holds:

```
toArray: order correct, fresh array each call, mutating the result leaves the container intact
first()/last() -> View        first(2)/last(2) -> arrays        first(0)/last(0) -> []
empty: first()/last() -> undefined            first(2)/last(2) -> []
children.isEmpty() -> false / true
```

Two of these are worth more than the line count suggests.

`last(2)` returns the last two in **container order**, not reversed — my probe confirmed indexes `[1, 2]`. That's precisely the sort of detail a hand-written replacement gets wrong, because "last N" reads naturally as "walk backward from the end," and nothing else in the suite would catch the flip. Pinning it before the proxy is replaced is the whole point of this exercise.

The `CollectionView#isEmpty` separation I checked by actually overriding it rather than trusting the prose: with `isEmpty() { return true; }` on the CollectionView and one child present, `children.isEmpty()` still reports `false` while `CollectionView#isEmpty()` reports `true`. So the two really are independent, and documenting them separately is correct — conflating them would be an easy mistake for anyone reimplementing the container, since the names are identical.

The fresh-array assertion is the right shape too: `toArray() !== toArray()` plus mutating the result and re-reading proves non-aliasing from both directions, so an implementation that returned the internal array directly fails rather than passes.

One bookkeeping note on the scoping split I gave you on #244: `toArray` was in my undocumented column and this PR documents it. That column will keep shrinking as these slices land, so it's worth recomputing at implementation time rather than treating my numbers as fixed — which is also why I flagged the classification as approximate.

Issue #238, opened 2026-08-29
v5: adopt Baseline browser target and remove data overrides


paulfalgout:
## Goal

Express the v5 browser compilation contract semantically and remove unnecessary hard-pinning of Browserslist data packages.

Related to #41 and corrects the browser-target policy established by #153.

## Decision

Use the undated Browserslist query:

```json
"browserslist": [
  "baseline widely available"
]
```

Do not add a Baseline date and do not publish an explicit list of resolved browser versions as the support contract. The query defines the capability floor; newer browser releases remain supported.

## Scope

- Replace the current last-two browser-family query with `baseline widely available`.
- Remove the `overrides` for:
  - `caniuse-lite`
  - `electron-to-chromium`
  - `node-releases`
  - `update-browserslist-db`
- Regenerate `package-lock.json` with the repository's pinned npm version.
- Update release-profile validation and documentation so the semantic query is authoritative.
- Keep resolved targets available as generated build evidence if useful, but not as the public browser-support contract.

## Reproducibility contract

The checked-in lockfile fixes the Browserslist data used by a commit. Browser-data updates must arrive through reviewed lockfile changes; dedicated package overrides are not required solely to keep `npm ci` deterministic.

## Acceptance criteria

- [ ] `package.json` uses only the undated `baseline widely available` browser query.
- [ ] All four browser-data overrides are removed.
- [ ] A clean `npm ci` reproduces the checked-in target resolution and artifacts.
- [ ] Browser-profile validation checks the semantic contract without requiring manually maintained public version ceilings.
- [ ] Build, artifact-integrity, and bundle-size checks pass, with any output or size change recorded.
- [ ] Documentation states that the target is a minimum capability floor and does not exclude newer browsers.

## Runtime cost boundary

This is build and release tooling only. No browser-profile validation code or Browserslist data may enter runtime entrypoints.

PR #247, opened 2026-08-29
fix(performance): allow profile digest transitions

### LINKED ISSUE #246: v5: permit synchronized release-profile digest transitions
## Goal

Allow an intentional release-profile revision to advance its SHA-256 in the candidate performance contract without weakening the exact-base performance authority.

Related to #127 and required by #238 / PR #243.

## Problem

The candidate-contract path correctly measures a pull request with its candidate performance contract, but the exact-base growth evaluator currently rejects every toolchain difference. That makes a synchronized release-profile update logically impossible even after the required workflow pin is refreshed.

## Contract

- Permit an unchanged digest or a transition only at toolchain.releaseProfile.sha256.
- Require the candidate digest to be a lowercase 64-character SHA-256 value.
- Reject every other toolchain addition, removal, or mutation.
- Continue relying on the exact-base bundle measurement to hash the actual candidate release-profile file and reject mismatches.
- Preserve all existing exact-base threshold, baseline, approval, artifact, graph, and budget-amendment enforcement.

## Acceptance criteria

- [ ] A SHA-only release-profile contract transition passes candidate contract validation.
- [ ] Malformed digests and all other toolchain mutations fail closed.
- [ ] A candidate report containing a release-profile digest mismatch remains rejected.
- [ ] A clean SHA-only candidate report passes with no new-production delta.
- [ ] The exception composes with an independently authorized budget-ceiling amendment.
- [ ] Focused performance tests, full test coverage, docs, lint, and diff checks pass.

## Non-goals

- Changing the release profile itself.
- Updating GitHub rules or the pinned required workflow.
- Changing runtime artifacts, package metadata, or performance budgets.

paulfalgout:
Closes #246
Related #127
Related #238

## What

- Permit a candidate performance contract to change only toolchain.releaseProfile.sha256 when the new value is a valid lowercase SHA-256.
- Keep every other toolchain field bound to the exact base through a full deep comparison.
- Test both halves of the trust boundary: candidate-contract validation and the exact-base measurer hashing the actual release-profile file.

## Why

The candidate measurement path can validate an intentional release-profile update, but the exact-base growth evaluator rejected every toolchain difference. That left PR #243 logically blocked even after the deferred ruleset workflow-pin refresh. This narrow transition preserves exact-base authority while allowing the profile digest to advance with its file.

## Scope

This changes performance-governance tooling, its tests, and the performance contract documentation. It does not change the release profile, active performance budget, workflows, GitHub rules, runtime source, distribution output, or package metadata.

## Deployment Impact

No redeploy, package publication, tag, or release is needed. The change affects only repository validation.

## Testing

- npm run test:performance — 91 passed
- npm run coverage — 70 files, 1,236 tests, 100% statements/branches/functions/lines; 19 agent benchmark contract tests passed
- npm run lint:ci
- npm run check:dist
- git diff HEAD^ --check

---
## Summary by cubic
Permits candidate performance contracts to update only `toolchain.releaseProfile.sha256` to a valid lowercase SHA-256 while keeping every other toolchain field bound to the exact base, so intentional release-profile digest updates no longer fail the growth evaluator. The candidate report must still prove the new digest matches the actual release-profile file.

<sup>Written for commit c7e69ca28ce2ca69c4be07ffb375ff9d307855af. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/config/performance-growth-approval.mjs
+++ b/config/performance-growth-approval.mjs
@@ -313,6 +313,30 @@ function exactObjectFields(value, fields) {
     isDeepStrictEqual(Object.keys(value).sort(), [...fields].sort());
 }
 
+function validReleaseProfileTransition(authorityToolchain, candidateToolchain) {
+  const authorityProfile = authorityToolchain?.releaseProfile;
+  const candidateProfile = candidateToolchain?.releaseProfile;
+  if (!authorityToolchain || typeof authorityToolchain !== 'object' ||
+      Array.isArray(authorityToolchain) || !candidateToolchain ||
+      typeof candidateToolchain !== 'object' || Array.isArray(candidateToolchain) ||
+      !authorityProfile || typeof authorityProfile !== 'object' ||
+      Array.isArray(authorityProfile) || !candidateProfile ||
+      typeof candidateProfile !== 'object' || Array.isArray(candidateProfile) ||
+      typeof candidateProfile.sha256 !== 'string' ||
+      !/^[a-f\d]{64}$/.test(candidateProfile.sha256)) {
+    return false;
+  }
+
+  const expectedToolchain = {
+    ...authorityToolchain,
+    releaseProfile: {
+      ...authorityProfile,
+      sha256: candidateProfile.sha256,
+    },
+  };
+  return isDeepStrictEqual(candidateToolchain, expectedToolchain);
+}
+
 export function validateCandidateGrowthContract(authority, candidate, { budgetAmendment } = {}) {
   const violations = [];
   if (!authority || typeof authority !== 'object' || !candidate || typeof candidate !== 'object') {
@@ -325,6 +349,12 @@ export function validateCandidateGrowthContract(authority, candidate, { budgetAm
   }
   for (const key of authorityKeys) {
     if (key === 'runtimeArtifacts' || key === 'productionGraphs') {
+      continue;
+    }
+    if (key === 'toolchain') {
+      if (!validReleaseProfileTransition(authority.toolchain, candidate.toolchain)) {
+        violations.push('Candidate performance contract changes exact-base toolchain beyond releaseProfile.sha256');
+      }
       continue;
     }
     if (key === 'baseline' && budgetAmendment?.status === 'accepted' &&
--- a/docs/performance-baselines.md
+++ b/docs/performance-baselines.md
@@ -252,13 +252,15 @@ approved path list must exactly match all existing artifacts above the strict
 greater-than-one-percent threshold. New subpaths and new artifact path/size pairs must
 exactly match the additive candidate contract and measured report. A candidate may
 only add runtime artifact and production graph entries: it cannot change the base
-thresholds, allowlist, baseline, ceiling, toolchain, forbidden modules, resource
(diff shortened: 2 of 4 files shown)


paulfalgout:
**Approve** at `c7e69ca2` (the SHA I fetched). No CI failures. This is the right shape of fix for the digest half of the problem, and I verified it against the actually-blocked PR rather than in the abstract.

Running the new `validateCandidateGrowthContract` with master's contract as authority and **#243's real contract** as candidate:

```
violations: 0                                      -> #243 is genuinely unblocked
same candidate but node bumped to 23.0.0           -> rejected
same candidate with the digest uppercased          -> rejected
```

The narrowing is tight in the way that matters. Rebuilding `expectedToolchain` as the authority object with only `sha256` swapped in, then `isDeepStrictEqual` against the whole candidate toolchain, means the permitted delta is exactly one field — and because deep-strict-equal compares key sets, an *added* or *removed* toolchain key fails too, which a field-by-field comparison would have missed. `lockedDependencies`, `canonicalHost`, `commands`, and the npm/node/lockfileVersion pins all stay bound to the base.

**The part I'd single out is the test in `contract.test.mjs`.** Setting the digest to `'0'.repeat(64)` picks a value that *passes* the newly relaxed `/^[a-f\d]{64}$/` check, then asserts `validateToolchain` still rejects it against the real file hash. That's testing the composition rather than the units: it demonstrates that relaxing layer one doesn't open a hole, because layer two independently binds the digest to the actual `release-profile.json` contents. A candidate can now advance the digest, but only to the true hash of the file it ships. That's the property that makes this safe, and it's good that it's pinned by a test rather than left as an argument in the PR description.

One scope note so it isn't mistaken for a broader fix: this resolves the digest class only. **#224 stays blocked**, because its failure comes from a different path — `bundle-size.mjs:201`, the measured-size-versus-ceiling enforcement, not candidate-contract validation. The `budgetAmendment.mode === 'consume'` branch visible in this diff shows the contract side of consumption is already handled; what #224 hits is the measurement side, where the candidate's 51995 is still compared against the base's 51975. That needs its own change.

paulfalgout:
Correction to my review above. I wrote "violations: 0 → #243 is genuinely unblocked," which overstated what I checked.

The verification itself holds: `validateCandidateGrowthContract` does now accept #243's contract, and rejects the adversarial variants. But #243's CI still fails after rebasing onto this merge, with the same error as before — it originates in `validateToolchain` (`config/bundle-size.mjs:229`), which this PR doesn't change. I verified one layer and generalized to overall CI.

This PR's own change is unaffected by that; the trust-boundary reasoning and the `'0'.repeat(64)` composition test stand. Details and the follow-up evidence are on #243.

PR #248, opened 2026-08-29
test(collection): characterize slice helpers

paulfalgout:
Related #241

## What

- Characterize ChildViewContainer.initial, rest, and without.
- Pin default and count behavior, exact child View identity and order, empty-container results, and fresh-array isolation.
- Document the ordinary v5 contract without adopting Underscore coercion or guard quirks.

## Why

Issue #241 removes the required Underscore runtime dependency while retaining useful collection ergonomics. These tests make the remaining non-callback slice and identity behavior executable before the proxy is replaced.

## Scope

This changes only ChildViewContainer unit tests and CollectionView reference documentation. It does not change runtime source, aliases, package metadata, performance configuration, distribution artifacts, or callback helper semantics.

## Deployment Impact

No redeploy or package publication is needed. This is test and documentation coverage only.

## Testing

- npm test -- test/unit/child-view-container.spec.js --reporter=dot — 59 passed
- npm run coverage — 70 files, 1,246 tests, 100% statements/branches/functions/lines; 19 agent benchmark contract tests passed
- npm run lint:ci
- npm run check:dist
- git diff --check

---
## Summary by cubic
Adds unit tests and reference docs that pin down the contract for `ChildViewContainer`'s `initial`, `rest`, and `without` slice helpers ahead of removing the Underscore runtime dependency (issue #241).

The tests document default and explicit nonnegative count behavior, exact child View identity matching, empty-container results, and fresh-array isolation so callers can't mutate the container through the returned array. The docs describe these helpers in the CollectionView reference without adopting Underscore coercion or guard quirks. No runtime code, aliases, package metadata, or distribution artifacts change.

<sup>Written for commit 006e635eaa2f0cfe92e596b3a2b9d009819f9d0e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -660,6 +660,18 @@ that many Views from the corresponding end of the container. A count of `0`
 returns `[]`. For an empty container, the no-count forms return `undefined` and
 the count forms return `[]`.
 
+`initial(count = 1)` and `rest(count = 1)` return new ordered arrays after
+excluding `count` Views from the end or start of the container, respectively.
+The count is a nonnegative integer: `0` returns a new array of every child View,
+and a count greater than or equal to the container length returns `[]`. An empty
+container also returns `[]`.
+
+`without(...views)` returns a new ordered array excluding the exact child View
+instances supplied. Models and lookalike objects do not exclude their associated
+Views. With no arguments it returns a new array of every child View. Changing the
+returned array's membership or order does not change the container. An empty
+container returns `[]`.
+
 `children.isEmpty()` reports whether the child container currently has zero
 Views. It is distinct from the overridable `CollectionView#isEmpty()` method,
 which controls whether a CollectionView renders its `emptyView`.
--- a/test/unit/child-view-container.spec.js
+++ b/test/unit/child-view-container.spec.js
@@ -127,6 +127,80 @@ describe('#ChildViewContainer', function() {
       });
     });
 
+    describe('#initial', function() {
+      it('returns a new ordered array without the last child view by default', function() {
+        const initialViews = container.initial();
+
+        expect(initialViews).to.have.lengthOf(2);
+        expect(initialViews[0]).to.equal(views[0]);
+        expect(initialViews[1]).to.equal(views[1]);
+        expect(container.initial()).to.not.equal(initialViews);
+      });
+
+      it('excludes a nonnegative integer count from the end', function() {
+        const oneView = container.initial(2);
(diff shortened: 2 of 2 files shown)


paulfalgout:
**Approve** at `006e635e` (the SHA I fetched). No CI failures. All thirteen documented claims check out against the shipped package:

```
initial()  [0,1]      initial(2) [0]    initial(0) [0,1,2]    initial(5) []
rest()     [1,2]      rest(2)    [2]    rest(0)    [0,1,2]    rest(5)    []
without(v0,v2) [1]    without(model) [0,1,2]      without() [0,1,2]
fresh arrays on every call; empty container -> [] for all three
```

Deciding to document the nonnegative-integer contract rather than inherit Underscore's coercion behavior is the right call, and I'd have made the same one. **But I think the excluded behavior deserves an explicit test rather than only a sentence in the PR description**, because some of what's being left behind is stranger than "coercion quirk" suggests. Current behavior:

```
initial("2")   [0]        <- string coerced
initial(1.5)   [0]        <- fractional truncated
initial(-1)    [0,1,2]    <- negative returns EVERYTHING
rest(-1)       [2]        <- negative returns ONLY THE LAST
```

`initial(-1)` and `rest(-1)` disagree completely — one clamps to the full container, the other slices from the end. That asymmetry isn't a coherent contract anyone would choose; it falls out of `_.initial` using `Math.max(0, length - n)` while `_.rest` passes `n` straight to `slice`. The new tests don't reference negative or coerced counts at all (I grepped the diff: zero matches), so right now the replacement can preserve this asymmetry, normalize it, or throw, and nothing in the suite reacts either way.

That's the one gap I'd close. Everything else in this program has moved claims from prose into executable form; "these inputs are out of contract" is a claim too. A test asserting the intended treatment — whether that's a thrown diagnostic, clamping, or normalization to a nonnegative integer — would make the boundary as enforceable as the contract inside it, and would stop `rest(-1) → [2]` from being accidentally preserved by an implementation that happens to reach for `slice`.

Not a blocker: the documented surface is correct and now pinned, and this is a decision about the next slice rather than a defect in this one.

PR #249, opened 2026-08-29
test(collection): characterize predicate helpers

paulfalgout:
Related #241

## What
- Characterize `find`, `filter`, `reject`, `every`, `some`, and `partition` on `ChildViewContainer`.
- Document function-only predicates, `(view, index)` callback arguments, optional callback context, ordering, short-circuiting, and fresh result arrays.
- State that structurally mutating the container during predicate traversal is unsupported.

## Why
These tests make the current public collection-helper behavior explicit before Underscore-backed internals are replaced. The contract keeps useful Marionette APIs while avoiding accidental promises around private storage, shorthand predicates, or mutation snapshots.

## Scope
- Test and current-reference documentation changes only.
- No runtime behavior, package dependencies, build configuration, or generated distribution artifacts change.
- Traversal helpers and the Underscore implementation replacement remain separate work under #241.

## Deployment Impact
No deployment or redeploy is required. This changes tests and documentation only.

## Testing
- `npx mocha --require test/setup/node.js test/unit/child-view-container.spec.js` — 74 passing.
- `npm run coverage` — 70 files and 1,261 tests; 100% statements, branches, functions, and lines.
- `npm run test:agent-benchmark-contract` — 19 passing.
- `npm run docs` — 34 HTML files and 320 links checked.
- `npm run lint`
- `npm run check:dist`
- `npm run size` — 51,878 bytes, unchanged.
- `git diff HEAD^ --check`

---
## Summary by cubic
Adds tests and documentation that pin down the current behavior of `find`, `filter`, `reject`, `every`, `some`, and `partition` on `ChildViewContainer`, so the Underscore-backed internals can be replaced under #241 without changing public behavior.

- Predicates are function-only; property-name and object-matcher shorthand are not supported.
- Predicates receive `(view, index)` with `this` bound to `context` when provided; additional callback arguments aren't public API.
- Tests and docs cover iteration order, short-circuiting, fresh result arrays, and empty-container results.
- Structurally mutating the container during traversal is documented as unsupported.
- No runtime behavior, dependencies, build configuration, or generated artifacts change; this is tests and docs only.

<sup>Written for commit ee37d292b444be203ce8206e0e1d20cc1bc4bd3f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/249?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -620,11 +620,11 @@ collection functions, including:
 * [each](http://underscorejs.org/#each)
 * [map](http://underscorejs.org/#map)
 * [reduce](http://underscorejs.org/#reduce)
-* [find](http://underscorejs.org/#find)
-* [filter](http://underscorejs.org/#filter)
-* [reject](http://underscorejs.org/#reject)
-* [every](http://underscorejs.org/#every)
-* [some](http://underscorejs.org/#some)
+* `find`
+* `filter`
+* `reject`
+* `every`
+* `some`
 * [contains](http://underscorejs.org/#contains)
 * [invoke](http://underscorejs.org/#invoke)
 * [toArray](http://underscorejs.org/#toArray)
@@ -635,7 +635,7 @@ collection functions, including:
 * [without](http://underscorejs.org/#without)
 * [isEmpty](http://underscorejs.org/#isEmpty)
 * [pluck](http://underscorejs.org/#pluck)
-* [partition](http://underscorejs.org/#partition)
+* `partition`
 
 These methods can be called directly on the container, to iterate and process
 the views held by the container.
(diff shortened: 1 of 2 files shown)


paulfalgout:
**Approve** at `ee37d292` (the SHA I fetched). No CI failures. Verified against the shipped package — short-circuiting on `find`/`every`/`some` all stop at one call, `context` binds `this`, `partition` returns a two-element split, index order is `[0,1,2]`, and result arrays are fresh per call.

**The doc wording deserves specific credit, because two of these rules are narrowings rather than characterizations.** I checked what actually ships today:

```
callback arity today          3 arguments (the third an array of the container's views)
filter({ cid: … })            works -> 1 result
filter('cid')                 works -> 3 results
```

So the current implementation *exceeds* the contract being documented. The phrasing handles that correctly rather than papering over it — "Additional callback arguments are not public API" is true today **and** stays true after the replacement regardless of arity, and "Marionette does not define Underscore-style property-name or object-matcher shorthand" declares the behavior undefined rather than absent. Both would have been false if written as "the predicate receives two arguments" or "shorthand throws." That precision is exactly right for a pre-refactor contract, and it's more carefully done than the equivalent in #248.

Not withholding the third argument as public API is also the substantively correct call: it currently hands the callback an array of the container's views, so promising it would expose internal storage as contract.

**One thing to carry forward rather than change here.** Because shorthand demonstrably works today, replacing the proxy is a behavior change for anyone using it, and `docs/migration-from-v4.md` currently says nothing about it (zero matches for shorthand or iteratee). The new prose declares the contract for readers starting fresh, but an existing caller of `children.filter({ cid })` gets no signal that their code stops working. Worth a migration-note line when #241 actually lands the replacement — that's the moment it becomes a break, and this PR is the record of what the behavior was beforehand.

PR #258, opened 2026-08-29
feat(collection): own child container helpers

paulfalgout:
Related #241

## What
- Replace Underscore-proxied `CollectionView#children` helpers with 19 Marionette-owned implementations.
- Remove the undocumented `forEach`, `detect`, `select`, `all`, `any`, and `include` aliases and add prototype-based native iteration.
- Add MN0024/MN0025 diagnostics, focused tests, migration guidance, and regenerated distributions.

## Why
The public child-container contract should be explicit, deterministic, and understandable without relying on Underscore iteratee shorthand or hidden proxy behavior. This is the first bounded runtime slice of #241; it keeps the private Underscore-backed sorting helpers for a later change.

## Scope
Impacts `CollectionView#children` collection helpers, their diagnostics, documentation, tests, and generated bundles. Callback helpers now require functions and expose only `(view, index)`; `pluck` reads direct View properties; `invoke` accepts direct string method names; positional counts require nonnegative integers. Existing private sorting behavior and the package peer dependency are intentionally unchanged.

## Deployment Impact
No website, npm package, tag, release, or application deployment is performed by this PR.

## Testing
- `npm test -- --run test/unit/child-view-container.spec.js test/unit/collection-view.spec.js` (91 focused tests)
- `npm run coverage` (1,325 tests; 100% statements, branches, functions, and lines)
- `npm run test:agent-benchmark-contract` (19 tests)
- `npm run test:performance` (92 tests)
- `npm run check:diagnostics` (25 entries)
- `npm run lint:ci`
- `npm run docs:check`
- `npm run check:dist`
- `npm run test:source`
- `npm run test:fixtures`
- `npm run size` (54.28 kB cumulative against the 75.00 kB development ceiling)

---
## Summary by cubic
Replaces the Underscore-proxied `CollectionView#children` helpers with 19 Marionette-owned implementations, removing the undocumented aliases and adding prototype-based native iteration. This is the first runtime slice of #241; private sorting helpers remain unchanged for later.

**Behavior Changes**
- Callback helpers now require functions and expose only `(view, index)`.
- `pluck` reads direct View properties; `invoke` accepts direct string method names.
- Positional counts require nonnegative integers.
- Existing private sorting behavior and the `underscore` peer dependency are unchanged.

**Diagnostics and Migration**
- Adds MN0024 and MN0025 runtime diagnostics with remediation guidance.
- Updates the v4 migration guide with the new child-container contract.

[...]

PR #259, opened 2026-08-29
refactor(collection): own child container sorting

paulfalgout:
Related #241

## What
- Replace the final Underscore-backed `partial` and `sortBy` behavior in ChildViewContainer with a local stable criteria sort.
- Preserve binary comparator behavior while making unary/string criteria evaluation once-only, stable, undefined-last, and exception-safe.
- Add focused sorting coverage, public reference guidance, and regenerated distributions.

## Why
This completes the owned ChildViewContainer implementation started in #258 and removes that production module’s final Underscore import. The explicit sorting contract is easier to inspect and reason about while preserving the useful Underscore 1.13 ordering semantics.

## Scope
Impacts only ChildViewContainer sorting, its tests, CollectionView reference documentation, and generated main bundles. The same `children` container remains in use. Binary `Array#sort` comparator behavior remains unchanged. Other production Underscore imports, package metadata, and the peer dependency are intentionally unchanged.

## Deployment Impact
No website, npm package, tag, release, or application deployment is performed by this PR.

## Testing
- `npm test -- --run test/unit/child-view-container.spec.js` — 100 tests passed
- `npm run coverage` — 1,334 tests; 100% statements, branches, functions, and lines
- `npm run test:agent-benchmark-contract` — 19 tests passed
- `npm run test:performance` — 92 tests passed
- `npm run check:diagnostics` — 25 entries validated
- `npm run lint:ci`
- `npm run docs:check` — 37 pages, 38 HTML files, 352 links
- `npm run check:dist`
- `npm run test:source`
- `npm run test:fixtures`
- Independent 10,000-case criterion-order comparison against Underscore 1.13 — no mismatch
- `npm run size` — 54.83 kB cumulative against the 75.00 kB development ceiling (+541 B / +1.00% from PR base)

---
## Summary by cubic
Replaces the final `underscore`-backed `partial`/`sortBy` usage in `ChildViewContainer` with a local stable criteria sort. This completes the owned container implementation started in #258 and addresses #241.

**Behavior**
- String and single-argument comparators evaluate each child’s criterion once, sort stably, and place `undefined` criteria last.
- Equal, `NaN`, or otherwise incomparable criteria keep their existing order.
- Two-argument comparators keep native `Array#sort` behavior.
- If evaluating or comparing a criterion throws, the error propagates without changing child order.

**Scope**
- Affects only `ChildViewContainer` sorting, its tests, `CollectionView` docs, and generated bundles.
- The same `children` container reference remains in use.

<sup>Written for commit 44da88de67d76092201d59d6efff63ac6572da47. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
```json
{
  "schemaVersion": 1,
  "headSha": "44da88de67d76092201d59d6efff63ac6572da47",
  "issueUrl": "https://github.com/marionettejs/marionette/issues/127",
  "approvedPaths": [
    "dist/marionette.cjs",
    "dist/marionette.js",
    "dist/marionette.umd.js"
  ],
  "evidenceUrls": [
    "https://github.com/marionettejs/marionette/issues/127#issuecomment-5463132559"
  ]
}
```

paulfalgout:
## Code-review approval — exact head `44da88de`

**Approved.** `Bundle size` shows only the exact-head maintainer approval outstanding; no contract violations. `+541 B`, 20,174 remaining under the envelope.

**Sorting rewrites are where subtle regressions hide, so I differential-tested rather than read.** With criteria `k = [2,1,2,undefined,1]` across ids `a…e`:

```
                      master            p259
unary criteria        b,e,a,c,d         b,e,a,c,d      criteria calls: 5 -> 5
string criteria       a,b,c,d,e         a,b,c,d,e
binary comparator     a,b,c,d,e         a,b,c,d,e
criteria throws       "boom"            "boom"
```

That ordering confirms all four claims at once: the two tie groups (`b,e` at 1 and `a,c` at 2) each preserve insertion order, so it's **stable**; `d` lands last, so **undefined-last** holds; the criteria function is invoked exactly **5 times for 5 items** rather than on every comparison, so **once-only** evaluation is real and not incidental; and a throwing criteria still propagates unchanged. Binary comparators are untouched, as scoped.

Once-only evaluation is worth calling out specifically — it's the property most easily lost when hand-rolling a criteria sort, since the obvious implementation calls the criteria inside the comparator and silently degrades to O(n log n) invocations. A criteria function with side effects or real cost would have noticed.

**`modules/child-view-container.js` is now Underscore-free** — `import { partial, sortBy } from 'underscore'` is gone and the module has no remaining import. That's the first production module fully off the dependency.

For scope calibration on the rest of #241, **26 production modules still import Underscore**:

```
mixins/         behaviors, common, delegate-entity-events, events, radio, requests,
                template-render, ui, view-events, view
modules/        application, behavior, collection-view, object, radio, view-region,
                common/{bind-events, bind-requests, merge-options, monitor-view-events,
                        normalize-methods, trigger-method}
utils/          build-event-args, error, extend, once-wrap
```

Two slices have cost +2,019 B combined against a ~6,904 B peer, so the trajectory still favours the eventual dependency-inclusive comparison — but the peer dependency stays declared until the last of those 26 clears, and consumers ship both until then. Also still worth wiring before that final slice: the `baselineExternalImports` drift I flagged on #242 is measured but never raises a violation, so "no Underscore import" isn't yet machine-enforced.

PR #260, opened 2026-08-29
refactor(core): own pseudo-class extension boundary

paulfalgout:
Related #241

## What
- Add narrow owned assignment primitives for inherited composition, own-only public inputs, and prototype-safe dynamic keys.
- Remove Underscore from Marionette’s shared pseudo-class `extend` implementation.
- Define and test the extension contract across all six exported pseudo-classes, update migration/reference docs, and regenerate distributions.

## Why
The `.extend()` boundary is shared framework infrastructure and should have explicit, agent-readable property semantics without relying on Underscore. Owning this boundary also establishes safe assignment primitives for later, separately characterized configuration slices.

## Scope
Impacts `utils/extend.js`, the new internal assignment helper, focused tests, class-extension documentation, migration guidance, and generated main bundles. Public `protoProps` and `staticProps` now contribute own enumerable string keys only; parent-constructor statics retain inherited-enumerable composition. Configuration/options merge sites are intentionally unchanged.

## Deployment Impact
No website, npm package, tag, release, or application deployment is performed by this PR.

## Testing
- Focused assignment/extension specs — 12 tests passed
- `npm run coverage` — 1,346 tests; 100% statements, branches, functions, and lines
- `npm run test:agent-benchmark-contract` — 19 tests passed
- `npm run test:performance` — 92 tests passed
- `npm run check:diagnostics` — 25 entries validated
- `npm run lint:ci`
- `npm run docs:check` — 37 pages, 38 HTML files, 352 links
- `npm run check:dist`
- `npm run test:source`
- `npm run test:fixtures`
- `npm run size` — 55.44 kB cumulative against the 75.00 kB development ceiling (+618 B / +1.13% from PR base)

---
## Summary by cubic
Replaces Underscore's `extend` with internal assignment helpers so Marionette owns its pseudo-class extension boundary (related #241). `extend` now copies only own enumerable string keys from `protoProps` and `staticProps`; inherited properties on those inputs are ignored, and parent-constructor statics still inherit as before.

**Migration**
- The `staticProps` hash no longer contributes inherited enumerable properties to the child constructor; move them to `protoProps` or the parent instead.
- `__proto__` keys are now defined as own properties rather than mutating the prototype chain.

<sup>Written for commit a3e87084b7a90c1e6f6ec1c5a99b84f5a35debd5. Summary will update on new commits.</sup>

[...]

PR #264, opened 2026-08-29
fix(dom): own attribute inputs safely

paulfalgout:
Related #241

## What

- Clone generated View element attributes from own enumerable string keys only.
- Preserve DOM property-versus-attribute assignment while safely defining a literal own `__proto__` property.
- Document the generated-element attribute contract and the v4 migration impact for inherited attribute maps.

## Why

Marionette's remaining Underscore-backed attribute boundaries admitted inherited enumerable input and could route a literal `__proto__` key through prototype mutation. This slice makes those boundaries explicit and safe without changing normal DOM assignment behavior.

## Scope

This affects generated View and CollectionView element attributes and direct `DomApi.setAttributes` calls. Existing property-versus-attribute routing and conditional `id`/`className` precedence are preserved. UI normalization, event registries, identifiers, collection helpers, and package removal are intentionally unchanged.

## Deployment Impact

## Testing

- `npx mocha --require test/setup/node.js test/unit/config/dom.spec.js test/unit/view-el.spec.js` (36 passing)
- `npm run coverage` (1364 tests, 100% coverage)
- `npm run lint:ci`
- `npm run check:diagnostics`
- `npm run docs:check`
- `npm run test:performance`
- `npm run check:dist`
- `npm run test:source`
- `npm run test:fixtures`

---
## Summary by cubic
Clones generated View element attributes from own enumerable string keys only, so inherited, symbol, and non-enumerable properties no longer leak into generated elements. A literal own `__proto__` key is defined as an own element property instead of mutating the element's prototype, and existing DOM property-versus-attribute routing stays intact.

**Migration**
- Views that relied on inherited `attributes` entries must move them onto the View's own `attributes` object.

<sup>Written for commit 1d113f2645e29cec2479fc5860d9b70ca61609e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/264?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `1d113f26`

**Approved. No CI failures, +1 byte.** That's a remarkable price for what it fixes.

**The `__proto__` case destroys the element on master, and I nearly missed it.** My first probe passed a *string* value and saw no difference between branches — because `Object.prototype.__proto__`'s setter silently ignores primitives. Your test uses an object value, which is what actually triggers it:

```
attributes hash with an own, enumerable, object-valued __proto__:

master   TypeError: el.setAttribute is not a function
p264     prototypeIntact=true  ownProtoKey=true  element still usable  other="o"
```

Assigning it replaces the element's `[[Prototype]]`, taking every DOM method with it — and it throws mid-way through `setAttributes`, so the element is left partially configured. Same shape as the extend-boundary bug in #260, but reached through the `attributes` hash, which is a far more likely landing spot for externally-derived config.

The string-versus-object distinction is worth recording somewhere durable, because it's exactly why this survived: the obvious test case is the one that can't fail. Your fixture gets it right by using `Object.defineProperty` with an object value.

**The declared narrowing checks out, and the things that should not have moved didn't:**

```
                                    master          p264
inherited attribute                 applied         skipped
own attribute                       applied         applied
id / class from attributes          aid / acls      aid / acls
property routing (input.value)      property        property
attribute routing (data-x)          attribute       attribute
```

Property-versus-attribute routing is the one I'd have been most worried about — a change there would be invisible in most tests and break form controls specifically — and `value` still lands as a property while `data-x` lands as an attribute on both.

paulfalgout:
**Correction to my review above — the finding holds but I reported the wrong symptom.**

I compared against a `dist/marionette.cjs` snapshot I had cached from `455f8bf7`, roughly ten merges before this PR's actual base, and labeled it "master". Re-running against the real base `f070286b`:

```
                                        455f8bf7 (what I used)      f070286b (real base)      this PR
inherited attribute                     applied                     applied                   skipped
object-valued __proto__ attribute       TypeError thrown            no throw                  no throw
element still functional afterwards     n/a — threw first           NO                        yes
```

So the defect is real and the mechanism is exactly as I described — the element's `[[Prototype]]` is replaced and its DOM methods go with it — but at the real base it happens **silently**. `setAttributes` completes without error and hands back an element whose `setAttribute` is gone. My write-up said it "throws mid-way through `setAttributes`"; that was true of the older snapshot, not of what you actually merged onto.

If anything the real behavior is worse than what I reported: a loud `TypeError` at least points at the call. Silent corruption surfaces later, somewhere unrelated.

The inherited-attribute narrowing I reported is unaffected — verified against the real base above.

Apologies for the noise; the cached snapshot was my error and I've rebuilt the baseline from `HEAD` for subsequent reviews.

PR #265, opened 2026-08-29
fix(events): own event name registry

### LINKED ISSUE #241: v5: remove the required Underscore runtime dependency

paulfalgout:
Related #241

## What

- Make event registration, removal, and dispatch consult exact own `_rdEvents` entries.
- Safely support `constructor`, `toString`, and literal `__proto__` event names without changing the plain event-store prototype.
- Characterize collision names, `all` wildcard parity, listener cleanup, and inherited-handler suppression in executable tests and migration/reference docs.

## Why

Inherited `Object.prototype` values were mistaken for handler arrays, so registering names such as `toString`, `constructor`, or `__proto__` could throw. Triggering also consulted inherited named and `all` handlers. Event names should be explicit registrations, independent of the private store's prototype.

## Scope

This affects only the `_rdEvents` name registry in Marionette's Events mixin. Plain-object storage, `all` snapshot/reentrancy behavior, event maps and space-separated names, callback context, listener cleanup, and literal `all` double-dispatch parity are preserved. Underscore traversal, `uniqueId`, `once` wrapping, listening-ID registries, UI maps, and package removal are intentionally unchanged.

## Deployment Impact

## Testing

- Focused Events and Backbone parity suite (28 passing)
- `npm run coverage` (1371 tests, 100% coverage)
- `npm run lint:ci`
- `npm run docs:check`
- `npm run check:diagnostics`
- `npm run test:source`
- `npm run check:dist`
- `npm run test:fixtures`

---
## Summary by cubic
Fixes #241 by making event registration, removal, and dispatch consult only explicit own `_rdEvents` entries. Names like `constructor`, `toString`, and `__proto__` are now treated as ordinary event names instead of resolving inherited `Object.prototype` values, which previously caused registration to throw and triggering to fire inherited handlers.

- New handler arrays are written with a safe setter that defines `__proto__` without altering the event store's prototype.
- Tests cover collision names, `all` wildcard parity, listener cleanup, and inherited-handler suppression; docs and migration notes are updated.
- No redeploy needed beyond adopting the resulting package; existing `all` reentrancy and listener cleanup behavior are unchanged.

<sup>Written for commit c0b68cab443cc3d347c3356ebf89974ebe2bae2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/265?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `c0b68cab`

**Approved. No CI failures.** +100 B, 19,173 remaining.

**There are two distinct failure modes on master here, and the second is the one I'd emphasize.**

Registration throws outright:

```
o.on(name, fn)        master                                       p265
toString              TypeError: handlers.push is not a function   ok, fires 1, off works
constructor           TypeError: handlers.push is not a function   ok, fires 1, off works
hasOwnProperty        TypeError: handlers.push is not a function   ok, fires 1, off works
__proto__             TypeError: handlers.push is not a function   ok, fires 1, off works
normal                ok                                           ok
```

`this._rdEvents[name]` returns the inherited `Object.prototype.toString`, the code takes it for an existing handler array, and `.push` explodes.

But dispatch fails independently, **with nothing registered at all**:

```
o.trigger('constructor')   master: TypeError: Cannot destructure property 'callback' of…
o.trigger('toString')      master: no error
                           p265:   both fine
```

That's the more dangerous half, because it doesn't require anyone to have used the name as an event. An application that derives event names from data — `trigger(model.get('type'))`, a channel name, a server-supplied action — crashes on the value `"constructor"` even though no listener was ever registered for it. And the fact that `toString` survives while `constructor` throws is precisely the inconsistency your description points at: which collision names are fatal depends on what shape the inherited value happens to have.

**Preserved, checked explicitly**: `all` wildcard parity (`all=1, named=1` on both), space-separated names (`'a b'` → 2 hits), and callback context binding. Those are the parts most easily lost when reworking a dispatcher, so it's good they're pinned in the new tests rather than left implied.

This is the third registry in a row where own-key discipline surfaced live bugs rather than just tidying semantics — after Requests in #263 and DOM attributes in #264. `mixins/ui.js` and `mixins/behaviors.js` are the remaining ones with the same shape, and on this hit rate I'd expect them to yield something too.

PR #267, opened 2026-08-29
refactor(events): own event argument traversal

paulfalgout:
Related #241

## What

- Remove the direct Underscore dependency from `buildEventArgs`.
- Replace object-map traversal with `Object.keys` and split-name traversal with direct indexed loops.
- Add focused parity coverage for recursive maps, arrays, ordering, special keys, callback context, falsy values, whitespace, and Symbol behavior.

## Why

`buildEventArgs` is the shared argument normalizer for Marionette Events and Requests overloads. Owning this narrow traversal removes one production Underscore import while keeping the accepted overload semantics explicit and executable.

## Scope

This affects only `utils/build-event-args.js`, focused tests, and generated distributions. Own-enumerable-string map selection, Object.keys ordering, array-as-map holes and extra keys, recursive space splitting, `context || callback` fallback, listener propagation, falsy passthrough, whitespace behavior, and Symbol errors are preserved. Event and request registries, listener lifecycle, UI, identifiers, `once`, EventDelegator, and package metadata are intentionally unchanged.

## Deployment Impact

## Testing

- Focused helper, Events, and Requests suites (39 passing)
- `npm run coverage` (1392 tests, 100% coverage)
- `npm run lint:ci`
- `npm run docs:check`
- `npm run check:diagnostics`
- `npm run check:release-profile`
- `npm run test:source`
- `npm run test:fixtures`
- `npm run check:dist`

---
## Summary by cubic
Refactors `buildEventArgs` to traverse event maps and space-separated names without Underscore, using `Object.keys` and indexed loops instead. This removes one production Underscore import and preserves the existing overload semantics for Marionette Events and Requests; addresses part of #241.

**Tests**
- Adds focused parity coverage for recursive maps, arrays, ordering, special keys, callback context, falsy values, whitespace, and Symbol behavior.

<sup>Written for commit 54bd1ab54cd236f04777121fe8faf74f379183c1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/267?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `54bd1ab5`

**Approved.** CI green, +135 B, 18,782 remaining. Production Underscore importers drop **25 → 24**, so this clears a module rather than only narrowing semantics.

Pure-parity claim, so I tested it as one — against a freshly built master rather than a cached artifact. Ten probes through the Events surface, all identical:

```
map form · space-split name · map-with-space-keys · context fallback · no-context-defaults-to-self
array input (hole + extra key) · falsy handler · whitespace name · symbol key · Object.keys ordering
```

And through Requests, which shares the same normalizer:

```
reply map form / reply "r3 r4" / stopReplying with a map    identical on both
```

**I paid particular attention to what might newly flow through**, since that's exactly what bit #266 — a correct boundary fix delivering inputs downstream that were previously swallowed. The two candidates here were array-likes and symbols, because `Object.keys` and Underscore's traversal can disagree on both. Neither differs: an array with a hole at index 0 and an extra non-index property yields the same two registrations either way, and a symbol-keyed map is accepted identically. Nothing new escapes this boundary.

One incidental observation, present on **both** branches so out of scope here: `stopReplying({r1: fn})` doesn't remove the handler — `request('r1')` still answers afterwards. Presumably a reference-identity mismatch in the map form. Parity is intact, and this PR isn't the place to change it, but it looks like a real gap in the map overload if nobody has logged it.

paulfalgout:
**Correction: the `stopReplying` observation I made here was wrong, and I've been repeating it since.**

I wrote that `stopReplying({r1: fn})` "doesn't remove the handler" and suggested it looked like a reference-identity gap worth logging. Re-reading `stopReducer`'s guard while reviewing #286:

```js
if (!handler ||
    callback && callback !== handler.callback && callback !== handler.callback._callback ||
    context && context !== handler.context) { continue; }
```

Declining to remove is the **correct** behavior when the supplied callback isn't the registered one — that's what makes `stopReplying(name, callback)` able to target a specific handler rather than clearing the name. My probe registered `() => "one"` and then called `stopReplying({ r1: () => {} })` with a *different* function literal, so skipping the delete was right and `request('r1')` continuing to answer was the correct outcome, not a bug.

`stopReplying('r1')` with no callback passes `callback === undefined`, the guard's `callback && …` short-circuits false, and the handler is deleted — which is the path that actually removes.

Nothing to file. Apologies for carrying it into several status summaries as an open finding; the error was mine in constructing the test, not in the code.

PR #268, opened 2026-08-29
refactor(events): own delegator traversal

paulfalgout:
Related #241

## What

- Remove EventDelegator's isolated Underscore `each` import.
- Replace `undelegateAll` traversal with an explicit cached-length indexed loop.
- Add focused parity coverage for ordering, capture resolution, live mutations, sparse arrays, and partial failures.

## Why

EventDelegator teardown was a self-contained production Underscore importer. Owning its array traversal removes that dependency while making the teardown semantics explicit for View and CollectionView lifecycle consumers.

## Scope

This affects only `config/event-delegator.js`, focused tests, and generated distributions. Initial-length capture, dense sparse-array reads, live replacement reads, ignored appends, `shouldCapture` receiver and call order, exact `removeEventListener` arguments, successful clearing, early missing-root behavior, and throw-before-clear partial teardown are preserved. Adapter overlays, event registries, View lifecycle, UI, identifiers, and package metadata are intentionally unchanged.

## Deployment Impact

## Testing

- Focused EventDelegator and View lifecycle suites (23 passing)
- `npm run coverage` (1396 tests, 100% coverage)
- `npm run lint:ci`
- `npm run docs:check`
- `npm run check:diagnostics`
- local release-profile and promotion dry-run validation
- `npm run test:source`
- `npm run test:fixtures`
- `npm run check:dist`

---
## Summary by cubic
Replaces Underscore's `each` in `EventDelegator.undelegateAll` with a native loop to remove the last isolated dependency from the teardown path. Behavior is unchanged, with focused tests added to cover the previous semantics.

**Refactors**
- Removes the `underscore` import from `config/event-delegator.js`.
- Uses a cached-length indexed loop that preserves existing teardown ordering, capture flag, and clearing behavior.
- Adds tests for ordering, live mutations, sparse arrays, and partial failures.

<sup>Written for commit 1a10c6f538123ebe7fc612cccf5f196e2b382323. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved removal of all registered event handlers, including correct handling of handler order and capture settings.
  - Ensured cleanup stops safely when a removal error occurs, preserving registrations for retry.

- **Tests**
  - Added coverage for successful cleanup, registry mutations, sparse registrations, and error scenarios.


paulfalgout:
## Code-review approval — exact head `1a10c6f5`

**Approved.** CI green, +86 B, 18,696 remaining. Production Underscore importers **26 → 25** counting `config/` — worth noting my earlier progress numbers on this issue undercounted, since I'd been grepping only `modules/ mixins/ utils/` and this file lives outside those.

The observable teardown is unchanged. Spying on `removeEventListener` through a real View destroy:

```
master  2 calls:  click/function/false , mouseover/function/false
p268    2 calls:  click/function/false , mouseover/function/false
```

Same count, same event names, same handler identity, same capture flag, same order.

**The loop is a faithful translation, and the equivalences are worth spelling out** because "replace `_.each` with a for loop" is the kind of change that quietly alters edge semantics:

- `_.each` over an array captures `length` up front and reads `events[i]` on each pass. The new loop does exactly that with `length = events.length` in the initializer — so appends during teardown are still ignored, and a live replacement at an unvisited index is still picked up.
- Sparse holes destructure `undefined` and throw in both — `_.each` passes the hole through to the same destructuring pattern rather than skipping it, so there's no behavior to lose here.
- `shouldCapture` is still called before `removeEventListener` and still receives the delegator as `this` — master got that from an arrow closure, the new code from being inline in a method.
- `events.length = 0` remains after the loop in both, so a mid-teardown throw leaves the array fully populated, including entries already removed from the DOM. Unchanged, and the new tests pin it.

That last one is the only genuinely sharp edge in this function, and I'm glad it's now covered rather than implied — a partial teardown that reports success would be very hard to diagnose from a leak downstream.

PR #269, opened 2026-08-29
refactor(events): own trigger method detection

paulfalgout:
Related #241

## What

- Remove `triggerMethod`'s direct Underscore `isFunction` import.
- Use native function detection for resolved event handlers.
- Add focused parity coverage for non-callables, dispatch ordering, handler errors, and listener errors.

## Why

`triggerMethod` is a small shared dispatch leaf whose only Underscore use is function detection. Native detection is equivalent on Marionette's supported Baseline engines and removes one production importer without changing the public dispatch contract.

## Scope

This affects only `modules/common/trigger-method.js`, focused tests, and generated distributions. Option/instance resolution precedence, handler receiver, arguments, method-before-event ordering, return values, missing and truthy non-callable handlers, handler-error suppression of event dispatch, and listener-error propagation are preserved. The preexisting method-name cache collision issue, Events registries, UI, lifecycle, and package metadata are intentionally unchanged.

## Deployment Impact

## Testing

- Focused trigger-method suite (12 passing)
- `npm run coverage` (1400 tests, 100% coverage)
- `npm run lint:ci`
- `npm run docs:check`
- `npm run check:diagnostics`
- local release-profile and promotion validation
- `npm run test:source`
- `npm run test:fixtures`
- `npm run check:dist`

---
## Summary by cubic
Replaces Underscore's `isFunction` check in `triggerMethod` with native `typeof method === 'function'` detection, removing one production dependency use without changing dispatch behavior. Adds focused test coverage for non-function handlers, ordering, and error propagation. Related to #241.

- Regenerated distribution files reflect the source change.
- Behavior for handler resolution, event ordering, return values, and error handling is preserved.

<sup>Written for commit 8de0a2c32596dbc64532fdfb3eeb14887d49f392. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/269?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `8de0a2c3`

**Approved.** CI green, importers **25 → 24**, and this is the first slice in the series to come in **smaller**: −12 B, headroom back up to 18,708. Native `typeof` detection costs less than importing `isFunction`.

Ten parity probes through the public dispatch surface, identical on both branches:

```
method-before-event ordering · handler return value · truthy non-callable ('nope') · numeric handler
missing handler · class as handler (TypeError both) · async handler (promise both)
handler error suppresses event dispatch · listener error propagates · args passthrough
```

The class-as-handler case is the one where `_.isFunction` and `typeof` could most plausibly have diverged, and they don't: both detect the class as a function, both then throw `TypeError` on the call. Async and generator handlers likewise.

**I reproduced the method-name cache collision you scoped out, in case it's useful for the follow-up issue.** It's genuinely live on master and identical here, so nothing about this PR touches it:

```
MnObject.extend({ onNormal(){return 'N'}, onConstructor(){return 'C'}, onToString(){return 'T'} })

triggerMethod('normal')       -> "N"        handler runs
triggerMethod('constructor')  -> undefined  onConstructor never called
triggerMethod('toString')     -> undefined  onToString never called
```

The handler is defined and simply never invoked — no error, no warning. That puts it in the same family as the Requests and Events registry bugs from #263 and #265, but it's arguably the most deceptive of the three, because the class visibly declares a handler for the event and the framework silently declines to call it. Warming the cache with a normal event first doesn't change it, so it isn't order-dependent.

Worth an issue on its own rather than riding along with a refactor — and worth noting that whichever slice eventually owns that cache will need the same treatment the other registries got.

PR #270, opened 2026-08-29
fix(events): isolate trigger method cache keys

paulfalgout:
Related #241

## What

- Store cached `triggerMethod` handler names in a null-prototype object.
- Restore matching method dispatch for prototype-collision event names such as `constructor`, `toString`, and `__proto__`.
- Add focused repeated-call and cross-target cache coverage.

## Why

The plain-object cache inherited `Object.prototype` values, so collision event names skipped handler-name generation and silently failed to call their documented `onEventName` methods. A null-prototype cache makes every event name an ordinary cache key without changing lookup or memoization behavior.

## Scope

This affects only the private method-name cache in `modules/common/trigger-method.js`, focused tests, and generated distributions. Module-wide cache lifetime, string-key coercion, handler/event receivers and arguments, dispatch order, return values, repeated calls, and cross-target reuse are preserved. Cache eviction, Events registries, UI, lifecycle, other Underscore importers, and package metadata are intentionally unchanged.

## Deployment Impact

## Testing

- Focused trigger-method suite (15 passing)
- `npm run coverage` (1403 tests, 100% coverage)
- `npm run lint:ci`
- `npm run docs:check`
- `npm run check:diagnostics`
- local release-profile validation
- `npm run test:source`
- `npm run test:fixtures`
- `npm run check:dist`

---
## Summary by cubic
Addresses #241 by fixing `triggerMethod` dispatch for event names that collide with `Object.prototype` properties like `constructor`, `toString`, and `__proto__`. The method-name cache now uses a null-prototype object, so these events generate their `onEventName` handler names instead of silently skipping them, with no change to cache lifetime, dispatch order, or other event names.

<sup>Written for commit d44b9463d2fc6d43a6c37560d5404e486cfd0de4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/270?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `d44b9463`

**Approved.** CI green, +16 B for a silent-dispatch bug — good trade.

Verified against master:

```
MnObject.extend({ onNormal(){…}, onConstructor(){…}, onToString(){…} })

                        master        p270
triggerMethod('normal')       "N"     "N"
triggerMethod('constructor')  undefined   "C"
triggerMethod('toString')     undefined   "T"
```

**The `__proto__` case is the one I'd highlight, because master doesn't merely fail to dispatch — it never generates a handler name at all.** I proxied the target to observe which `on*` property gets read:

```
master   generated name: (none)          result: undefined
p270     generated name: on__proto__     result: "P" once that handler is defined
```

On master the cache lookup returns the inherited `Object.prototype.__proto__` accessor, the code treats that as a cached name, and the whole name-generation step is skipped — so no property is ever read on the instance. That's why it's invisible: there's no failed lookup to observe, just nothing happening.

**Memoization survives, which was my main regression worry** given the cache is the thing being changed:

```
master  memo = ,,        (all three empty — the bug)
p270    memo = C,C,C     repeated call on same target, and a second target, both resolve
```

Colon-separated names still coerce correctly (`some:thing` → `onSomeThing`), so the null-prototype swap didn't disturb name generation itself.

Nice to see this land right after being scoped out of #269 — and worth noting the fix is the same shape as #263 and #265: the defect wasn't in the logic, it was in using a plain object as a dynamic-key store. That's now five boundaries with the identical root cause, which is a reasonable argument for making null-prototype (or the owned `setProperty`) the default for any new keyed store rather than a fix applied per site.

PR #272, opened 2026-08-29
refactor(common): own monitor view event traversal

paulfalgout:
## Linked issue

- Related #241

## What

- Replace `monitorViewEvents`' Underscore `each` dependency with a narrow local child traversal.
- Preserve array-like and object collection semantics, including captured bounds, dense indexed reads, key snapshots, and live values.
- Add focused regression coverage for null, mutation, receiver, ordering, and exception behavior.

- Makes the child-event traversal explicit in the owning module so agents can reason about lifecycle propagation without reconstructing an external polymorphic helper.

## Public behavior contract

- Canonical behavior after this PR: child lifecycle events traverse the immediate-child collection synchronously in existing order and preserve current mutation, receiver, argument, return, and error behavior.
- Superseded behavior removed, if any: the runtime dependency on Underscore `each` in this module.
- Compatibility exception and removal condition, if any: existing array-like and own-enumerable object child collections remain supported as part of the established traversal contract; no transitional fallback is added.

## Runtime cost

- Startup/hot path: remains a single linear traversal with synchronous callbacks.
- Allocations/retention: array-like traversal adds no collection allocation; object traversal snapshots own keys as Underscore did.

## Validation

Commands:

```sh
npx vitest run test/unit/common/monitor-view-events.spec.js --coverage.enabled=false
npm test
npm run test:coverage
npm run lint:ci
npm run test:source
npm run build
npm run check:dist
npm run test:dist
npm run test:fixtures
npm run test:performance
git diff --check
```

Results: 6 focused tests; 74 files and 1,415 full-suite tests; 100% coverage; 19/19 agent benchmark; 92/92 performance-contract tests; lint, source/dist validation, install fixtures, build, and diff check passed. Independent exact-head parity review found zero defects.

## Package and browser impact

- Entry points or install fixtures affected: main Marionette bundles are regenerated; package fixture tests pass.
- Browser contracts affected: none; lifecycle event behavior is unchanged.
- Production/dev/test module-graph impact: removes one production Underscore importer without adding a new external dependency.

## Rollback or deprecation

- Revert this commit to restore the Underscore traversal; no data migration or deprecation sequence is involved.

## Scope

Impacts only `monitorViewEvents`, its focused tests, and generated main distribution artifacts. It does not change unrelated view, region, event, request, or renderer behavior.

## Deployment Impact

## Review notes

- Review parity around two separate array-like length reads, initial bounds/key snapshots, live child values, dense sparse traversal, exception short-circuiting, and exact `child.triggerMethod(event, child)` receiver/arguments.

---
## Summary by cubic
Replaces the `underscore` `each` dependency in `monitorViewEvents` with a local `eachChild` traversal (related to #241). Child lifecycle events now traverse the immediate-child collection with a local helper instead of `underscore`'s `each`; behavior is unchanged.

**Refactors**
- Preserves array-like and object collection semantics, including length/key snapshots and live values.
- Adds regression tests for null collections, mutation, receiver, ordering, and exception behavior.
- Regenerates distribution bundles.

<sup>Written for commit 94ba1f67b90f59f25a59a46baa1f95e811df8f4d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `94ba1f67`

**Approved.** CI green, +174 B, 18,105 remaining. Production Underscore importers **23 → 22**.

Driving real child propagation through a Region so `_getImmediateChildren` is actually consulted, all three collection shapes behave identically:

```
                master                                              p272
array           a:before:attach,b:before:attach,a:attach,b:attach   identical
object          a:before:attach,b:before:attach,a:attach,b:attach   identical
array-like      a,b,c before:attach then a,b,c attach               identical
```

Order, event pairing, and child receiver all match.

**One thing I checked and want to record as a non-finding**, since it looks like a difference on inspection: `eachChild` reads `children.length` twice — once as `candidateLength` for the array-like test, then again for the loop bound. I expected that to diverge from Underscore's single read against a `length` getter with side effects. It doesn't: `_.isArrayLike` reads it and `_.each` reads it again, so master does the same. Instrumenting a getter gives **4 reads on both** across the two propagation passes. No behavioral difference, and no reason to change it.

The `MAX_ARRAY_INDEX` bound matches `_.isArrayLike`'s, including the `NaN` case falling through to the object path since `NaN >= 0` is false.

Worth noting on the running arithmetic: this slice is +174 B for one import, where #271 was +413 B. Fourteen imports cleared has now cost roughly 4.1 kB against a ~6.9 kB peer, with 22 modules still to go. The per-import cost varies a lot depending on how much `_.each` polymorphism has to be reproduced, so the total is still genuinely open — which is the case for establishing the dependency-inclusive scenario now rather than at the end, while a partial state is still measurable.

PR #273, opened 2026-08-29
refactor(utils): replace Underscore once wrapper

paulfalgout:
## Linked issue

- Related #241

## What

- Replace `onceWrap`'s Underscore `once` dependency with a narrow owned wrapper.
- Preserve cleanup ordering, reentrancy, error, receiver, argument, memoization, and callback-identity behavior.
- Add focused regression tests and regenerate the main distribution artifacts.

- Makes the one-shot callback lifecycle explicit in the owning utility so agents can reason about unbinding and retry behavior without reconstructing Underscore's `once`/`before` layers.

## Public behavior contract

- Canonical behavior after this PR: the wrapper consumes itself before cleanup, invokes cleanup before the user callback, forwards dynamic receiver and arguments, memoizes successful results, and exposes the original callback through `_callback`.
- Superseded behavior removed, if any: the runtime dependency on Underscore `once` in this utility.
- Compatibility exception and removal condition, if any: none; this is an exact behavior-preserving implementation replacement.

## Runtime cost

- Startup/hot path: removes Underscore's nested `once`/`before` call layer from each registration.
- Allocations/retention: removes one per-registration closure layer and adds no new retained collection.

## Validation

Commands:

```sh
npx vitest run test/unit/utils/once-wrap.spec.js test/unit/mixins/events.spec.js test/unit/mixins/requests.spec.js test/unit/radio.spec.js --coverage.enabled=false
npm test
npm run test:source
npm run build
npm run test:dist
npm run lint:ci
git diff --check
```

Results: 68 focused/integration tests; 75 files and 1,419 full-suite tests; source and distribution validation, lint, build, and diff check passed. Independent exact-head differential and integration review found zero defects.

## Package and browser impact

- Entry points or install fixtures affected: main Marionette bundles are regenerated; no entrypoint contract changes.
- Browser contracts affected: none; one-shot callback behavior is unchanged.
- Production/dev/test module-graph impact: removes one production Underscore importer without adding an external dependency.

## Rollback or deprecation

- Revert this commit to restore the Underscore wrapper; no data migration or deprecation sequence is involved.

## Scope

Impacts only `onceWrap`, its focused tests, and generated main distribution artifacts. It does not change event names, request/reply routing, Radio APIs, or lifecycle contracts.

## Deployment Impact

## Review notes

- Review cleanup-before-callback ordering, consumed-before-reentrancy/throws, later-call return values, dynamic `this`/arguments, `_callback` identity, and the exact wrapper passed to cleanup.

---
## Summary by cubic
Replaces Underscore's `once` wrapper in `onceWrap` with an owned implementation so one-shot callback behavior no longer depends on that import.

- Preserves the existing contract: cleanup runs before the callback, results are memoized, receiver and arguments are forwarded, and `_callback` keeps the original callback.
- Removes the production `underscore` import from `utils/once-wrap.js`.
- Adds focused regression tests for ordering, reentrancy, memoization, and identity.
- Regenerates the main distribution artifacts.

<sup>Written for commit bf6bfc964c3081569286e0eb825dfbe52755568f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/273?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `bf6bfc96`

**Approved.** CI green, +77 B, 18,028 remaining. Production Underscore importers **22 → 21**.

Six behaviors exercised through the public `once` surface, all identical:

```
                                      master   p273
fires exactly once                    n=1      n=1
receiver + argument forwarding        ctx, 1|b ctx, 1|b
off(originalFn) removes the wrapper   yes      yes      <- _callback identity
reentrant re-trigger inside callback  n=1      n=1
handler still bound during callback   false    false    <- cleanup precedes callback
callback throws, then re-trigger      calls=1  calls=1
```

**The throw case was the one I expected to diverge**, and it doesn't. `_.once` is `_.before(2, fn)`, which decrements its counter *before* invoking, so a first call that throws leaves the wrapper consumed — a later trigger returns the (unset) memo rather than retrying. The phrase "memoizes successful results" in your description reads like it might have changed that; it hasn't. Both branches call the throwing callback exactly once and never retry. Worth knowing the parity is deliberate rather than accidental, because "retry after a failed one-shot" is a plausible thing for someone to assume later and quietly add.

The cleanup-before-callback ordering is confirmed observationally too — inspecting the registry from inside the callback shows the handler already unbound on both, which is what makes the reentrancy result fall out rather than being separately enforced.

PR #274, opened 2026-08-29
refactor(error): replace Underscore property copying

paulfalgout:
## Linked issue

- Related #241

## What

- Replace `MarionetteError`'s Underscore `pick`/`extend` property pipeline with a narrow owned fixed-key pipeline.
- Preserve allowed-property access order, inheritance, descriptors, precedence, stack handling, URL/string behavior, prototype identity, and public exports.
- Add focused regression coverage, including exclusion of unrelated enumerable `Object.prototype` pollution, and regenerate main distribution artifacts.

- Makes the public error-construction contract explicit in its owning module so agents can see the whitelist, precedence, access order, and stack paths without reconstructing two polymorphic Underscore helpers.

## Public behavior contract

- Canonical behavior after this PR: only the fixed `errorProps` allow-list is materialized; native Error values are applied first and options second, including inherited, non-enumerable, and present-`undefined` allowed properties.
- Superseded behavior removed, if any: the runtime dependency on Underscore `pick`/`extend`, plus accidental copying of unrelated enumerable `Object.prototype` pollution through Underscore's temporary objects.
- Compatibility exception and removal condition, if any: none. The pollution behavior contradicted the existing allow-list and invalid-property contract and is intentionally removed.

## Runtime cost

- Startup/hot path: an isolated 500,000-construction differential with capture disabled measured about 8.8% faster than the Underscore pipeline; native Error/stack work still dominates real construction.
- Allocations/retention: removes Underscore rest/flatten/allKeys intermediates while retaining two staged snapshots required for exact access/mutation ordering.

## Validation

Commands:

```sh
npx vitest run test/unit/utils/error.spec.js --coverage.enabled=false
npx vitest run --coverage.enabled=false --reporter=dot
npx eslint utils/error.js test/unit/utils/error.spec.js
npm run test:source
npm run build
npm run check:dist
npm run test:dist
npm run test:performance
git diff --check
```

Results: 19 focused error tests; 75 files and 1,428 full-suite tests; ESLint 10.9.1, build, source/dist validation, 92 performance-contract tests, deterministic rebuild, and diff check passed. Independent exact-head differential review found zero defects.

## Package and browser impact

- Entry points or install fixtures affected: the public `MarionetteError` export and main bundles are exercised; export identity is unchanged.
- Browser contracts affected: both callable and non-callable `Error.captureStackTrace` paths remain supported, using the existing negative guard and fallback stack.
- Production/dev/test module-graph impact: removes one production Underscore importer without adding an external dependency.

## Rollback or deprecation

- Revert this commit to restore the Underscore property pipeline; no data migration or deprecation sequence is involved.

## Scope

Impacts only `MarionetteError`, its focused tests, and generated main distribution artifacts. It does not change other error call sites, diagnostics, lifecycle behavior, or external error property names.

## Deployment Impact

## Review notes

- Review fixed-key get-before-has order, native/options snapshot precedence, inherited/non-enumerable/present-undefined allowed values, ordinary output descriptors, captureStackTrace native/fallback descriptors, constructor/prototype/export identity, and the intentional rejection of unrelated Object.prototype pollution.

---
## Summary by cubic
Replaces `MarionetteError`'s Underscore `pick`/`extend` property pipeline with a fixed-key owned-property pipeline (related to #241). The old path could copy unrelated enumerable `Object.prototype` pollution through Underscore's temporary objects; the new path only materializes the allow-listed `errorProps`.

- Preserves allowed-property precedence, access order, inheritance, descriptors, stack handling, URL/string behavior, prototype identity, and public exports.
- Removes one production Underscore dependency and regenerates dist artifacts.

<sup>Written for commit e3c1450b22a3f9703cd099b70efe12c07be1486c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/274?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `e3c1450b`

**Approved.** CI green, +266 B, 17,762 remaining. Production Underscore importers **21 → 20**.

The pollution fix is real, and I verified it by actually polluting `Object.prototype` rather than reasoning about it:

```
Object.defineProperty(Object.prototype, 'POLLUTED', { value:'leak', enumerable:true })

master   own 'POLLUTED' on the error:  COPIED
p274                                    excluded
```

Everything else is parity:

```
                                         master        p274
non-allow-listed option                  excluded      excluded
name / code / message / instanceof Error identical     identical
stack present                            yes           yes
prototype identity                       preserved     preserved
inherited allowed property (message)     "from-proto"  "from-proto"
message explicitly set to undefined      undefined     undefined
```

**The inherited-allowed-property case is the one worth calling out as preserved.** Passing `Object.create({message:'from-proto'})` still materializes `message`, which is the opposite of the own-only narrowing applied everywhere else in this series — and correctly so, because the allow-list is what bounds the surface here, not ownership. It would have been easy to reach for `Object.hasOwn` reflexively by now and quietly break error construction for anyone building options objects by delegation.

On the pollution finding's practical weight: it requires `Object.prototype` to carry an enumerable property in the first place, so this is defense-in-depth rather than a bug users are hitting. What it does buy is that a prototype-pollution payload can't ride outward through a thrown Marionette error into whatever consumes it — a reasonable property for a diagnostics object to have, and the sort of thing that's much cheaper to fix now than after it's load-bearing.

PR #279, opened 2026-08-29
refactor(radio): replace underscore result lookups

paulfalgout:
Related #241

## Linked issue

- Related #241

## What

- Replace the Radio mixin's three Underscore `result` calls with Marionette's owned `getValue` helper.
- Preserve the exact channel, event, request, binding, and destroy-listener initialization order.
- Add focused receiver, zero-argument, falsy short-circuit, ordering, and error-cutoff coverage and regenerate affected main distribution artifacts.

## Why

The shared literal-key helper now owns the configuration lookup contract. Using it here removes another complete production Underscore importer without duplicating lookup semantics.

- Makes Radio option resolution follow the same explicit, tested convention as entity-event configuration.
- Locks the initialization sequence and short-circuit boundary so agents can change Radio behavior without inferring order from Underscore calls.

## Public behavior contract

- Canonical behavior after this PR: resolve `channelName`; stop on any falsy value; otherwise create the channel, resolve and bind `radioEvents`, resolve and bind `radioRequests`, then register destruction cleanup.
- Superseded behavior removed, if any: direct use of Underscore `result` in the Radio mixin.
- Compatibility exception and removal condition, if any: none; the owned helper preserves the relevant literal-key receiver, argument, inheritance, getter, proxy, and error behavior.

## Runtime cost

- Bundle: the size/resource contract passes; the direct production Underscore importer count drops from 15 to 14.
- Startup/hot path: the existing three configuration resolutions now call the already-shared small helper; no additional initialization step is added.
- Allocations/retention: no new per-instance state or retained objects are added.

## Scope

- Affects only Radio mixin initialization for `channelName`, `radioEvents`, and `radioRequests`.
- Does not change Radio channel implementation, event/request normalization, destruction behavior, or other Underscore consumers.

## Deployment Impact

## Validation

Commands:

```sh
npm ci --no-audit --no-fund
npm exec vitest -- run test/unit/mixins/radio.spec.js test/unit/utils/get-value.spec.js --reporter=dot
npm run lint:ci
npm run build
npm test -- --reporter=dot
npm run test:source
npm run test:dist
npm run test:fixtures
npm run size
npm run test:performance
npm run performance:timing
git diff --check
```

Results: focused tests passed 19/19; full tests passed 1,447/1,447 across 76 files; source, distribution, packed fixture, lint, build, size/resource, 92 performance-contract, hosted timing, and diff validation passed. Independent exact-head review passed 35/35 focused assertions with no findings.

## Package and browser impact

- Entry points or install fixtures affected: main ESM, CJS, UMD, and minified artifacts are regenerated; packed ESM/CJS/type fixtures pass.
- Browser contracts affected: none; no new platform API is introduced.
- Production/dev/test module-graph impact: the Radio mixin stops importing Underscore; other production consumers still retain it.

## Rollback or deprecation

## Review notes

- Tests assert the full normal-path sequence and that falsy `channelName` values prevent every later lookup and binding.
- Generated ESM, CJS, UMD, and minified formats share one `getValue` helper rather than duplicating it.

---
## Summary by cubic
Replaces the Radio mixin's three Underscore `result` calls with Marionette's owned `getValue` helper, so Radio option resolution follows the same contract as entity-event configuration. This drops the mixin's direct production Underscore import, related to #241.

**Changes**
- Preserves the existing order: resolve `channelName`, create the channel, resolve and bind `radioEvents`, resolve and bind `radioRequests`, then attach the destroy handler.
- Stops all later lookups and bindings when `channelName` is falsy.
- Adds tests covering receiver binding, zero-argument functions, falsy short-circuit, ordering, and error cutoffs.
- Regenerates `dist` artifacts (ESM, CJS, UMD, minified) with the new helper.
- No behavior change or migration required.

<sup>Written for commit fcbd5c9495f526c946070d9a7c87f516440edbaf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/279?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `fcbd5c94`

**Approved.** CI green, **−14 B**, 17,341 remaining. Production Underscore importers **16 → 15**.

Three pure substitutions with identical arguments and unchanged ordering — the falsy `channelName` early return still precedes channel creation, `bindEvents` still precedes `bindRequests`, and the `destroy` listener is still attached last. Nothing else in the file moved.

**Static review is genuinely sufficient here**, which I don't say lightly given `node_modules` is still empty. The substitution can only change behavior if `getValue` differs from `_.result`, and I checked that correspondence line by line on #278. All three call sites use the two-argument form, so they exercise only the simplest path — no fallback, no function-fallback invocation, just lookup-then-call-if-callable.

**The −14 B is the interesting number.** #278 cost +273 B to build the helper and its tests; this first migration gives 14 back while removing an importer. That's the shape the sequencing predicted: front-load the contract, then each of the remaining ~17 `result(` sites is a cheap or negative substitution rather than a fresh opportunity to re-derive `undefined`-versus-`null`. On that basis I'd expect the importer count to fall faster than the byte total rises for the next stretch.

The added coverage matches the risk correctly too — receiver, zero arguments, falsy short-circuit, ordering, and error cutoff are exactly the properties a reordering or an eager-evaluation mistake would break, and they're the ones the substitution itself can't guarantee.

PR #280, opened 2026-08-29
refactor(common): replace underscore mixin helpers

paulfalgout:
Related #241

## Linked issue

- Related #241

## What

- Resolve default instance options through Marionette's owned `getValue` helper.
- Compose the fixed Events and Requests mixins with the existing own-property assignment primitive.
- Add focused option-order, method identity/descriptor, key-order, builtin-key, prototype, and inherited-pollution coverage and regenerate affected main distribution artifacts.

## Why

These two owned primitives express the exact contracts needed by the Common mixin and remove its complete Underscore import without duplicating behavior.

- Gives agents one explicit path for callable configuration values and one explicit path for fixed-source mixin composition.
- Locks down merge order, receiver behavior, method identity, descriptors, and the intentional inherited-pollution boundary instead of relying on broad Underscore helper semantics.

## Public behavior contract

- Canonical behavior after this PR: resolve default options once with the instance receiver and zero arguments, copy own defaults before own passed options, assign `this.options`, then call `mergeOptions` with the original arguments.
- Superseded behavior removed, if any: direct Underscore `result` and `extend` use in the Common mixin.
- Compatibility exception and removal condition, if any: inherited enumerable pollution is intentionally no longer promoted into the Common mixin API; the fixed Events and Requests sources use their own enumerable methods only.

## Runtime cost

- Bundle: the size/resource contract passes; the direct production Underscore importer count drops from 14 to 13.
- Startup/hot path: fixed mixin composition remains a one-time module evaluation; option initialization keeps the same lookup and two-source assignment shape.
- Allocations/retention: no new retained runtime object is added; each options initialization still creates one fresh options object.

## Scope

- Affects Common mixin construction and `_setOptions` only.
- Does not change event/request implementations, option selection precedence, class-option merging, or other Underscore consumers.

## Deployment Impact

## Validation

Commands:

```sh
npm ci --no-audit --no-fund
npm exec vitest -- run test/unit/mixins/common.spec.js test/unit/utils/get-value.spec.js test/unit/utils/assign-in.spec.js --reporter=dot
npm run lint:ci
npm run build
npm test -- --reporter=dot
npm run test:source
npm run test:dist
npm run test:fixtures
npm run size
npm run test:performance
npm run performance:timing
git diff --check
```

Results: focused tests passed 19/19; full tests passed 1,451/1,451 across 76 files; source, distribution, packed fixture, lint, build, size/resource, 92 performance-contract, hosted timing, and diff validation passed. Independent exact-head review passed 39/39 focused assertions with no findings.

## Package and browser impact

- Entry points or install fixtures affected: main ESM, CJS, UMD, and minified artifacts are regenerated; packed ESM/CJS/type fixtures pass.
- Browser contracts affected: none; no new platform API is introduced.
- Production/dev/test module-graph impact: the Common mixin stops importing Underscore; other production consumers still retain it.

## Rollback or deprecation

## Review notes

- Fixed-source composition preserves method order, identity, assignment descriptors, Common mixin identity, and its ordinary object prototype.
- Excluding inherited/Object.prototype pollution and safely treating a hypothetical own `__proto__` as data are intentional owned-boundary improvements over Underscore `extend`.

---
## Summary by cubic
Replaces Underscore's `result` and `extend` helpers in the Common mixin with Marionette's owned `getValue` and `assignOwn` primitives, removing the direct Underscore import from this mixin and locking down option resolution and mixin composition behavior. Related to #241.

**Behavior changes**
- `_setOptions` now resolves default options via `getValue`, using the instance receiver and zero arguments.
- Common mixin composition now uses `assignOwn`, preserving method order and identity and no longer promoting inherited enumerable properties.
- Added focused tests covering option order, method identity, descriptors, key order, builtin keys, prototype, and inherited pollution.
- Regenerated `dist/` artifacts; no change to event/request behavior or option selection precedence.

<sup>Written for commit 38e9e125ceaf9018c2e88780c825fd8fb3eb62c6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/280?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `38e9e125`

**Approved.** CI green, +5 B, 17,336 remaining. Production Underscore importers **15 → 14**.

Two substitutions, both onto primitives whose semantics are now pinned:

`result(this, 'options')` → `getValue(...)` is the mechanical one, equivalent by the correspondence I checked line by line on #278.

`extend(CommonMixin, EventsMixin, RequestsMixin)` → `assignOwn(...)` is the substantive one, and reading `assign()` confirms the delta is exactly the intended narrowing. It iterates `for (const key in source)` — own plus inherited enumerables, same reach as `_.extend` — then filters with `Object.hasOwn` when `ownOnly` is set. So key order is preserved (insertion order for string keys on plain literals), accessors still flatten to data properties because both read `source[key]`, and a mixin key named `toString` or `__proto__` still lands, the latter through `setProperty`'s `defineProperty` path. The only behavior removed is copying inherited enumerables from the two mixin literals — which, since both are module-level object literals, means `Object.prototype` pollution and nothing else.

Keeping `assignIn` alongside `assignOwn` is the right shape too: the `_.extend` equivalent still exists for callers that need it, so this is a deliberate choice at one site rather than a global narrowing applied by fiat.

**The pollution test being `async` is the detail I'd highlight.** `assignOwn(CommonMixin, …)` runs at module load, so its result is already composed by the time any ordinary test executes — you cannot observe it after the fact. Re-importing the module with pollution installed is the only way to exercise that path, and it's easy to write a test that appears to cover this and actually asserts nothing. Same category as the `assert.throws` probe on #231 and the read-count getter on #278.

Still source-only review; `node_modules` remains empty. Adequate here, since both substitutions are onto primitives I've now read in full and the semantic delta is characterizable without execution.

PR #284, opened 2026-08-30
refactor(dom): replace underscore attribute helpers

paulfalgout:
Related #241

## Linked issue

- Related #241

## What

- Replace `config/dom.js`'s complete Underscore import with a captured native key reader and indexed own-key traversal.
- Preserve the exact `setAttributes` branch and getter/trap order while continuing to use the existing safe property writer.
- Add focused attribute-map contract coverage and regenerate affected distribution artifacts.

## Why

`setAttributes` only needs a narrow own-enumerable string-key snapshot, not Underscore's general collection API. Owning that loop directly removes another production dependency edge while keeping DOM writes predictable and inexpensive.

- Makes the accepted attribute input domain, snapshot timing, branch order, mutation behavior, and error cutoff explicit in local code and tests.
- Prevents agents from substituting a superficially equivalent iterator that reads attribute values before the element property-existence check.
- Locks module-load key-reader capture and an indexed loop so later global or iterator mutation cannot silently change the contract.

## Public behavior contract

- Canonical behavior after this PR: `setAttributes` snapshots own enumerable string keys in property order for object/function inputs, ignores nullish and primitive inputs, checks `attr in el` before reading each attribute value, writes existing properties safely, and otherwise calls `setAttribute`.
- Arrays, boxed strings, functions, built-in names, and an own `__proto__` retain their documented JavaScript own-key behavior; inherited, symbol, and non-enumerable keys are excluded.
- Superseded behavior removed: direct Underscore `keys` and `each` use in the default DOM adapter.
- Compatibility exception and removal condition: none; the owned implementation preserves the prior observable contract.

## Runtime cost

- Bundle: the size/resource contract passes; direct production Underscore importers drop from 10 to 9. Brotli deltas versus the exact base are ESM +51 B, CJS +42 B, UMD +38 B, and minified +32 B.
- Startup/hot path: one `Object.keys` reference is captured at module load; attribute application remains one indexed linear pass.
- Allocations/retention: one key snapshot per `setAttributes` call, matching the prior Underscore path without an additional iterator allocation.

## Scope

- Affects only the default DOM adapter's `setAttributes` implementation and its generated main artifacts.
- Does not change custom DOM adapters, `setDomApi`, attribute selection, property assignment semantics, other DOM operations, or remaining Underscore consumers.

## Deployment Impact

## Validation

Commands:

```sh
npm ci --no-audit --no-fund
npm exec vitest -- run test/unit/config/dom.spec.js test/unit/utils/assign-in.spec.js test/unit/mixins/view.spec.js --reporter=dot
npm run lint:ci
npm run build
npm test -- --reporter=dot
npm run test:source
npm run test:dist
npm run test:fixtures
npm run size
npm run test:performance
npm run performance:timing
git diff --check
```

Results: focused tests passed 83/83; full tests passed 1,496/1,496; source, distribution, packed CJS/ESM/types/Vite fixtures, lint, build, size/resource, 92 performance-contract, hosted timing, and diff validation passed. Independent exact-head review passed 49/49 focused integration assertions with no findings.

## Package and browser impact

- Entry points or install fixtures affected: main ESM, CJS, UMD, and minified artifacts are regenerated; packed CJS/ESM/type/Vite fixtures pass.
- Browser contracts affected: no intended behavior change; DOM property and attribute calls preserve their prior ordering and error behavior.
- Production/dev/test module-graph impact: the default DOM adapter stops importing Underscore; other production consumers still retain it.

## Rollback or deprecation

## Review notes

- This intentionally does not reuse `eachOwn`: that helper reads the value before invoking its callback, while `setAttributes` must probe `attr in el` before reading `attrs[attr]`.
- The key reader is captured at module load and traversal uses a captured-length indexed loop to preserve the prior Underscore timing and avoid an Array iterator dependency.

---
## Summary by cubic
Replaces Underscore's `keys` and `each` in `setAttributes` with a native `Object.keys` snapshot and an indexed loop, so the default DOM adapter no longer imports `underscore` and attribute application stays one linear pass. Object and function inputs keep the same contract; nullish and primitive inputs now no-op instead of throwing or parsing strings.

- Captures `Object.keys` at module load and uses a captured-length loop, preserving the prior snapshot timing without an iterator allocation.
- Drops the `underscore` import from the default DOM adapter; no other module's behavior changes.
- Regenerates the `dist` artifacts with Brotli deltas of +32 to +51 B and adds ~270 lines of focused tests covering key order, `attr in el` probing before value reads, and edge cases like `__proto__` and boxed primitives.
- Relates to #241.

<sup>Written for commit 07b7b10865d72d81a26d99a32901a778f4b32349. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/284?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of attribute updates for null, undefined, primitive, callable, boxed-string, and sparse-array inputs.
  * Preserved predictable property-reading and update ordering, including proper error propagation.

* **Tests**
  * Expanded coverage for attribute updates, built-in properties, special property names, getters, setters, and failure scenarios.


paulfalgout:
## Code-review approval — exact head `07b7b108`

**Approved.** CI green, +163 B, 16,962 remaining. Production Underscore importers **11 → 10**.

**The added guard is the part I'd credit specifically, because it's exactly what a straight translation gets wrong:**

```js
const attrsType = typeof attrs;
if (attrs == null || attrsType !== 'object' && attrsType !== 'function') { return; }
```

`_.keys` and `Object.keys` disagree on primitives — `_.keys('abc')` returns `[]` because it checks `isObject` first, while `Object.keys('abc')` returns `['0','1','2']`. Dropping straight to `objectKeys(attrs)` would have turned `setAttributes(el, 'abc')` from a silent no-op into three bogus attribute writes. The guard restores `_.keys`' behavior precisely, and it admits functions, which `_.isObject` also does.

Everything else lines up: keys are snapshotted before the loop in both, iteration order is unchanged, the `attr in el` branch still routes through `setProperty` so #264's element-corruption fix stays intact, and capturing `Object.keys` at module load matches the `isString` tag-reader pattern from #277.

**The test list is the most thorough in this series**, and notably it pins ordering and failure sequencing rather than just outcomes — "propagates membership errors *without reading the attribute value*", "propagates attribute getter errors *before writing the property*", "propagates property write errors *after reading*". Those are exactly the observations that distinguish a faithful loop from one that reorders reads and writes, and they're invisible to a test that only checks the final element state. `retains Object.keys behavior for boxed strings and sparse arrays` covers the divergence the guard exists for, from the other direction.

Still source-only review; `node_modules` remains empty. Adequate here — the change is a bounded loop rewrite whose one risky edge is explicitly handled and covered.

PR #288, opened 2026-08-30
refactor(radio): own radio composition

paulfalgout:
Related #241

## What
- Remove the complete `modules/radio.js` Underscore import.
- Compose Radio and Channel with Marionette-owned assignment helpers, and synthesize forwarded APIs with captured own-key snapshots and indexed loops.
- Define reset-all around the v5 own-enumerable channel-map contract and add focused composition, forwarding, mutation, proxy, and error-order coverage.

## Why
Radio was one of six remaining Marionette runtime modules importing Underscore. Its fixed APIs and internal channel registry are explicit string-key maps, so Marionette can own their iteration and assignment semantics without retaining Underscore callback or array-like branches.

Agents no longer need to infer whether `Radio._channels` is an array-like collection or an own-key registry, whether inherited methods become public Radio APIs, or when reset-all snapshots and reads registry entries. The implementation and focused tests now state those contracts directly.

## Public behavior contract
Radio and Channel composition uses own enumerable string keys with prototype-safe assignment. Forwarded top-level methods preserve method inventory order, calling receiver, arguments, return values, and lazy channel-method lookup. `Radio.reset()` snapshots own enumerable channel names in standard key order, reads each channel lazily, ignores later additions, observes deletions, and stops at the first error. A numeric `length` key is an ordinary channel-map entry; inherited keys and Underscore’s malformed array-like interpretation are not part of the v5 contract.

## Runtime cost
No new runtime module is introduced. Captured `Object.keys` plus indexed loops remove Underscore callback dispatch while retaining one key snapshot where ordering requires it. Exact-base Brotli-11 deltas are +55 B ESM, +50 B CJS, +53 B UMD, and +47 B minified; every artifact remains below the per-path approval threshold.

## Validation
- `npm ci --no-audit --no-fund`
- `npm exec vitest -- run test/unit/radio-composition.spec.js test/unit/radio.spec.js test/unit/mixins/radio.spec.js test/unit/requests.spec.js --reporter=dot` — 83 tests passed
- `npm test -- --reporter=dot` — 83 files, 1,548 tests passed
- `npm run lint:ci`
- `npm run build`
- `npm run check:dist`
- `git diff --check`
- Independent exact-head review approved `9fce05638d546a71dd50cfe887e04f0ccf92b2a7` with zero findings.

## Scope
Impacts Radio static composition, Channel prototype composition, top-level method forwarding, reset-all iteration, focused unit coverage, and committed distributions. It does not change Events or Requests semantics, named-channel reset validation, the remaining Underscore importers, package manifests, release configuration, or GitHub settings.

## Package/browser impact
The public package exports and browser support profile are unchanged. Main runtime output remains externally dependent on Underscore until the remaining five importers are migrated. No manifest or lockfile change is required for this intermediate slice.

## Deployment Impact
No application or website deployment is required. No package, tag, website, or release is published by this pull request.

## Rollback/deprecation
Rollback is a normal revert of this commit. No compatibility alias or dual path is added; the own-key channel-map contract is canonical for v5.

## Review notes
The deliberate semantic narrowing is reset-all behavior when `_channels` is externally replaced by a malformed numeric-length object. Review own-map snapshot, lazy-read, mutation, and error-cutoff behavior rather than Underscore array-like parity.

---
## Summary by cubic
Removes the complete Underscore import from `modules/radio.js` by composing Radio and Channel with Marionette-owned assignment helpers, and defines `reset()` around an own-enumerable channel-map contract. Relates to #241.

**Behavior changes**
- `Radio` and `Channel.prototype` are now assembled with `assignOwn`; forwarded top-level methods preserve order, receiver, arguments, and return values.
- `Radio.reset()` snapshots own enumerable channel names, reads each channel lazily, ignores channels added later, observes deletions, and stops at the first error.
- A numeric `length` key is an ordinary channel-map entry; inherited keys and Underscore’s array-like interpretation are no longer part of the v5 contract.

**Validation**
- Adds `test/unit/radio-composition.spec.js` with 281 lines covering composition, forwarding, mutation, proxy, and error-order behavior.
- Runs green: 1,548 tests, lint, build, and dist check.
- Bundle growth is +55 B ESM, +50 B CJS, +53 B UMD, and +47 B minified — all below the per-path approval threshold.
- Public exports and browser support are unchanged; main runtime output still depends on Underscore until the remaining five importers are migrated.

<sup>Written for commit 9fce05638d546a71dd50cfe887e04f0ccf92b2a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/288?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
## Code-review approval — exact head `9fce0563`

**Approved.** CI green, +205 B, 16,165 remaining. Production Underscore importers **6 → 5**.

**The thing I most wanted to check survives intact: `reset()` still dispatches on `arguments.length`, not on the truthiness of `channelName`.** That's the fix from #211, where `!channelName` meant `Radio.reset('')` silently reset *every* channel. A rewrite of this function is exactly where that could regress, and the arity guard is untouched.

The rest is faithful. The nested `each`/`keys` forwarding loop becomes explicit indexed loops with `objectKeys` captured at module load, matching the pattern from #277 and #284. `setProperty` on the forwarded assignments means a method name of `__proto__` on `Events`/`Requests` would store as data rather than mutate `Radio`'s prototype — theoretical here, but consistent. `extend` → `assignOwn` on both `Radio` and `Channel.prototype` is the same own-only narrowing as #280/#285/#287.

The reset-all loop reads `this._channels[name]` inside the iteration rather than capturing values up front — which is what `_.each` does too, so a channel deleted mid-iteration yields `undefined` and throws on `.reset()` in both versions. "Observes deletions… stops at the first error" describes that accurately.

**One calibration on the array-like point, including a correction to myself.** The contract says Underscore's "malformed array-like interpretation" is excluded, and that's right to state — but triggering it requires `_channels.length` to be a *number*, and a channel named `length` stores a Channel object there. So this is latent hardening rather than a live defect. I overstated the equivalent point on #283, where I described the `ui.js` heuristic as something that would silently bind nothing; that also needs a numeric `length` *value*, which is already an invalid `ui` entry. Both changes are worth making, but neither is fixing something reachable through valid configuration.

Still source-only review; `node_modules` remains empty.

PR #289, opened 2026-08-30
refactor(view): own view mixin helpers

paulfalgout:
Related #241

## What
- Remove the complete `mixins/view.js` Underscore import.
- Resolve View options with Marionette-owned `getValue` and validate selector strings with the shared `isString` helper.
- Compose ViewMixin from own enumerable mixin keys through `assignOwn`, with focused option-order, fallback, coercion, error, and composition tests.

## Why
ViewMixin was one of five remaining Marionette runtime modules importing Underscore. Existing owned helpers already define the v5 lookup, string-classification, and composition contracts, so this slice removes another dependency edge without adding runtime machinery.

Agents no longer need to infer `_.result` behavior independently across `el`, element attributes, child-event maps, and event prefixes. Focused tests now state receiver binding, zero arguments, read counts, evaluation/error order, undefined-only fallback, coercion, and fixed mixin ownership explicitly.

## Public behavior contract
`el`, `tagName`, `attributes`, `id`, `className`, `childViewEvents`, `childViewTriggers`, and `childViewEventPrefix` retain inherited lookup, owning receiver, zero-argument function resolution, and error propagation. Existing `id` and `className` truthiness guards still perform a second read before resolution. Prefix fallback occurs only when the property itself is undefined; a resolver returning undefined still produces `undefined:`. String validation uses the shared String-tag contract. Fixed ViewMixin composition includes own enumerable string keys only and safely owns built-in names including `__proto__`.

## Runtime cost
No new helper or allocation path is introduced; all replacements reuse helpers already present in the main bundle. Exact-base Brotli-11 deltas are 0 B ESM, -1 B CJS, +2 B UMD, and -9 B minified, for a small aggregate reduction.

## Validation
- `npm ci --no-audit --no-fund`
- `npm exec vitest -- run test/unit/mixins/view-composition.spec.js test/unit/mixins/view.spec.js test/unit/view-el.spec.js test/unit/utils/is-string.spec.js test/unit/utils/get-value.spec.js --reporter=dot` — 66 tests passed
- `npm test -- --reporter=dot` — 84 files, 1,559 tests passed
- `npm run lint:ci`
- `npm run build`
- `npm run check:dist`
- `git diff --check`
- Independent exact-head review approved `4f21505f9255503f0e8723687e3abcc07cd4320f` with zero findings.

## Scope
Impacts ViewMixin option resolution, element validation, child-event proxy setup, fixed mixin composition, focused unit coverage, and committed distributions. It does not change View public exports, behavior parsing, Events internals, the remaining Underscore importers, package manifests, release configuration, or GitHub settings.

## Package/browser impact
The package export surface and supported browser profile are unchanged. Main runtime output remains externally dependent on Underscore until the remaining four importers are migrated. No manifest or lockfile change is required for this intermediate slice.

## Deployment Impact
No application or website deployment is required. No package, tag, website, or release is published by this pull request.

## Rollback/deprecation
Rollback is a normal revert of this commit. No compatibility alias or dual path is added; owned helpers and own-key composition are canonical for v5.

## Review notes
Pay particular attention to the intentionally preserved `id`/`className` double-read guards and the distinction between a missing prefix property and a prefix resolver that returns undefined.

---
## Summary by cubic
Removes ViewMixin's Underscore import by replacing `result` and `isString` with the Marionette-owned `getValue` and `isString` helpers, and composes ViewMixin from its own enumerable keys via `assignOwn`. Advances #241 by removing one of the five remaining Underscore dependency edges. View option lookups (`el`, `tagName`, `attributes`, `id`, `className`, `childViewEvents`, `childViewTriggers`, `childViewEventPrefix`) keep their existing behavior, including error propagation.

**Refactors**
- ViewMixin option resolution and element string validation now use `getValue` and `isString`.
- ViewMixin composition uses only own enumerable string keys and safely handles built-in names like `__proto__`.
- Added focused tests for option resolution, selector validation, and composition ownership.

**Review notes**
- The `id` and `className` truthiness guards still perform a second read before resolution.
- Prefix fallback happens only when the property itself is undefined; a resolver returning `undefined` still yields `undefined:`.
- Bundle size is effectively unchanged, and no deployment or migration is required.

<sup>Written for commit 4f21505f9255503f0e8723687e3abcc07cd4320f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/289?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved view configuration handling, including element resolution, child-view events, event prefixes, and mixin composition.
  * Added safer validation and error handling for string-based element tags and event settings.

* **Tests**
  * Expanded coverage for view option resolution, event proxy behavior, validation, and mixin composition edge cases.


paulfalgout:
## Code-review approval — exact head `4f21505f`

**Approved.** CI green, **−8 B**, 16,173 remaining. Production Underscore importers **5 → 4**.

Straight substitutions onto primitives already pinned: eight `result` → `getValue` calls, `isString` at the selector guard (`view.js:47`), and `extend` → `assignOwn` for the six-mixin composition.

**Both quirks your contract calls out are preserved exactly, and I'd rather see them stated than quietly cleaned up.**

The `id`/`className` double read is intact — `if (this.id) { attrs.id = getValue(this, 'id'); }` reads the property for truthiness and again for resolution, so a getter still fires twice. Anyone tidying this into a single read would change observable behavior for accessor-backed properties, and it would look like an obvious simplification.

The `'undefined:'` case is the sharper one. `getValue`'s fallback substitutes only when the *property* is `undefined`, so a `childViewEventPrefix` **function** that returns `undefined` yields `prefix = undefined`, and `prefix + ':'` produces the literal string `'undefined:'`. That's what `_.result` did, and the ordering in `getValue` — fallback check before the callable check — is what preserves it. A reimplementation that applied the fallback to the *resolved* value instead would silently start returning `'childview:'` there, which is arguably nicer and definitely different.

Composition and lookup semantics carry over unchanged: inherited property lookup, owning receiver, zero-argument invocation, and error propagation all come from `getValue`'s correspondence to `_.result` verified on #278.

Still source-only review; `node_modules` remains empty.

PR #306, opened 2026-08-30
fix(view): make getRegion a pure query

paulfalgout:
Related #132

## What

- Make `View#getRegion(name)` return an own registered Region without rendering or changing View state.
- Preserve render-before-lookup behavior for `showChildView`, `detachChildView`, and `getChildView`, including overridable `getRegion` implementations, without rerendering an already-rendered View.
- Update current documentation, the v4 migration ledger, tests, and generated distribution files for the pure-query contract.

## Why

The v5 lifecycle contract separates read-only queries from rendering mutations. `getRegion` previously rendered an unrendered View as a lookup side effect, which made optional inspection change lifecycle state and DOM. This slice removes that side effect while retaining deterministic behavior for child operations that depend on a resolved Region.

## Scope

This changes the shared `View` Region API and its generated browser/CommonJS distributions. Direct selector-backed usage such as `view.getRegion('content').show(child)` must render the parent first. Child operations continue to render before lookup and now do so even when a custom `getRegion` override does not delegate to the base method. Application Region lookup, website publication, package publication, tags, releases, and repository policy are intentionally unchanged.

## Deployment Impact

No deployment is required beyond consuming a future Marionette package containing this change. No website, npm package, tag, or release is published by this pull request.

## Testing

- `npm test -- --coverage.enabled=false` — 89 files, 1,623 tests passed.
- `npm run lint:ci -- --quiet`
- `npm run docs:check` — 40 pages built and 376 internal links validated.
- `npm run check:dist`
- `npm run test:source`
- `npm run test:fixtures`
- `npm run size` — cumulative 61.23 kB of 75.00 kB.
- `git diff --check`

Coverage includes unrendered and already-rendered Views, inherited and prototype-collision names, destroyed Views, selector-backed direct Region usage, non-delegating overrides, child-operation ordering, missing-name `MN0020`, and generated artifact reproducibility.

## Summary by CodeRabbit

* **New Features**
  * Region lookups now work without triggering a View render.
  * Child View operations render an unrendered View before accessing its Region.
  * Region queries return consistent results for missing, inherited, and destroyed Views.

* **Documentation**
  * Updated Region behavior, usage examples, and migration guidance.
  * Clarified when rendering is required and recommended child-view operations where appropriate.

* **Bug Fixes**
  * Prevented unintended rendering during direct Region queries while preserving expected child-view behavior.

---
## Summary by cubic
Makes `View#getRegion` a pure query that returns an own registered region without rendering the view (part of the v5 lifecycle contract in #132). Previously, calling `getRegion` on an unrendered view rendered it as a side effect, so optional inspection changed lifecycle state and DOM; now the view stays untouched until an explicit render or child operation.

**Behavior changes**
- Direct usage like `view.getRegion('content').show(child)` on an unrendered parent must render the parent first.
- Child operations `showChildView`, `detachChildView`, and `getChildView` still render before lookup and now dispatch through overridden `getRegion` implementations even when they don't delegate to the base method.
- Updates docs, v4 migration ledger, tests, and generated distribution files.

<sup>Written for commit 6c0d8411310215f246fc9f1269222eba8ac63bab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/306?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
Claude exact-head review for `c9f77322e13499190a26b09f54c842e8d3847f59`: **APPROVE — no blocking defects identified.**

Review context covered the pure `getRegion` contract, child-operation render-before-override dispatch, the explicit no-double-render tests, migration wording, and exact-head validation. Claude raised only non-blocking verification/documentation questions; the actual diff and local validation confirm `getChildView` is documented and tested as render-triggering, and `MN0020` remains the unknown-Region-name contract. No code change was required from this final review.

Auditable artifact: `/Users/paulfalgout/.ai-reviews/20260830T160145Z-review-with-claude.json`.

paulfalgout:
**Approved** at `c9f77322` (the SHA I fetched). Bundle size passes: +104 B aggregate (61.23 kB against the 75.00 kB ceiling), largest artifact delta +0.23%, so no growth approval needed.

## Consistent with the two slices before it

`getRegionForChild` renders, then dispatches through the overridable `view.getRegion(name)` — the same render-before-dispatch order `emptyRegions` adopted in #305, and it fixes the same latent inconsistency: a non-delegating `getRegion` override previously suppressed the render entirely. One helper across three call sites rather than the same two lines pasted three times.

Removing `should call render if getRegion is called without being rendered` from `region.spec.js` is correct — it asserted precisely the behavior being retired.

Parameterizing the child-operation suite over the three methods with a `populateOnRender` flag is the right call too; those blocks would otherwise be near-identical and would drift.

## The cross-PR seam test is the one I wanted

`dispatches the destroyed render no-op before reporting MN0020` pins exactly the coupling I flagged on #305: these operations call `view.render()` on a View that may be destroyed, and their safety depends on the destroyed-render guard from #303. Now that dependency has a test that fails if the guard is ever relaxed, instead of being an invariant three commits away with nothing connecting them.

Worth noting the same seam still has no test for `emptyRegions()`, which reaches `render()` the same way. This PR's test is the template for it.

The direct-lookup test is also good about *how* it fails: `region.show(child)` throwing MN0005 leaves the child unrendered, undestroyed, and unowned, so the new failure mode is clean rather than half-applied.

## Non-blocking: MN0005's remediation does not name the new cause

This PR turns `view.getRegion('content').show(child)` on an unrendered parent from working into throwing — the most user-visible change in the set. The migration ledger handles it well and says exactly the right thing: *render the parent explicitly before calling a selector Region's `show` directly*.

The runtime diagnostic does not. MN0005 still reads:

> Ensure the Region selector resolves within its parent element, or explicitly allow a missing element where supported.

Both suggestions are the wrong fix here. The selector is fine and `allowMissingEl` would convert a loud failure into a silent no-op, which is worse. Someone hitting this after upgrading gets pointed away from the actual cause, and MN0005's `docsAnchor` goes to the region-el page rather than anything about render ordering.

That matters more than usual because this PR substantially increases how often MN0005 fires — it is now the standard outcome of the migration path, not just a selector typo. Extending the remediation to mention rendering the parent first would cost one line in the catalog and close the gap between what the ledger tells a reader and what the runtime tells a caller. Given how much of #131/#132 has been about removing inference burden from diagnostics, this is the one place in the slice where a caller still has to infer.

Nothing blocking.

paulfalgout:
Addressed the valid MN0005 remediation feedback in exact head `6c0d8411310215f246fc9f1269222eba8ac63bab`: the diagnostic now tells callers to render the parent View before direct selector-backed Region operations, while retaining the existing selector-resolution and supported missing-element guidance.

Exact-head validation: 1,623 tests; lint; diagnostics catalog (28 entries); docs (40 pages, 376 links); reproducible dist; size 61.23/75.00 kB; diff check — all pass. Independent review: **APPROVED, no findings**. Claude exact-head review: **APPROVE, no blockers**; it raised only an optional wording-order preference, and no further change was warranted.

Claude artifact: `/Users/paulfalgout/.ai-reviews/20260830T161024Z-review-with-claude.json`.

paulfalgout:
**Approval stands** at `6c0d8411`. Catalog metadata only — runtime, tests, docs, and `dist/` are byte-identical to `c9f77322`, so sizes and the Bundle size pass carry over. `Build and validate` (which runs `check:diagnostics`) is green on the amended entry.

The remediation fix is right, and the ordering is the part that makes it work:

> **Render the parent View before direct selector-backed Region operations**, ensure the selector resolves within its parent element, or explicitly allow a missing element where supported.

Leading with the render puts the most likely post-migration cause first while keeping both original suggestions intact for the cases where they still apply. Importantly it does not promote `allowMissingEl` — that would have converted a loud failure into a silent no-op, which is the wrong direction for a diagnostic whose whole value here is being noisy.

One adjacent note, not a request for this PR. The catalog is now correct, but the *runtime* side of MN0005 is unchanged: the thrown error still carries `url: 'marionette.region.html#additional-options'`, so a developer following the link from a console error lands on Region element options rather than anywhere describing the render requirement. The catalog remediation is the machine-readable path — agents, tooling, generated docs — and that is the right layer to have fixed first. The human-facing `url` is the same `/errors/` versus `<page>.html#anchor` split I raised on #300, where MN0028 is currently the only diagnostic pointing at an error page. Whenever that convention gets settled one way or the other, MN0005's `url` is a good candidate to move with it, since it now has a documented cause its anchor does not cover.

Nothing blocking.

PR #309, opened 2026-08-30
chore(diagnostics): preserve retired lifecycle codes

paulfalgout:
Related #147

## What
- add a `retired` diagnostic status and bump the internal catalog schema to version 2
- restore MN0028 and MN0029 as permanently reserved historical identities without restoring their runtime throws
- reject retired runtime/lint mappings and keep generated reference routes with clearly historical severity and surface labels

## Why
MN0028 and MN0029 reached merged public master before their strict lifecycle errors were replaced with lifecycle-safe no-ops. Deleting them left their published identities available for accidental reuse despite the catalog policy prohibiting deletion and reuse. `retired` is distinct from `deprecated`: it is never emitted and has no replacement.

The original severity and surfaces remain stored as historical classification; `status` is authoritative and the docs mark those values as historical for display. The permanent routes avoid dead references to the former identities. Schema version 2 explicitly signals the new enum value.

## Scope
- diagnostic catalog schema, validation, and generated reference pages
- diagnostic policy documentation and focused validator/docs tests
- no production runtime, package entrypoint, lifecycle behavior, or distribution changes
- app-frontend and marionette.toolkit were audited and do not consume Marionette diagnostic codes or numbering

## Deployment Impact
No application or form redeploy is required. This PR does not publish documentation, npm artifacts, tags, releases, or deployment refs.

## Testing
- `npm run coverage` — 93 files, 1,695 tests, 100% statements/branches/functions/lines, 19/19 agent benchmark contract tests
- `npm run check:diagnostics` — 29 catalog entries
- `npm run docs:check` — 41 pages, 42 HTML files, 386 internal links
- `npm run lint:ci`
- `git diff --check`

Test-first evidence: focused tests initially failed because `retired` was not schema-valid and MN0028/MN0029 were absent. The final focused catalog suite passes 36 tests, including generic retired runtime/lint rejection, exact MN0028 reintroduction rejection, no retired slug reuse, preserved raw historical values, and generated historical labels.

Claude review challenged the status, schema-version, and historical-field decisions. The implementation now explicitly distinguishes retirement from deprecation, bumps the schema version, preserves raw classification values, labels them only at render time, and adds regression coverage. The follow-up review conditionally approved the direction; its two requested gaps are covered by the existing synthetic retired tests plus the added raw-value assertion.

---
## Summary by cubic
Adds a `retired` diagnostic status so MN0028 and MN0029 keep their published identities without being reusable. Previously, deleting those lifecycle errors left their codes open to accidental reuse; now they stay cataloged permanently and cannot be emitted or mapped.

- Bumps the catalog schema to version 2 to add the new enum value.
- `check:diagnostics` rejects retired runtime emissions and lint mappings.
- Generated reference pages label retired severity and surfaces as historical.
- No production runtime, package, or distribution changes, so no redeploy is needed.

<sup>Written for commit 85f68ff57294bf8e52e294f4de5871408920bf7d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/309?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/config/diagnostics/catalog.json
+++ b/config/diagnostics/catalog.json
@@ -1,6 +1,6 @@
 {
   "$schema": "./catalog.schema.json",
-  "schemaVersion": 1,
+  "schemaVersion": 2,
   "diagnostics": [
     {
       "code": "MN0001",
@@ -463,6 +463,39 @@
       ],
       "remediation": "Pass a non-empty string feature name to setEnabled; documented Marionette names and application-owned names are supported.",
       "docsAnchor": "/errors/MN0027/",
+      "surfaces": [
+        "runtime"
+      ],
+      "benchmarkCategory": "view"
+    },
+    {
+      "code": "MN0028",
+      "slug": "region-destroyed-operation",
+      "status": "retired",
+      "category": "lifecycle",
+      "severity": "error",
+      "objects": [
+        "Region"
+      ],
+      "remediation": "Calls to show, empty, or reset after Region destruction are lifecycle-safe no-ops. Use a live Region when the operation must take effect.",
+      "docsAnchor": "/errors/MN0028/",
+      "surfaces": [
+        "runtime"
+      ],
+      "benchmarkCategory": "region"
+    },
+    {
+      "code": "MN0029",
+      "slug": "view-destroyed-set-element",
+      "status": "retired",
+      "category": "lifecycle",
+      "severity": "error",
+      "objects": [
+        "CollectionView",
+        "View"
+      ],
+      "remediation": "Calls to setElement after View or CollectionView destruction begins are lifecycle-safe no-ops. Use a live instance when element replacement must take effect.",
+      "docsAnchor": "/errors/MN0029/",
       "surfaces": [
         "runtime"
       ],
--- a/config/diagnostics/catalog.schema.json
+++ b/config/diagnostics/catalog.schema.json
@@ -13,7 +13,7 @@
       "const": "./catalog.schema.json"
     },
     "schemaVersion": {
-      "const": 1
+      "const": 2
     },
     "diagnostics": {
       "type": "array",
@@ -56,7 +56,8 @@
           "enum": [
             "defined",
             "active",
-            "deprecated"
+            "deprecated",
+            "retired"
           ]
         },
         "category": {
--- a/docs/diagnostic-catalog.md
+++ b/docs/diagnostic-catalog.md
@@ -9,6 +9,7 @@ and the public agent benchmark. The catalog is stored in
 The catalog is static project metadata. Production entrypoints must not import the
 catalog, and the catalog is not part of the production package surface. Runtime
 diagnostics may embed a compact catalog code, but they must not load the full catalog.
+Schema version 2 adds explicit retired identities without restoring their emissions.
 
 ## Runtime error contract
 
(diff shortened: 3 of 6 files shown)


paulfalgout:
**Approved** at `85f68ff5` (the SHA I fetched). No runtime, `dist/`, or entrypoint changes, so no bundle movement and no growth gate.

This closes the finding I raised on #308, and goes further than what I suggested there in three ways worth naming.

**It enforces rather than records.** I proposed a status value; a status value alone is a comment. The validator now rejects any runtime source emitting a retired code and any ESLint rule mapping to one, with both paths tested. Reuse fails `check:diagnostics` at the emission site instead of depending on someone noticing.

**It reserves the slug too.** I only raised the code number. `does not allow a retired slug to be reused by another code` covers the other half of the identity — a future diagnostic claiming `region-destroyed-operation` would be exactly as confusing as one claiming MN0028, and I missed that.

**`retired` is kept distinct from `deprecated` rather than approximated by it.** Deprecated implies still emitted, with a replacement; retired means never emitted, no replacement. `rejects a replacement on a retired diagnostic` is what stops the two blurring later, which is the failure mode for status enums that grow by analogy.

The rewritten remediations are the detail I'd have missed entirely. MN0028 now reads *"Calls to show, empty, or reset after Region destruction are lifecycle-safe no-ops. Use a live Region when the operation must take effect."* A retired identity is exactly what someone arrives at from a stale source — an old log, a cached agent answer, alpha-era docs — so routing them to current behavior rather than leaving a dangling "this throws" claim is worth more than the reservation itself.

The `surfaces` overload was going to be my one note — a consumer filtering on `surfaces.includes('runtime')` would read MN0028 as live — and the policy already handles it explicitly: *"the places that report the diagnostic, or historically reported it for a retired entry"*, with `status` declared authoritative. Keeping the historical classification and disambiguating in the spec is the better call than blanking the fields.

## Non-blocking: the class of failure is still unguarded

The reservation is a hardcoded snapshot. `reserves the retired MN0028 and MN0029 identities` asserts `deep.equal` against exactly those two entries, which pins *these* codes permanently but establishes nothing about the next one. If MN0015 were dropped from the catalog tomorrow, every check here still passes.

That matters because the policy prohibiting deletion already existed before #308 — the PR body says so — and #308 deleted two codes through a green CI anyway. The rule was written down and unenforced, which is precisely the gap this PR closes for emission but not for deletion. The validator only ever sees the current catalog; it has no reference point for what used to be there.

A contiguity invariant would close it without needing history: codes today run MN0001 through MN0029 with no gaps — I checked — so requiring the set to be exactly `MN0001..MN{max}` is satisfiable right now and makes any future deletion a detectable hole. Cheaper than diffing against a released catalog, and it fails at the moment of deletion rather than whenever someone next reuses the number.

Nothing blocking.

paulfalgout:
## Exact-head Claude review

Claude reviewed head `85f68ff57294bf8e52e294f4de5871408920bf7d` after the final schema-version and raw-value assertions and reported **no hard blockers**. Its remaining confirmations are satisfied on this exact head:

- `scripts/diagnostics/catalog.mjs` rejects both runtime and lint mappings whenever `status` is `retired`; focused tests exercise the real validator with synthetic MN0001 and exact MN0028 emissions.
- `config/diagnostics/catalog.json` is the machine-readable index and retains `status: retired`, `severity: error`, and `surfaces: [runtime]` for MN0028/MN0029; docs add `historical` only while rendering.
- duplicate code, slug, and documentation-route validation remain generic, and deprecated-without-replacement is already schema-rejected.
- repository search found no MN0028/MN0029 runtime reference and no diagnostics schema-version equality check outside the catalog/schema/test contract; the two named consumers also contain no Marionette diagnostic-code handling.

Claude found the scope to match the smallest-complete-contract goal and requested no design changes. Cubic independently reports no issues, required CI is green, mergeability is clean, and there are zero unresolved threads.

PR #310, opened 2026-08-30
docs(dom): align jQuery adapter guidance

paulfalgout:
Related #147

## Linked issue

- Related #147

## What

- Correct current reference docs so native core remains `$el`-free while the explicitly configured jQuery adapter creates and synchronizes View/CollectionView `$el` and Behavior mirrors the host wrapper.
- Describe Behavior and UI lookup results through the configured DOM API instead of assuming jQuery in native core.
- Remove stale Region `$el` and jQuery event-delegation pseudocode from source comments, then refresh the generated source maps that embed those comments.

## Why

Two current-reference statements contradicted the shipped adapter and its tests: one said the adapter does not add `$el`, and another said Behaviors never have `$el`. Source comments also described native-capable paths as jQuery-specific or referred to unsupported Region `$el`.

Agents could infer contradictory DOM contracts depending on which current reference page or source comment they read. The corrected wording makes the native default, optional adapter, and intentionally unsupported v4 behaviors explicit.

## Public behavior contract

- Canonical behavior after this PR: native core has no `$el`; selecting `marionette/jquery-dom-api` before View construction gives View/CollectionView synchronized jQuery `$el`, Behavior mirrors its host wrapper, and `$()` returns adapter-shaped query results.
- Superseded behavior removed, if any: none; this PR removes only false documentation and stale implementation comments.
- Compatibility exception and removal condition, if any: `$el` remains an explicit opt-in compatibility surface justified by active app-frontend usage; it does not restore Backbone.View inheritance, wrapped/selector `el`, or jQuery event delegation.

Historical v3/v4 behavior was checked as evidence: View and CollectionView inherited Backbone.View `$el`/`$()`, and Behavior proxied the host. Behavior `$el` during `initialize` is an intentional tested v5 strengthening, not a claim of exact v4 timing.

## Scope

- Current DOM interaction and Behavior reference wording.
- Source comments in UI binding, Behavior lookup, Region element management, and native event delegation.
- Generated UMD source maps updated because they embed source comments; executable distribution files and runtime bytes are unchanged.
- No runtime logic, declarations, package exports, or migration behavior changed.
- app-frontend has extensive active `$el` and jQuery-shaped `view.$()` usage, so migration must explicitly select or compose the adapter; marionette.toolkit has no active dependency on this surface.

## Runtime cost

- Bundle: exact CI comparison reports 0 B executable artifact change.
- Startup/hot path: no executable logic changed.
- Allocations/retention: no production graph or structural proxy changed.

## Deployment Impact

No application or form redeploy is required. This PR does not publish the website, npm package, tags, releases, or deployment refs.

## Validation

## Testing

- `npx vitest run test/unit/jquery-dom-api.spec.js test/unit/view.ui-bindings.spec.js test/unit/behavior.spec.js --coverage.enabled=false --reporter=dot` — 3 files, 84 tests.
- `npm run check:dist` — generated artifacts current.
- `npm run lint:ci`.
- `npm run docs:check` — 41 pages, 42 HTML files, 386 internal links.
- `git diff --check`.

The first CI run correctly caught stale generated source maps after source-comment edits; the follow-up commit contains only the refreshed maps.

## Package and browser impact

- Browser contracts affected: none; existing native and optional-jQuery behavior is documented accurately.

## Rollback or deprecation

Revert the documentation/source-comment commit and its generated source maps together if a stated adapter contract is disproved. No runtime rollback or deprecation path is introduced.

## Review notes

Please distinguish the intended optional `$el` contract from broader v4 compatibility. The repository verifies jQuery `^4.0.0`/4.0.0; this PR makes no promise about future major versions.


paulfalgout:
**Not approving yet** at `68a75619` — one blocking CI failure with a concrete fix. The content of the change is correct; it just needs a `dist/` regeneration.

## `Verify distributable artifacts` fails, and it is not flake

Node 24 fails on the `check:dist` step. That looks impossible for a comments-and-docs PR, so I traced it rather than assuming a flaky run:

- `dist/marionette.js` on master contains **zero** comment lines — the build strips them, so the bundles themselves really are unaffected by comment edits.
- But `dist/marionette.min.js.map` and `dist/marionette.umd.js.map` carry `sourcesContent`, which embeds the **original source text, comments included**. Both currently contain the literal string `useful for doing jquery DOM lookups` — one of the comments this PR rewrites.
- `dist/` is byte-identical to master in this PR, so the committed maps still describe the old comments while the sources describe the new ones. `check:dist` is right to reject that.

Master's own CI at `93042b63` passed the same step, so `dist/` was in sync before this branch — this PR introduced the drift.

The fix is a rebuild and commit of `dist/`. I would expect only the two `.map` files to change; `marionette.js`, `.cjs`, `.umd.js`, and `.min.js` should come back byte-identical since comments are stripped from all four. The maps are not among the eight artifacts the bundle-size contract tracks, so this should not move sizes or trigger a growth gate.

Worth generalizing, because it will recur: **"comments only" does not imply "distribution unchanged"** in a repo that ships source maps with embedded content. Any source edit, including a pure comment or whitespace edit, is a `dist/` edit here. The Scope line saying no distribution artifacts changed is the assumption that led to the red build.

## The rest is correct

All six source edits are comment-only — I checked each `-`/`+` pair, and no executable line moves in `mixins/ui.js`, `modules/behavior.js`, `modules/view-region.js`, or `runtime/event-delegator.js`.

Both documentation corrections are accurate against shipped behavior, which matters because these were assertions of *absence* — the kind that quietly steer a reader away from a working feature:

- *"The adapter does not add `$el`"* was false. `view-region.js:815-819` computes `Dom.wrapEl && Dom.wrapEl(el)` and assigns `this.$el = wrappedEl` when the adapter provides `wrapEl`.
- *"Behaviors do not have a `$el` property"* was false. `behavior.js:40-41` mirrors `view.$el` in the constructor, and `_syncElement` at 92-93 re-mirrors it, which is what makes the new "including after `setElement()`" clause true rather than optimistic.

Adding **"before constructing Views"** to the adapter sentence is the most valuable part of this change. `$el` is established during `setElement` inside construction, so configuring the adapter afterwards leaves already-built Views without it — silently, with no diagnostic. That ordering requirement was previously undocumented, and it is exactly the kind of thing that produces a confusing bug report rather than an obvious failure.

Happy to re-review as soon as `dist/` is regenerated.

paulfalgout:
**Approved** at `d86e3c7c`. The regeneration is exactly the shape I expected, and I verified it directly rather than inferring it from a green check.

Only `dist/marionette.min.js.map` and `dist/marionette.umd.js.map` changed. All eight artifacts the bundle-size contract tracks — `marionette.js`, `.cjs`, `.umd.js`, `.min.js`, both Backbone shims, both jQuery DomApi files — are byte-identical to master, which confirms the diagnosis: comments never reached the bundles, only the maps' `sourcesContent`.

Content check on the maps themselves: the stale `useful for doing jquery DOM lookups` is gone and the replacement `configured DOM lookup scoped to the behavior` is present. So the maps now describe the sources that are actually in the tree.

Sizes are unmoved and `Bundle size` passes, as expected — maps are not tracked artifacts. `Build and validate`, `Node 26 advisory`, `Release preflight`, and the macOS package smoke are green; `Node 24`, which carries the `Verify distributable artifacts` step that failed before, was still running when I looked. The approval rests on the direct evidence above rather than on that check, but it should follow.

The documentation corrections and the six comment-only source edits are unchanged from my previous review and still stand — in particular the **"before constructing Views"** qualifier, which documents an ordering requirement that otherwise fails silently.

PR #312, opened 2026-08-30
docs(mnobject): reconcile the v5 contract

paulfalgout:
Related #147
Related #131

## What
- Reconcile the MnObject reference with the v5 named export, constructor options, opaque client IDs, built-in Radio, and Events API.
- Correct destruction ownership, lifecycle arguments, reentry/retry boundaries, and late-callback failure behavior.
- Remove the false v4 default-export/`Object` compatibility recipe and replace invalid placeholder examples with concrete v5 code.

## Why
- The page was copied from v4 and still taught the retired `backbone.marionette` package, default namespace, `Object` alias, and incorrect destroy cleanup/signatures.
- Agents and consumers need to distinguish `listenTo` ownership from direct `on` listeners, built-in Marionette Radio from `backbone.radio`, and retryable `before:destroy` failures from terminal later failures.

## Scope
- Documentation only: `docs/marionette.mnobject.md`.
- No runtime, lifecycle implementation, compatibility alias, distribution artifact, website manifest, or release behavior changes.
- app-frontend and marionette.toolkit both use named MnObject at runtime; neither requires the removed Object alias. Their custom destroy overrides and Radio migrations remain separate consumer work.

## Deployment Impact
- No application or package deployment is required.
- This PR does not publish the documentation website. The page remains outside `docs-site/next.json`.

## Testing
- `npx vitest run test/unit/object.spec.js test/unit/mixins/destroy.spec.js test/unit/mixins/radio.spec.js test/unit/common/merge-options.spec.js --coverage.enabled=false` — 4 files, 45 tests passed.
- `node test/fixtures/no-default-export/validate.mjs` — passed.
- Direct ESM smoke of the documented constructor, channelName, lifecycle arguments, and destroyed state — passed.
- `npm run docs:check` — 41 pages, 42 HTML files, and 386 links validated.
- `npm run lint:ci` — passed with zero warnings.

`docs:check` does not execute this page’s unmarked snippets because the page is not yet in the published v5 manifest. The direct smoke and focused runtime tests validate the changed claims without publishing or expanding the page’s release status.

---
## Summary by cubic
Updates the `MnObject` documentation to match the v5 contract: named `MnObject` export from `marionette`, constructor options via `this.options`, opaque client IDs, built-in Radio, and the Events API. Clarifies `destroy` lifecycle semantics, including listener ownership, retry boundaries, and failure behavior. Removes the v4 default-export and `Object` alias recipe.

- Corrects destruction ownership: `listenTo` subscriptions are unbound, `on` listeners are not.
- Documents reentrancy and retry boundaries for `before:destroy` and `destroy`.
- Replaces v4 compatibility examples with concrete v5 code.
- Only docs change; no runtime, release, or deployment impact.

<sup>Written for commit 8f2fce953f7de92ee2907a925c4cd4a37a55644a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/312?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approved** at `8f2fce95` (the SHA I fetched). Documentation only, one `.md` file, so no `dist/` implications.

This page makes precise behavioral claims rather than prose, so I checked them against the source instead of reading for plausibility. They hold — including the qualifiers, which are where this kind of page usually drifts.

- *"`isDestroyed()` is `false` during `before:destroy` and `true` during `destroy`"* — `mixins/destroy.js` sets `_isDestroyed = true` between the two `triggerMethod` calls. Exactly right, and it is the sort of detail that is invisible without reading the mixin.
- *"nested calls from either lifecycle event return the same MnObject without restarting teardown"* — the `_isDestroyed || _isDestroying` guard covers both windows.
- The retry boundary is correct in both directions. A throw from `before:destroy` hits `delete this._isDestroying; throw error`, leaving `_isDestroyed` unset so a later top-level call genuinely restarts; a throw after `_isDestroyed = true` is terminal because later calls early-return. *"make `before:destroy` callbacks safe to run again"* is the right advice to draw from that.
- *"Listeners registered directly on the instance with `on` are not removed automatically"* — `destroy` calls `stopListening()` and never `off()`. Correct, and worth documenting since the asymmetry is not guessable.
- The Radio sentence is the most carefully worded thing on the page. `bindEvents` routes through `listenTo`, so `stopListening()` releases subscriptions; replies are released by `_destroyRadio` calling `this._channel.stopReplying(null, null, this)`. The context argument is precisely why *"does not reset the shared Radio channel or remove unrelated channel handlers"* is true rather than merely hoped for.
- *"Own enumerable ... with values other than `undefined`"* — three qualifiers, and `merge-options.js` has all three: `propertyIsEnumerable.call(options, key)` (own-only), `option !== undefined`, and `setProperty` for the assignment.

## One consequence worth naming explicitly

`_destroyRadio` is registered with `this.on('destroy', ...)` and therefore runs *inside* `triggerMethod('destroy', ...)`. So a throwing `onDestroy` handler that runs before it can leave Radio **replies** registered on a channel that is shared with other objects.

The page's *"remaining cleanup may be interrupted"* is accurate and covers this. But it is the one interruption with blast radius beyond the instance: a leaked `listenTo` subscription is inert on a dead object, while a leaked reply keeps answering requests from unrelated code, and `stopReplying` was the thing that would have released it. Since the page already draws the owned-versus-shared distinction so carefully, saying that the shared channel is what a late failure can strand would sharpen it.

## Minor

*"Constructor options are shallow-copied into `this.options`"* is slightly narrower than what happens. `_setOptions` builds `assignOwn({}, getValue(this, 'options'), options)`, so a class-level `options` property participates — and because it goes through `getValue`, it may be a function resolved against the instance. Constructor options win the overlap, so the sentence is not wrong, but a subclass author defining `options` would not learn from this page that theirs is merged underneath.

Nothing blocking.

paulfalgout:
Claude review: Approved at exact head `8f2fce953f7de92ee2907a925c4cd4a37a55644a`; no blocking issues.

The exact diff was reviewed after qualifying cleanup on successful lifecycle completion and neutrally describing late failure behavior. I independently rechecked the two highest-risk facts at this SHA: `mixins/destroy.js` passes `(this, options)` to both lifecycle triggers, and the public npm/root entry is the named `MnObject` export from `marionette`. The focused destroy tests assert the argument pairs and reentry/retry state. `docs:check` validated the new cross-page fragments. The page remains unpublished and its unmarked snippets are not executed; the PR discloses that limitation and records the direct ESM smoke.

PR #313, opened 2026-08-30
docs(routing): define the v5 integration boundary

paulfalgout:
## What

- Define that Marionette v5 has no router export, routing dependency, or router integration protocol.
- Document `Backbone.Router` through the optional `marionette/backbone` shim, including the preserved router identity and unpatched `Backbone.History` boundary.
- Remove the inherited v4-era third-party router catalog and obsolete repository edit link.

## Why

The reference page still presented migration history and an undated list of third-party projects as the current contract. The replacement is based on the v3/v4 removal rationale and the current v5 runtime: routing remains application-owned, while Backbone users have an explicit optional integration path.

## Scope

- Documentation only: `docs/routing.md`.
- No runtime API, router implementation, compatibility alias, dependency, or published-site manifest changes.
- The third-party catalog is intentionally removed because the repository does not validate those projects as supported v5 integrations.

## Deployment Impact

No deployment or redeploy is required beyond the normal documentation pipeline. This does not publish the website, package, tag, or release.

## Testing

- `npm test -- test/unit/backbone-shim.spec.js` — 6 tests passed, including `Backbone.Router` identity/Event behavior and the `Backbone.History` non-patch.
- `npm run check:docs-examples` — 9 executable examples validated across 7 fixture validators.
- `npm run docs:check` — built 41 pages and validated 42 HTML files, 386 internal links, and both new heading fragments.
- `npm run lint:ci`
- `npm run test:fixtures`
- `git diff --check`

---
## Summary by cubic
Rewrites `docs/routing.md` to define Marionette v5's integration boundary: no router export, routing dependency, or integration protocol. Backbone users have an explicit path through the optional `marionette/backbone` shim, and other routing libraries can be used directly with ordinary Marionette APIs.

- The shim preserves `Backbone.Router` identity, adds Marionette `Events` methods, and does not patch `Backbone.History`.
- Removes the v4-era third-party router catalog and the obsolete repository edit link because those projects aren't validated as supported v5 integrations.
- Documentation-only change: no runtime API, dependency, or published-site manifest updates.

<sup>Written for commit d07c193cda2cfa8a5c8cf02b389c1ed008ac66ea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/313?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -1,45 +1,42 @@
 # Routing in Marionette
 
-Users of versions of Marionette prior to v4 will notice that a router is no longer a [bundled class](./classes.md).
-The [Marionette.AppRouter](https://github.com/marionettejs/marionette.approuter) was extracted
-and the core library will no longer hold an opinion on routing.
+Marionette does not export a router, depend on a routing library, or require a
+routing integration protocol. Route handlers are ordinary application code and
+can coordinate Marionette objects without a Marionette-specific adapter.
 
-## Some Routing Solutions
+## Using `Backbone.Router`
 
-Besides the router [bundled with Backbone](http://backbonejs.org/#Router) there are many viable
-routing solutions available.  Some specifically designed for Backbone or Marionette and some
-that are generic solutions for any framework.  Here are a few of those options.
+Applications that choose Backbone can load the
+[bundled Backbone shim](./optional-backbone.md#using-the-bundled-backbone-shim)
+before constructing routers:
 
-## Marionette Community Routers
+```javascript
+import 'marionette/backbone';
+import Backbone from 'backbone';
 
-### [Marionette.AppRouter](https://github.com/marionettejs/marionette.approuter)
+const Router = Backbone.Router.extend({
+  routes: {
+    '': 'home'
+  },
 
-Previously bundled router. Extends [backbone.router](http://backbonejs.org/#Router) and is helpful
-for breaking a large amount of routes on a single backbone.router instance into smaller more managable
-approuters.
+  home() {
+    // Navigate the application with its normal Marionette APIs.
+  }
+});
 
-### [Marionette.Routing](https://github.com/blikblum/marionette.routing)
+new Router();
+Backbone.history.start();
+```
 
-An advanced router for MarionetteJS applications. Includes nested routes, states, rendering,
-async operations, lazy loading routes, Radio channel eventing, and inherits most of CherryTree
-features while maintaining a similar to Marionette API.
+The shim preserves the identity of [`Backbone.Router`](https://backbonejs.org/#Router)
+and adds Marionette's `Events` methods to its prototype. It does not add a
+Marionette router or patch `Backbone.History`; routing and history behavior remain
+Backbone contracts.
 
-### [Backbone.Eventrouter](https://github.com/RoundingWellOS/backbone.eventrouter)
+## Using Other Routers
 
-A highly opinionated, simplistic Backbone.Router coupled with a Radio channel.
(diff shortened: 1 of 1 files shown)


paulfalgout:
**Approved** at `d07c193c` (the SHA I fetched). Documentation only, one `.md` file, no `dist/` implications.

Every factual claim on the page checks out against the tree:

- *"does not export a router"* — no router export in `index.js`.
- *"depend on a routing library"* — `peerDependencies` carries only `backbone` and `jquery`, both marked optional; no routing package anywhere.
- *"The shim ... adds Marionette's `Events` methods to its prototype"* — `Backbone.Router.prototype` is in the shim's patch list, and the module re-exports the same `Backbone` object, so the identity claim holds literally.
- *"does not ... patch `Backbone.History`"* — the list is Model, Collection, View, Router. History is genuinely absent.

Removing the third-party router catalog is the right call and worth defending explicitly: an undated list of unvalidated projects sitting under a current-reference heading reads as endorsement, and for an agent it reads as a supported integration matrix. Deleting it is more honest than letting it rot, and *"Other routing libraries can be used directly"* conveys everything the list was actually load-bearing for. Pointing the AppRouter history at the v3→v4 guide keeps the rationale reachable without presenting it as current contract.

## Worth adding: which object sits on which side

The page says the shim *"does not ... patch `Backbone.History`"* and that *"routing and history behavior remain Backbone contracts"*. Both true, but the consequence is easy to miss in a routing context specifically.

After the shim, `Backbone.Router.prototype` carries **Marionette's** `Events`, while `Backbone.history` — the singleton every router in the example depends on, started on the last line — keeps **Backbone's**. So the two objects an application touches while wiring routes are on opposite sides of the boundary, and an application doing `this.listenTo(Backbone.history, 'route', handler)` from a Marionette object is crossing implementations.

I have not tested whether that interop works, so I am not claiming it is broken — Backbone's `on`/`off` may well accept what Marionette's `listenTo` hands them. But this page exists to let a reader or an agent infer the boundary without experimenting, and naming `Backbone.history` as the unpatched object in a routing setup is exactly the kind of thing it is well placed to say.

The *"before constructing routers"* ordering advice is sound and worth keeping for the same reason — a Router constructed ahead of the shim would register through Backbone's methods and later unregister through Marionette's.

Nothing blocking.

paulfalgout:
Claude exact-head review: **approved** for `d07c193cda2cfa8a5c8cf02b389c1ed008ac66ea` with no blocking issues.

Review scope: the complete final `docs/routing.md`, current shim boundary, v3/v4 removal rationale, and exact local validation evidence.

Verified on this head:

- `npm test -- test/unit/backbone-shim.spec.js` — 6 passed, including Router identity/Event behavior and the direct `Backbone.History` non-patch assertion.
- `npm run check:docs-examples`
- `npm run docs:check` — 42 HTML files and 386 internal links, including fragment validation.
- `npm run lint:ci`
- `npm run test:fixtures` — includes ESM/CJS `marionette/backbone` package-subpath coverage.
- `git diff --check`

The review initially requested clearer wording and direct anchor/History evidence. The wording was tightened before the commit; the link checker and focused shim test provide the requested evidence. No further changes were recommended.

PR #314, opened 2026-08-30
docs(behavior): make examples native v5

paulfalgout:
Related #147

## What

- Replace retired `backbone.marionette` imports and incidental jQuery tooltip calls in the Behavior reference with the canonical `marionette` package and native DOM behavior.
- Remove unowned v4-era JSFiddle links and clarify the View/Behavior initialization boundary.
- Execute the defining/attaching example in the existing packed-package fixture and pin its click, destroy, native UI, and constructor-argument contracts.

## Why

The page mixed recently reconciled v5 Behavior contracts with examples inherited from the jQuery-first v3/v4 environment. Those examples made a third-party `.tooltip()` plugin and the retired package name look like Marionette v5 APIs. The updated examples keep the historical intent—reusable warning and tooltip behavior—without locking core Behavior UI to the optional jQuery adapter.

## Scope

- Behavior reference documentation and its existing executable fixture.
- One focused lifecycle contract test for explicit Behavior state versus unchanged host constructor arguments.
- No runtime implementation, jQuery adapter, public alias, fallback path, site manifest, or unrelated documentation changes.

## Deployment Impact

No application redeploy is required. The normal documentation pipeline validates the change, but this PR does not publish the website, package, tag, or release.

## Testing

- `npx vitest run test/unit/behavior-dom-delegation-contract.spec.js test/unit/behavior-lifecycle.spec.js test/unit/behavior.spec.js test/unit/behavior-dependencies-contract.spec.js test/unit/behavior-composition.spec.js test/unit/mixins/behaviors-composition.spec.js test/unit/mixins/behaviors.spec.js test/unit/behavior-ui-contract.spec.js test/unit/behavior-communication-contract.spec.js --reporter=dot` — 9 files, 130 tests passed.
- `npm run test:fixtures` — all packed-package fixtures passed, including the revised Behavior docs example.
- `npm run docs:check` — 10 executable examples, 41 pages built, 42 HTML files and 386 internal links validated.
- `npm run lint:ci`
- `git diff --check`
- Stale-pattern scan confirmed no retired package imports, `.tooltip()` calls, JSFiddle links, or corrected phrases remain in `docs/marionette.behavior.md`.

---
## Summary by cubic
Updates the Behavior reference examples to use native Marionette v5 APIs instead of the retired `backbone.marionette` package and the jQuery `.tooltip()` plugin.

- Replaces remaining `backbone.marionette` imports and jQuery tooltip calls with `marionette` and native DOM `title` assignment.
- Removes v4-era JSFiddle links and clarifies that Marionette does not inject Behavior state into the View's constructor arguments.
- Adds the defining/attaching example as an executable fixture covering click, destroy, native UI, and options behavior.
- Adds a lifecycle test confirming Behavior state is set explicitly and host constructor arguments pass through unchanged.

<sup>Written for commit 70aa9cad36e690a9ed54ed3c020f58510acdbcde. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/314?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Updated Behavior API examples to use current import paths and native DOM updates.
  * Clarified View and Behavior initialization order and constructor argument handling.
  * Added an executable example demonstrating collaborator injection and option access.
  * Removed outdated live-example links and expanded behavior attachment guidance.

* **Tests**
  * Added coverage for executable documentation examples, rendering, tooltips, destruction, and alerts.
  * Verified View constructor arguments are preserved while Behavior initialization receives the expected options.

### DIFF
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -13,7 +13,7 @@ allowing you to share common user-facing operations between your views.
 collection interactions to be utilized across your application. Unlike the other
 Marionette classes, `Behavior`s are not meant to be instantiated directly.
 Instead a `Behavior` should be instantiated by the view it is related to by
-[attaching the a behavior class definition to the view](#using-behaviors).
+[attaching a behavior class definition to the view](#using-behaviors).
 
 ## Documentation Index
 
@@ -44,9 +44,13 @@ The easiest way to see how to use the `Behavior` class is to take an example
 view and factor out common behavior to be shared across other views.
 
 ```javascript
-import { View } from 'backbone.marionette';
+import { View } from 'marionette';
 
 const MyView = View.extend({
+  template() {
+    return '<button class="destroy-btn" type="button">Destroy</button>';
+  },
+
   ui: {
     destroy: '.destroy-btn'
   },
@@ -61,14 +65,10 @@ const MyView = View.extend({
   },
 
   onRender() {
-    this.ui.destroy.tooltip({
-      text: 'What a nice mouse you have.'
-    });
+    this.getUI('destroy')[0].title = 'What a nice mouse you have.';
   }
 });
 ```
-
-[Live example](https://jsfiddle.net/marionettejs/pa8ryv03/)
 
 Interaction points, such as tooltips and warning messages, are generic concepts.
 There is no need to recode them within your Views so they are prime candidates
@@ -76,8 +76,9 @@ to be extracted into `Behavior` classes.
 
 ### Defining and Attaching Behaviors
 
+<!-- executable-example: behavior-defining-attaching -->
 ```javascript
-import { Behavior, View } from 'backbone.marionette';
+import { Behavior, View } from 'marionette';
 
 const DestroyWarn = Behavior.extend({
   // You can set default options
@@ -90,7 +91,7 @@ const DestroyWarn = Behavior.extend({
     destroy: '.destroy-btn'
   },
 
-  // Behaviors have events that are bound to the views DOM.
+  // Behaviors have events that are bound to the view's DOM.
   events: {
     'click @ui.destroy': 'warnBeforeDestroy'
   },
@@ -114,18 +115,21 @@ const ToolTip = Behavior.extend({
   },
 
   onRender() {
-    this.ui.tooltip.tooltip({
-      text: this.getOption('text')
-    });
+    this.getUI('tooltip')[0].title = this.getOption('text');
   }
 });
 
-const MyView = View.extend({
+export const MyView = View.extend({
+  template() {
+    return [
+      '<button class="destroy-btn" type="button">Destroy</button>',
+      '<span class="tooltip">More information</span>'
+    ].join('');
+  },
+
   behaviors: [DestroyWarn, ToolTip]
 });
 ```
-
-[Live example](https://jsfiddle.net/marionettejs/b1awta6u/)
 
 Each behavior will now be able to respond to user interactions as though the
 event handlers were attached to the view directly. In addition to using array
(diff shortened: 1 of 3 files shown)


paulfalgout:
Claude exact-head review: **approved** for `70aa9cad36e690a9ed54ed3c020f58510acdbcde` with no blocking issues.

The review covered the complete three-file commit, including the v3/v4 jQuery-era precedent, canonical v5 package name, native UI boundary, lifecycle wording, and executable evidence.

Verified for this head:

- `package.json` names the public package `marionette`; the v4 migration guide records the rename from `backbone.marionette`.
- Existing lifecycle tests prove host `preinitialize` precedes Behavior initialization for both View and CollectionView.
- The new focused test proves the documented View-specific boundary: Behavior state is visible only when explicitly set, while View `initialize` receives the original constructor arguments exactly.
- The packed docs fixture renders the documented example, verifies native `title`, clicks the Behavior event target, checks its warning, and confirms host destruction.
- 9 focused Behavior files / 130 tests, `npm run test:fixtures`, `npm run docs:check`, `npm run lint:ci`, and `git diff --check` passed locally.

Claude found no factual overreach, invented compatibility, or merge blocker. Its remaining suggestions were optional future coverage, not changes required for this slice.

paulfalgout:
**Approved** at `70aa9cad` (the SHA I fetched). More than a docs pass — this makes the page's central example verifiable, and in doing so fixes an example that could not have worked.

## The previous example was not runnable

On master, the `MyView` in *Defining and Attaching Behaviors* has no `template`. It renders nothing, so `.destroy-btn` and `.tooltip` never exist, `ui` resolves to nothing, and `'click @ui.destroy'` binds to nothing. Anyone copying it got a silent no-op and no error to explain it.

Compounding that, `this.ui.destroy.tooltip({ text: ... })` is a jQuery UI / Bootstrap plugin call — not Marionette, and not jQuery core. It would throw even for an application that had explicitly configured the optional jQuery adapter, because the adapter supplies jQuery, not that plugin. The page was effectively documenting a third-party dependency it never declared.

So adding the templates and switching to `this.getUI('tooltip')[0].title = ...` is a correctness fix rather than modernization. `export const MyView` is what lets the fixture reach it.

## The fixture extracts the documented code, which is the important part

`marker: '<!-- executable-example: behavior-defining-attaching -->'` means the snippet on the page *is* the code under test. There is no parallel copy to drift out of sync — the failure mode where an example rots while a lookalike test keeps passing is structurally excluded.

The assertions are consequences rather than restatements:

- `tooltip.title === 'Tooltip text'` is what makes the jQuery-to-native replacement **verified**. Asserting the element merely exists would pass even if the Behavior never set anything.
- `destroyButton.click()` producing `['You are destroying!']` and `view.isDestroyed() === true` exercises the whole path the page is teaching — Behavior `ui` binding, the `@ui`-scoped event, `getOption` defaulting, and `this.view.destroy()` reaching the host.
- Stubbing `window.alert` and restoring it in `finally`, with a destroy fallback if an assertion left the View alive, keeps a failure here from cascading into unrelated fixtures.

## The removed sentence was actively wrong

> The view `initialize` is called later with the information eventually injected by the behavior.

That reads as though the Behavior's injection arrives through the host's `initialize` **arguments**. It does not. The new test pins the real model precisely: `calledWithExactly(options, extra)` with `to.equal` identity checks proves the host receives its untouched constructor arguments, while `this.behaviorState` shows the injected state arriving on the instance. Asserting argument identity rather than shape is what makes that distinction airtight.

Dropping the JSFiddle links is right for the same reason as #313's router catalog — unowned, v4-era, and they would demonstrate the retired package under a current-reference heading.

Nothing blocking.

PR #316, opened 2026-08-30
fix(radio): preserve owner-scoped request cleanup

paulfalgout:
Related #147

## What

- Restore owner context when selectively removing Radio request replies.
- Replace stale v4 common-helper documentation with the owned v5 contracts.
- Add real-Radio ownership regressions and packed executable docs examples.

## Why

The common-reference audit found that v5 dropped the owner context that v4.1.3 passed to `stopReplying`. On a shared channel, owner A could therefore remove owner B's replacement reply when both used the same request name and callback. The same page still described Backbone-owned APIs, channel-wide cleanup, retired package imports, broken examples, and unowned JSFiddles.

## Scope

Impacts selective `unbindRequests`, generated runtime artifacts, common-helper documentation, focused unit coverage, the alpha.2 changelog, and one packed-doc fixture. No alias, fallback, repository policy, release, tag, package publication, or website publication is included.

## Agent-development failure addressed

- Makes owner-scoped cleanup, raw versus merged options, `getOption`, and `triggerMethod` ordering explicit and executable instead of requiring agents to infer v5 behavior from v4 prose.
- Removes misleading global-unbind wording that could cause an agent to break sibling owners on shared channels.

## Public behavior contract

- Canonical behavior after this PR: selective and no-map `unbindRequests` cleanup removes only replies owned by the caller; other owners remain registered.
- Superseded behavior removed: selective cleanup without context could remove another owner's replacement reply.
- Compatibility exception and removal condition, if any: none. The fix restores v4.1.3 owner scoping and removes no supported API.

## Runtime cost

- [ ] Static or documentation only
- [ ] Development or test only; excluded from production entrypoints
- [x] Changes an existing production path
- [ ] Adds an explicitly opt-in runtime path

Measured impact or explanation:

- Bundle: one existing `stopReplying` call receives the existing owner argument; generated artifacts are updated.
- Startup/hot path: no startup change; one extra argument during selective cleanup only.
- Allocations/retention: no new allocation; prevents accidental removal of another owner while preserving expected cleanup.

## Validation

- [x] Tests cover the acceptance criteria and relevant edge cases.
- [x] Docs, examples, declarations, diagnostics, and rule metadata agree where relevant.
- [x] No private consumer, private fixture, or undocumented context is required.
- [x] I ran the commands listed below and am reporting their actual results.

Commands:

```sh
[...]

PR #317, opened 2026-08-30
docs(utils): define standalone helper contracts

paulfalgout:
Related #147

## Linked issue

- Related #147; this completes one bounded utility-reference slice of the broader agent-oriented documentation work.

## What

- Reframe `extend` as Marionette-owned pseudo-class inheritance and replace the invalid legacy example with an executable packed-package contract.
- Document target-first utility exports, owner-scoped event/request cleanup, option precedence, fresh method normalization, and the package-backed `VERSION` value.
- Add a packed `docs-utils-contract` fixture that executes both examples against the current tarball.

- Removes stale Backbone ownership language, invalid placeholder JavaScript, and an unowned JSFiddle that could lead coding agents toward the retired package/API model.
- Makes standalone helper signatures and cleanup ownership executable and discoverable from the current package.

## Public behavior contract

- Canonical behavior after this PR: `extend` is Marionette-owned; exported Common helpers take their target first and retain the current owner-scoped, options-first, and fresh-normalization semantics documented in `docs/common.md`; `VERSION` matches installed package metadata.
- Superseded behavior removed, if any: documentation no longer presents `extend` as a Backbone alias or recommends the legacy JSFiddle/placeholder example.
- Compatibility exception and removal condition, if any: none. This PR adds no alias, fallback, or dual path.

## Runtime cost

- Bundle: hosted bundle validation reports 0 B change across every production artifact.
- Startup/hot path: no production runtime code changes.
- Allocations/retention: no production module or retained-resource changes.

## Validation

Commands:

```sh
npm run test:fixtures
npx vitest run test/unit/utils/extend.spec.js test/unit/utils/proxy.spec.js test/unit/backbone.marionette.spec.js --reporter=dot
npm run docs:check
npx eslint test/fixtures/docs-utils-contract/validate.mjs --max-warnings=0
git diff --check
```

Actual results: the complete packed-fixture matrix passed; focused utility coverage passed 3 files / 39 tests; docs validation passed 14 executable examples, 41 pages, 42 HTML files, and 386 internal links; fixture ESLint and diff checks passed.

## Package and browser impact

- Entry points or install fixtures affected: no entrypoint changes; one new packed documentation fixture validates the existing root `marionette` entrypoint.
- Production/dev/test module-graph impact: production graph unchanged; test-only fixture added.

## Scope

This changes `docs/utils.md` and adds one documentation fixture. It does not change runtime behavior, exports, compatibility paths, package metadata, release policy, or website publication.

## Deployment Impact

No application or form redeploy is required. The documentation website is not published by this pull request; pull-request documentation CI only builds and validates the site artifact.

## Rollback or deprecation

Reverting this single documentation commit restores the prior page and removes the fixture. No data, runtime migration, or staged deprecation is involved.

## Review notes

Claude approved exact head `d26f73241283e93efec081414edf55599db3833a` after the trigger-ordering and option-precedence assertions were strengthened. Cubic found no issues on the same head. CodeRabbit generated no actionable comments; its template warning is addressed by this body.


paulfalgout:
Claude exact-head review: **approved** for `d26f73241283e93efec081414edf55599db3833a` with no remaining content blocker.

The bounded review received the complete commit patch plus the current implementations of `extend`, the target-first proxies, Common helper methods, diagnostic URL construction, root exports, package scripts/exports, packed-fixture harness, and the v4 migration ledger. Claude verified the documentation and fixture claims line by line.

Feedback addressed before this exact head:

- Added an event listener and ordered call log so the fixture proves `triggerMethod` invokes the `on*` method before the event and returns the method result.
- Added a conflicting direct `enabled = true` property while constructor options contain `false`, proving defined options take precedence rather than merely testing falsy handling.
- Documented why the fixture constructs marker text indirectly, preventing the ownership scanner from counting fixture markers as documentation examples.

Claude confirmed the owned `extend`, target-first signatures, owner-scoped cleanup, fresh `normalizeMethods` result, selected own-enumerable option copying, and package-backed `VERSION` statements match the supplied source. It found no factual overpromise or compatibility freeze.

Exact-head validation: full `npm run test:fixtures` rebuilt/tested dist, packed the current tree, and installed that tarball into every fixture; focused utility suites passed 3 files / 39 tests; `npm run docs:check` validated 14 examples, 41 pages, 42 HTML files, and 386 links; fixture ESLint and `git diff --check` passed.

Cubic, hosted CI, mergeability, and unresolved-thread gates remain independently required.

paulfalgout:
**Approved** at `d26f7324` (the SHA I fetched). Documentation plus one packed fixture; no runtime or `dist/` changes.

## The executable-docs mechanism is self-policing, which is what makes this worth doing

I checked whether a new fixture actually runs and whether a marker can go stale, since "executable example" is only worth the name if both hold:

- `test/fixtures/run.mjs` discovers fixtures by directory scan and throws if it finds none, so adding `docs-utils-contract/` is sufficient — there is no registration step to forget.
- `check-executable-examples.mjs` enforces a **bidirectional one-to-one** correspondence: a documented marker with no fixture validator is an error, a fixture marker with no documentation example is an error, and duplicates on either side are errors.

So an example cannot *look* executable without being executed, and a fixture cannot keep pointing at a marker someone renamed. That is the property that separates this from a test that merely resembles the docs.

It also explains something in the fixture that reads like fussiness but is not. Building the marker as `` `<!${'--'} executable-example: ${id} --${'>'}` `` rather than a plain template literal keeps the *source text* from containing a marker whose id is the literal `${id}` — which the checker would register as a validator marker and then reject as having no documentation example. The concatenation is load-bearing.

## The assertions are chosen well

- `assert.notEqual(normalized, normalizedInput)` — an **identity** check, so it pins that `normalizeMethods` returns a fresh map. A `deepEqual` would pass just as happily if the input had been mutated in place, which is the actual failure being guarded.
- `assert.equal(proxyExample.unrelatedReplyAfterCleanup, 'other')` — this re-verifies #316's owner-scoped cleanup **against the packed tarball**. Layering matters here: #316's unit test proves the source is right, this proves the shipped artifact is. Those are different claims, and a build or export regression would only fail the second.
- `VERSION` compared against the *installed* `node_modules/marionette/package.json` catches a stale constant baked into `dist`. A source-level test structurally cannot detect that, because it would read the same repo file the constant was generated from.
- `assert.equal(proxyExample.optionValue, false)` pins that a legitimately falsy option is returned rather than treated as absent — the same failure class as the `context || callback` collapse fixed in #308.
- `SpecialService.extend === extend` is the assertion that makes it *pseudo-class* inheritance rather than plain prototype chaining: `extend` propagates to subclasses, so a subclass can itself be extended.

`Radio.reset()` at the end keeps channel state from leaking into whatever runs next, which matters more than usual now that two fixtures exercise real channels.

Nothing blocking.

PR #318, opened 2026-08-31
fix(view): no-op terminal entity delegation

paulfalgout:
Related #131

## Linked issue

- Related #131

## What

- Make base `View#delegateEntityEvents`, `CollectionView#delegateEntityEvents`, and direct attached `Behavior#delegateEntityEvents` calls chainably no-op once owning-View destruction starts.
- Add transition-focused tests for callable model and collection maps on live, destroying, and destroyed hosts.
- Define entity-event ownership, arguments, resolver timing, safe rewiring, cleanup, terminal delegation, event-map diagnostics, and optional Backbone boundaries in current docs with a packed-package executable fixture.

## Why

Late entity delegation could evaluate application callbacks and create fresh subscriptions after teardown had begun. That made async cleanup fragile and could leak a direct Behavior subscription from its `onDestroy` window. The terminal no-op keeps lifecycle cleanup safe without adding new state or changing live-host delegation.

- Coding agents previously had to infer handler context, resolver timing, model replacement order, and teardown behavior from distributed implementation details.
- The current reference now states an inspectable terminal contract and the executable fixture verifies the important ownership and package-boundary claims.

## Public behavior contract

- Canonical behavior after this PR: base host and attached-Behavior entity delegation returns the receiver without resolving maps or binding once owning-View destruction begins; live-host delegation and teardown undelegation remain unchanged.
- Superseded behavior removed, if any: late base delegation no longer evaluates maps or creates subscriptions during or after host destruction.
- Compatibility exception and removal condition, if any: none. Reusing a Behavior after calling `Behavior#destroy()` while its host remains live is outside this contract.

## Scope

- Affects the existing entity-event delegation path for View, CollectionView, and attached Behavior instances.
- Adds current-reference documentation and one isolated packed-package docs fixture.
- Does not change event emission, live-host delegation, `undelegateEntityEvents()`, Behavior ownership, or optional Backbone/jQuery adapter entrypoints.

## Runtime cost

- Bundle: two constant-time lifecycle guards; generated distribution artifacts updated.
- Startup/hot path: two boolean reads before an explicit delegation operation; no startup loop or render-path change.
- Allocations/retention: no new allocation or state; prevents late listener retention.

## Deployment Impact

- No website, package, tag, or release publication is performed by this PR.
- No consumer redeploy is required before merge; the behavior ships with the next normal Marionette package release.
- No repository or organization settings change is required.

## Validation

Commands:

```sh
npx vitest run test/unit/mixins/view.spec.js test/unit/collection-view/collection-view-viewmixin.spec.js test/unit/behavior.spec.js test/unit/mixins/delegate-entity-events.spec.js --reporter=dot
npm test
npm run test:source
npm run test:fixtures
npm run lint:ci
npm run docs:check
npm run check:dist
git diff --check
```

Results: 130 focused tests passed; the full suite passed 93 files and 1,706 tests; source, packed fixtures, lint, docs (41 pages and 386 links), distribution checks, and whitespace validation passed.

## Testing

- Unit tests prove callable host and Behavior model/collection resolvers remain functional on live hosts and are not evaluated during or after destruction.
- Reentrant teardown tests trigger entities before final cleanup, distinguishing prevention of transient rebinding from ordinary end-of-destroy cleanup.
- The packed fixture verifies handler ownership, unchanged arguments, explicit entity replacement, cleanup, unrelated-listener preservation, terminal chainability, and plain Backbone entities without importing the shim.

## Package and browser impact

- Entry points or install fixtures affected: main package distribution artifacts and the new `docs-entity-events` packed fixture; no export-map change.
- Browser contracts affected: only explicit entity-event delegation after View destruction starts.
- Production/dev/test module-graph impact: no new production dependency or entrypoint; the fixture installs Backbone and jsdom only in its isolated validation directory.

## Rollback or deprecation

- Revert both logical commits if an active consumer demonstrates a required late-delegation contract; no compatibility fallback or dual path is retained.

## Review notes

- Historical v3/v4 behavior was investigated as evidence; no documented rationale for late rebinding was found.
- `app-frontend` and `marionette.toolkit` contain no direct calls or overrides of these delegation methods.
- Claude reviewed exact head `cf4245252da986a74786a4cee7f83ad160856140` and approved with no blockers. Its earlier review identified the direct Behavior teardown leak, which this head closes; its fixture follow-up led to executable verification of the plain Backbone claim.

---
## Summary by cubic
Makes `View#delegateEntityEvents`, `CollectionView#delegateEntityEvents`, and direct `Behavior#delegateEntityEvents` calls no-op and return the receiver once the owning View's destruction starts (related to #131). Before, such late calls could still resolve entity-event maps, run application callbacks, and bind fresh subscriptions during teardown, making async cleanup fragile and potentially leaking a direct Behavior subscription from its `onDestroy` window. Live-host delegation and `undelegateEntityEvents()` are unchanged, and no migration or rollout step is required.

**Testing and docs**

- Adds unit tests covering callable model and collection maps on live, destroying, and destroyed hosts.
- Adds an executable packed-package fixture verifying handler ownership, arguments, rewiring, cleanup, terminal chainability, and plain Backbone entities without the shim.
- Rewrites `docs/events.entity.md` to state the terminal delegation contract and adds a matching row to the v4 migration table.

<sup>Written for commit cf4245252da986a74786a4cee7f83ad160856140. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/318?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
Claude exact-head review for `cf4245252da986a74786a4cee7f83ad160856140`: I sent the complete in-scope runtime, tests, docs, fixture diff, lifecycle boundary, historical/consumer findings, and validation evidence. Claude approved with no blockers. An earlier pass identified the direct attached-Behavior teardown leak, so the final runtime guard covers both host and direct Behavior entry points. Its follow-up on the optional-Backbone documentation led to a packed-fixture assertion using a plain `Backbone.Model` without importing `marionette/backbone`. No further code change was requested for this exact head.

paulfalgout:
**Approved** at `cf424525` (the SHA I fetched). +51 B aggregate (62,944 B against the 75,000 B ceiling, +0.08%), well under the growth threshold; `dist/` regenerated alongside the source.

## No authorization machinery needed here, and I checked rather than assumed

#302 needed the WeakMap because `destroy()` routed through the very methods being guarded. That risk does not exist here: the only callers of `delegateEntityEvents` are the two constructors (`view-region.js:799`, `collection-view.js:109`) and `_delegateBehaviorEntityEvents`, which is reached from `View#delegateEntityEvents` itself. Nothing on a teardown path invokes it, so a plain flag check is sufficient.

Correspondingly, `undelegateEntityEvents` is left unguarded on both classes — right, since cleanup has to run precisely when the guarded direction is closed.

The Behavior guard keys on `this.view._isDestroying || this.view._isDestroyed` rather than the Behavior's own state, which is what "once **owning-View** destruction starts" requires. The dereference is safe unconditionally: `this.view = view` is the first statement of the Behavior constructor and nothing deletes it, so there is no window where a constructed Behavior has no `view` to consult.

## "not evaluate or bind" is the assertion that makes the guard placement matter

The parameterized tests assert that a callable `modelEvents` / `collectionEvents` map is **not evaluated**, not merely that nothing gets bound. That distinction is the whole point of putting the guard first: a check placed after map resolution would bind nothing while still invoking application code on a host that is already tearing down — silently, and at the least predictable moment. Asserting only "no subscriptions exist afterward" would pass in that broken arrangement.

`should not delegate a retained behavior after its owning view is destroyed` covers the leak named in the description — a Behavior that outlives its host and re-delegates from its own `onDestroy` window, creating subscriptions after cleanup has already run past them.

Nothing blocking.

Issue #41, opened 2026-05-21
v5.0.0 final tracker


paulfalgout:
## Purpose

Track final Marionette v5.0.0 release readiness after v5.0.0-beta.1.

Use this as the parent issue for v5.0.0 final release work. Implementation issues will be filed separately and linked back here as they are created.

### Beta blockers

- [ ] All v5.0.0-beta.1 blockers in the beta tracker are complete.

### Phase 5 - Carry-over v4 bugs

- [ ] Plan issue 40: `uiRegEx` `*` to `+`. Future implementation issue: TBD.
- [ ] Plan issue 41: `_initialEvents` rebinds collection on swap. Future implementation issue: TBD.
- [ ] Plan issue 42: `triggerMethod` cache leak. Future implementation issue: TBD.
- [ ] Plan issue 43: `getRegion` / `getRegions` / `hasRegion` no longer auto-render. Future implementation issue: TBD.
- [ ] Plan issue 44: Trigger `remove:region` on direct region destroy. Future implementation issue: TBD.

### Phase 6 - Pre-1.0 polish

- [ ] Plan issue 45: `MarionetteError` via `class extends Error` and `Error.cause`. Future implementation issue: TBD.
- [ ] Plan issue 46: `_rdInternal` flag to `Symbol`. Future implementation issue: TBD.
- [ ] Plan issue 47: Resolve `Application` lifecycle decision by shipping the chosen implementation or explicitly deferring it. Future implementation issue: TBD.
- [ ] Plan issue 48: Resolve Radio policy hooks by shipping the chosen implementation or explicitly deferring it. Implementation issue: #82.
- [ ] Plan issue 49: Decide on `triggerMethod` exposure via Radio top-level forwarder. Future implementation issue: TBD.
- [ ] Plan issue 50: Resolve DEV_MODE warnings mechanism by shipping the chosen implementation or explicitly deferring it. Future implementation issue: TBD.
- [ ] Plan issue 51: Ship stable `index.d.ts`, shim `.d.ts`, and jQuery DomApi `.d.ts`. Future implementation issue: TBD.

### Final release gates

- [ ] Bundle-size CI is green and tracking real numbers. Implementation issue: #73.
- [ ] Stable type declarations are shipped.
- [ ] At least one representative existing Marionette application or fixture validates the migration path on a beta tag with a documented diff or fixture log.

## Public behavior boundary

[...]

PR #320, opened 2026-08-31
docs(features): document v5 feature flag contract

paulfalgout:
Related #41

## What

- Reframe feature flags as one module-global registry per loaded Marionette copy and remove the v3/v4 breaking-minor-release rationale.
- Document the exact built-in defaults, API validation and return behavior, bootstrap/redelegation timing, custom-name compatibility, and inert `DEV_MODE` behavior.
- Add a packed executable documentation fixture that verifies the registry contract and restores the built-in states it changes.

## Why

The existing page still taught the historical v3/v4 rollout rationale and left important v5 behavior implicit. Coding agents and maintainers need a precise distinction between Marionette global flags, application-owned configuration, and future lifecycle-aware composition APIs.

## Scope

This changes feature-flag documentation and its executable package fixture only. Runtime code, exports, generated distributions, Application, State, and extension-hook APIs are intentionally unchanged. Arbitrary non-empty string flag names remain supported for v4 compatibility, but are not presented as the canonical home for new application configuration.

## Deployment Impact

No website, npm package, tag, or release is published by this PR. No application redeploy is required; the change takes effect when documentation is later published through the normal release process. Runtime artifacts have no code delta.

## Testing

- `npm test -- test/unit/runtime/features.spec.js test/unit/view.triggers.spec.js test/unit/mixins/view.spec.js test/unit/mixins/view-composition.spec.js` — 4 files, 75 tests passed.
- `npm run docs:check` — 17 executable examples across 12 validators; 41 pages built; 42 HTML files and 386 links validated.
- `npm run test:fixtures` — all packed-package fixtures passed, including the new feature-flag contract.
- `npm run lint:ci` — passed with zero warnings.

The fixture covers exact state assignment/return, boolean reads, invalid and unknown names, custom string compatibility, the three built-in values used by the documented bootstrap example, and built-in-state cleanup.

---
## Summary by cubic
Rewrites the feature-flag documentation to describe the v5 module-global registry contract, and adds an executable fixture that verifies it. Addresses #41 by distinguishing Marionette flags from application-owned configuration.

- Documents exact defaults, `isEnabled` truthiness behavior, and `setEnabled` return/validation, including error code `MN0027` for invalid names.
- Clarifies bootstrap timing, `delegateEvents()` refresh for late changes, and that `DEV_MODE` is inert in v5.
- Notes arbitrary string names remain accepted for v4 compatibility but are not recommended for new app configuration.
- Adds `test/fixtures/docs-feature-flags` fixture that validates the documented behavior and restores built-in registry states.
- No runtime code or npm artifacts change; no release or redeploy is involved.

**Migration**

- Remove v4 `DEV_MODE` configuration calls; the docs no longer present custom flags as the canonical home for application config.

<sup>Written for commit 8f48b095763b130e3eb3468d1996c8889727f354. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/320?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
Claude exact-head review for `8f48b095763b130e3eb3468d1996c8889727f354`: I sent the complete feature-doc and packed-fixture diff, v5 runtime evidence, v4 historical evidence, consumer findings, and exact validation results. Claude found no hard correctness blocker and approved the push. It initially asked for proof of setup timing and `DEV_MODE` history; I verified View/CollectionView construction and explicit redelegation paths in source, verified v4 `DEV_MODE` deprecation-warning behavior from the audit tag and current v5 removal ledger, clarified module-global and redelegation wording, and added blank/non-string validation coverage. An independent subagent also found no blocker. Exact-head local validation passed: 75 focused tests, `npm run docs:check`, `npm run test:fixtures`, `npm run lint:ci`, and `git diff --check`.

paulfalgout:
**Approved** at `8f48b095` (the SHA I fetched). Documentation plus one packed fixture; no runtime or `dist/` changes.

## The fixture handles the one hazard this page has

`FEATURES` is a module-global registry, and fixtures share a process — so an example that flips built-in flags and walks away poisons whatever runs next. This one restores them in a `finally` **and** asserts the restoration succeeded rather than trusting it. Given that the documented example exists precisely to demonstrate mutating global state, that is the difference between a safe fixture and a slow-burning cross-fixture failure.

The contract it pins matches current master rather than the earlier alpha shape: custom string names supported, exact-state return from `setEnabled`, and MN0027 raised only for non-strings and blank strings. `isEnabled('  ')` returning `false` while `setEnabled('  ', ...)` throws is a deliberate asymmetry — queries stay total, mutations are validated — and both halves are asserted.

`DEV_MODE` is worth documenting explicitly, and the claim checks out: there is no `DEV_MODE` anywhere in `runtime/`, `modules/`, `mixins/`, or `index.js`. Since v3/v4 carried the concept, "calling `setEnabled('DEV_MODE', true)` is inert" is exactly what a migrating reader needs — it will appear to succeed, because arbitrary names are accepted, and then do nothing.

## This closes the concern I raised on #308

On #308 I noted that removing the own-key check made `__proto__: null` on the `FEATURES` literal load-bearing on its own, with nothing obviously protecting it. There is something: `features.spec.js` asserts, for each of `constructor`, `toString`, `__proto__`, and `hasOwnProperty`:

```js
expect(Object.hasOwn(FEATURES, name)).to.be.true;
```

The `__proto__` case is the one that carries the weight. On a normal object, `FEATURES['__proto__'] = true` attempts a prototype assignment, silently no-ops for a non-object value, and creates **no own property** — so that assertion fails. It can only pass on a null-prototype registry. Deleting `__proto__: null` therefore breaks the suite rather than silently reintroducing the collision. Concern resolved; I was wrong that it was unguarded.

## Optional

Line 63 says `setEnabled` accepts any non-empty string, which is accurate and predicts prototype-property names correctly by implication. Worth considering one sentence making it explicit, though: `constructor`, `toString`, and `__proto__` are precisely the names a careful reader would assume are unsafe or special-cased, and their being ordinary first-class flags is a deliberate design property rather than an accident. Cheap to state, and it saves the reader an experiment.

Nothing blocking.

PR #321, opened 2026-08-31
fix(view): no-op terminal UI binding

paulfalgout:
Related #41

## What

- Make public `View#bindUIElements()` return the view without rebinding once destruction has started or completed.
- Apply the same terminal-state no-op to direct `Behavior#bindUIElements()` calls while preserving normal unbinding and `getUI()` diagnostics.
- Add focused lifecycle coverage and document the v5 contract for views, collection views, and attached behaviors.

## Why

Async cleanup can race with destruction. Rebinding UI after the terminal lifecycle starts cannot restore a usable view and previously required consumers to scatter `isDestroyed()` checks around otherwise harmless cleanup paths. The v5 contract now makes terminal rebinding deterministic and chainable while retaining strict errors for genuinely unbound `getUI()` access.

Marionette v3.5.1 and v4.1.3 had no terminal guard here, but their implementation did not establish a useful post-destroy contract. The new behavior is intentionally narrow: it suppresses only base UI rebinding after destruction begins, keeps unbinding active, and does not hide custom override side effects that do not delegate to the base method.

## Scope

- Global behavior for Marionette View, CollectionView, and attached Behavior UI binding.
- Distribution bundles and source maps are rebuilt from the source change.
- app-frontend has one ordinary `getUI()` consumer and no direct bind/unbind calls; no migration is required.
- marionette.toolkit has no `bindUIElements()`, `unbindUIElements()`, or `getUI()` consumer.
- Event delegation, direct Behavior destruction with a live host, and the existing `getUI()` error contract are intentionally unchanged.

## Deployment Impact

No deployment or publication is performed by this PR. Normal package release consumers will receive the behavior through the rebuilt Marionette bundles. No application-specific redeploy is required outside a future normal dependency update.

## Testing

- `npx vitest run test/unit/destroyed-bind-ui-elements.spec.js test/unit/view.lifecycle.spec.js test/unit/behavior.spec.js test/unit/collection-view.spec.js` — 87 tests passed.
- `npm test` — 94 files and 1,718 tests passed.
- `npm run test:source` — passed.
- `npm run lint:ci` — passed.
- `npm run docs:check` — 17 executable examples, 12 validators, 41 pages, and 386 links passed.
- `npm run check:dist` — passed after rebuilding all tracked distribution artifacts.
- `git diff --check marionettejs/master...HEAD` — passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented UI binding and resolution while a View or CollectionView is being destroyed or has been destroyed.
  * Ensured attached and retained Behaviors also skip UI binding during host destruction.
  * Preserved cleanup through `unbindUIElements()` and existing unbound UI diagnostics.
  * Confirmed UI binding continues to work for active views and after a failed destruction attempt.

* **Documentation**
  * Updated lifecycle, Behavior, and v4-to-v5 migration guidance for terminal UI-binding behavior.


paulfalgout:
**Approved** at `4b52ce63` (the SHA I fetched). +19 B aggregate (62,963 B against the 75,000 B ceiling, +0.03%), nowhere near the growth threshold; `dist/` and its maps regenerated from the source change.

## The guarded and unguarded directions are split correctly

The valuable half of this is what was **not** guarded. `unbindUIElements` stays open and is invoked from `destroy()` at `mixins/view.js:182`, immediately before `_undelegateViewEvents()` and `Dom.detachEl`. Closing the rebinding direction while leaving teardown's own unbinding intact is the whole contract, and getting that backwards would have been an easy and very confusing mistake.

I checked whether any teardown path reaches `bindUIElements`, since that is what forced #302's WeakMap: it does not. The callers are `setElement` on both classes (each already guarded), the render path at `view-region.js:855`, and the behavior fan-out reached from `View#bindUIElements` itself. A plain flag check is sufficient here.

The Behavior guard keys on `this.view` rather than the Behavior's own state, consistent with #318, and that dereference is safe unconditionally since `this.view` is assigned as the Behavior constructor's first statement and never deleted.

## Non-blocking: the `delegateEvents` reorder makes it the lone outlier

```diff
-    if (this._isDestroyed || this._isDestroying) { return this; }
+    if (this._isDestroying || this._isDestroyed) { return this; }
```

Semantically identical — both operands are plain boolean reads with no side effects. But it moves `delegateEvents` **away** from the file's prevailing order rather than toward it. In `mixins/view.js` after this PR:

- line 124 (`delegateEvents`, changed here) — `_isDestroying || _isDestroyed`
- lines 135, 145, 167, and 214 — `_isDestroyed || _isDestroying`

Line 214 is this PR's own new `bindUIElements` guard, so the commit adds a guard in one order and flips an existing one to the other. Four against one either way; whichever is preferred, having a single dissenter invites exactly this kind of churn again later. Not worth a push on its own — worth folding into the next touch of this file.

Nothing blocking.

paulfalgout:
Claude exact-head review for `4b52ce639da751ddaefde3b5ed6408344f67b038`: I sent the complete runtime, tests, documentation, and generated-distribution diff together with v3/v4 history, app-frontend and marionette.toolkit consumer evidence, and exact validation results. Claude found no hard correctness blocker.

Its sole requested cleanup was to align the View terminal guard ordering with Behavior and the existing terminal delegation guards. I applied that change, rebuilt every tracked distribution artifact, reran the 12 focused terminal UI tests plus `npm run check:dist` and `git diff --check`, then submitted the exact resulting delta for re-review. Claude approved exact head `4b52ce63` with no blockers and no missing tests.

What changed because of the review: only the commutative guard ordering was normalized; the lifecycle contract and scope did not change. The prior full content passed 87 focused tests, 94 files / 1,718 full-suite tests, `test:source`, `lint:ci`, `docs:check`, `check:dist`, and diff validation.

PR #323, opened 2026-08-31
fix(region): preserve terminal detach ownership

paulfalgout:
## Linked issue

- Related #131

## What

- Make Region#detachView return undefined once Region destruction begins.
- Preserve the current View for authorized Region teardown, including replaceElement restoration and failed-destroy retry.
- Document the terminal detach contract and regenerate distributable artifacts.

- A before:destroy callback could silently transfer a live View out of Region-owned teardown, leaving ownership dependent on callback timing and private implementation details.

## Public behavior contract

- Canonical behavior after this PR: Region#detachView returns undefined without transferring its current View once destruction begins. While a destroying Region remains registered, View#detachChildView delegates to the same no-op.
- Superseded behavior removed: detachView no longer clears currentView or detaches DOM from within Region teardown.
- Compatibility exception and removal condition: none. Live Region detach behavior and private authorized teardown are unchanged.

## Runtime cost

- Bundle: +21 B aggregate Brotli-11; 62,984 B against the 75,000 B ceiling; no artifact exceeded the 1% approval threshold.
- Startup/hot path: no startup path changes; hosted timing produced no final-head threshold warning.
- Allocations/retention: no new allocation or retained-resource proxy.

## Validation

Commands:

- npm run lint:ci — passed.
- npm test — 94 files and 1,721 tests passed.
- npm test -- test/unit/region-lifecycle.spec.js — 23 tests passed at the final head.
- npm run docs:check — 17 executable examples, 41 pages, and 386 links passed at the final head.
- npm run check:dist — passed after generated artifacts were committed.
- git diff --check — passed.

## Package and browser impact

- Entry points or install fixtures affected: root marionette entry point only through Region behavior; generated main distributions updated. Backbone and jquery-dom-api subpaths are unchanged.
- Browser contracts affected: terminal Region ownership and DOM teardown only; live detach behavior is unchanged.
- Production/dev/test module-graph impact: no production graph change and no new module or dependency.

## Rollback or deprecation

- Revert the guard, tests, migration entry, changelog entry, and generated distributions together if a verified active consumer requires callback-time ownership transfer. No compatibility alias or dual path is retained.

## Review notes

- Exact head c95cf671 has bounded Claude approval, exact-head Cubic success, CodeRabbit no-actionable-comments review, all CI green, clean mergeability, and zero unresolved review threads.
- Direct CollectionView child detach, Application lifecycle, and other #131 transitions remain separate slices.
- No deployment, website publication, npm publication, tag, GitHub release, or repository policy change is included.


paulfalgout:
**Approved** at `cdf660b4` (the SHA I fetched). +21 B aggregate (62,984 B against the 75,000 B ceiling); `dist/` regenerated.

## This closes a window I verified past on #301

On #301 I checked the post-destroy detach contract and confirmed `detachView()` returns `undefined` because `_empty` clears `currentView`. That was correct for the **destroyed** state and incomplete for the one before it: during `before:destroy`, `_isDestroying` is set but `currentView` is still populated, so the unguarded method would happily detach a live View and clear ownership — letting it escape the teardown that was already underway.

I later raised exactly this window as a cross-guard inconsistency on #307 without connecting it back to detach. Worth recording, since the pattern is now three-for-three: `setElement`, entity delegation, and detach all had the same gap.

## The guard leaves the authorized path alone without needing a token

#302 needed the WeakMap because teardown routed through the guarded methods. Traced this one: `destroy → reset → empty → emptyRegion() → _empty(view, true)`, and `_empty` delegates to the **private** `_detachView`. The public `detachView` is reachable only from application code and from `View#detachChildView`. So guarding the public entry point costs nothing on the teardown path, and reusing the existing `canMutateRegion` helper keeps it consistent with the `show`/`empty`/`reset` guards rather than introducing a fourth spelling.

## The retry test is the one that earns its place

`preserves ownership when terminal detach precedes a failed destroy` composes two contracts that were each individually correct: terminal detach, and the retryable-destroy boundary from #224. The defect lived only in their interaction — a `before:destroy` callback detaches, `destroy` then throws, `_isDestroying` is cleared for retry, and the retry finds an empty Region while a live View is loose with nothing owning it. Neither contract's own tests would have caught that.

The `replaceElement` case is the right second choice too, since that path has element restoration to get wrong on top of ownership.

## Minor scope note

`View#detachChildView(name)` delegates to the Region's public `detachView`, so it now returns `undefined` when that Region is destroying or destroyed. The window is narrow — it requires a handler on the Region's own `before:destroy` — but the Scope section names CollectionView child detach as out of slice without mentioning that Region-backed child detach changes transitively. Worth a line in the migration entry rather than a code change.

Nothing blocking.

paulfalgout:
**Approval stands** at `24b5b67b`. One migration-row sentence, docs only, so no `dist/` regeneration is involved.

This is the note I raised, placed where I would have wanted it. The wording is better than a bare statement of the outcome:

> `View#detachChildView` has the same terminal result **because it delegates to that Region method**

Naming the mechanism rather than just the behavior lets a reader generalize — the transitive result follows for any API that routes through the guarded Region method, so the next one does not need its own migration row to be predictable. That is the difference between documenting a fact and documenting a rule.

paulfalgout:
**Approval stands** at `c95cf671`, and this push corrects my note rather than just adopting it.

I wrote that `View#detachChildView` returns `undefined` "when that Region is destroying **or destroyed**." The second half is wrong, and the new wording gets it right:

> While a destroying Region **remains registered**, its owning View's `detachChildView` delegates to the same terminal no-op.

Once destruction completes, `Region#destroy` calls `this._parentView._removeReferences(this._name)` (`view-region.js:536`), which deletes the entry from `_regions`. From that point `getRegionForChild` resolves nothing and `getRequiredRegion` throws **MN0020** — a different outcome entirely from a silent `undefined`. So the transitive no-op is specific to the destroying window, when the Region is mid-teardown but still reachable by name.

That distinction is exactly the sort a caller would get wrong from my phrasing: they would write a `catch` for a diagnostic that never arrives during destroying, or none at all and be surprised after. The added test pins the half that is easy to lose — `does not detach through an owner while its Region is destroying` covers the narrow window, and the MN0020 path after full destruction was already covered by the named-operation contract from #306.

Issue #329, opened 2026-08-31
v5: close out the production authorship audit


paulfalgout:
## Current status — September 9, 2026

Repository readiness work is merged in [#500](https://github.com/marionettejs/marionette/pull/500) at [master `70eaeaf5`](https://github.com/marionettejs/marionette/commit/70eaeaf582a3ae062b5ff48650cfda12cfa3bf9f). It is available in source and verified tarballs; no new npm version or tag was published. The registry remains beta.1 on `latest`.

The 46 historical paths are reconciled to 62 current non-declaration TypeScript files, including type-only support modules. Current Radio/runtime boundaries supersede older intermediate decisions. No unresolved align item remains. The bounded authorship gate is satisfied; #147 owns the future release certificate.

## Scope and stopping rule
The public inventory in this issue's historical comments covers production paths changed since fork revision `662996b2`, with `align`, `retain`, or `separate behavior/API issue` dispositions. Historical Marionette and Toolkit style is evidence, not a requirement to restore obsolete APIs or dependencies.

## Delivered
- [x] Initial production inventory and disposition/rationale for each group.
- [x] Separate focused decisions for public/helper/Radio/Region behavior (#334–#339) and bounded source polish (#340).
- [x] Core source consolidated under `src/` (#402/#409), distributions removed from Git (#407), internal naming/comments clarified (#416), and public typing completed (#429).
- [x] Behavior initialization correction reviewed separately from style (#430); docs voice aligned separately (#431).

## Remaining acceptance criteria
- [x] Update the public inventory's historical path references to the current `src/` and optional-package tree, and account for production additions/moves since the original inventory. The original comparison and rationale remain historical evidence.
- [x] Record any specific unresolved `align` item with an owning issue/PR, or explicitly record none. Do not restart an open-ended cosmetic audit.
- [x] Link the completed applicable validation from the cleanup PRs and the final release evidence in #147; do not represent a candidate's green checks as a future release-commit run.
- [x] Record that the bounded authorship gate in ROADMAP.md is satisfied once the inventory is current and every corrective item is accounted for.

Retain technically justified differences for correctness, concurrency, performance, packaging or dependency removal. No additional render/update guards or abstraction vocabulary should be introduced merely for defensive completeness. Future changes need a concrete finding and proportionate validation.

paulfalgout:
## Inventory entry: Application lifecycle and ownership

| Field | Assessment |
| --- | --- |
[...]

PR #333, opened 2026-08-31
refactor(application): clarify lifecycle operation facts

paulfalgout:
Related #329

## What

- Rename private lifecycle-operation facts to describe the committed condition: `didStop`, `didReachTargetState`, and `stopDeferred`.
- Replace vague callback-local `current` names with `nextOperation` and document why only the current operation may commit or restore state.
- Remove two trivial child-Application Map wrappers in favor of direct `Map` operations, then regenerate distribution artifacts.

## Why

The v5 Application lifecycle is necessarily more complex than v4 because it is Marionette's single asynchronous lifecycle surface and must settle repeated, overlapping, superseded, and reentrant operations deterministically. This PR does not restyle that machinery merely to imitate historical source. It makes only the naming and helper changes that improve the existing implementation without creating new prototype extension seams or weakening its invariants.

This is the focused Application entry in the production-source authorship audit from #329. Historical style remains evidence rather than an automatic requirement.

## Scope

Impacted: private implementation naming and direct control flow in `modules/application.js`, plus generated distribution artifacts.

Not changed: public API, prototype surface, lifecycle event signatures, Promise behavior, state transitions, child ownership behavior, documentation, or package exports.

## Deployment Impact

No deployment or consumer migration is required. The normal package build includes the regenerated distribution artifacts; no website, npm package, tag, or release is published by this PR.

## Testing

- `npm run lint:ci -- --no-cache`
- `npm run test:source -- --run test/unit/application-lifecycle.spec.js test/unit/application-ownership.spec.js test/unit/object-application-composition.spec.js`
- `npm run build`
- `npm run coverage` — 96 files, 1,818 tests, 100% statements/branches/functions/lines
- `npm run test:dist`
- `npm run docs:check`
- `npm run size` — 70.48 kB cumulative against the 75.00 kB ceiling
- `git diff --check`

The current head has the exact same repository tree as the previously validated `b605b6382a085b0776a043036851b1c00d9a1611`; exact-head CI is rerunning after the intervening style experiment was fully backed out.


paulfalgout:
Superseded by #354, which reapplies the relevant Application cleanup to current master after the child lifecycle, root View/Region ownership, and Application State work. The replacement also includes the remaining production-runtime audit, roadmap reconciliation, and bundle tree-shaking regression coverage.

PR #343, opened 2026-08-31
feat(region): require explicit string names

### LINKED ISSUE #339: v5: narrow named Region APIs to non-empty strings
Related #329

## What

Make the named Region APIs use one explicit name contract before the v5 API freezes:

- accept non-empty string names;
- continue treating ordinary string collisions such as `constructor`, `toString`, and `__proto__` as valid own names;
- reject empty, object-coercible, array, and Symbol names with one stable diagnostic at registration and required/optional named operations;
- remove tests and implementation branches that currently make JavaScript property-key coercion part of the v5 contract.

The affected surface is `addRegion`, `addRegions` object keys, `removeRegion`, `hasRegion`, `getRegion`, `showChildView`, `detachChildView`, and `getChildView`.

## Why

Marionette's docs and real consumers use string Region names. V3/v4 inherited wider behavior from ordinary object property access; it was not a designed Region-name API. Current v5 tests now deliberately preserve object and array coercion, tolerate a name with no string coercion, and separately repair empty-string ownership. That turns incidental JavaScript behavior into framework surface and makes lookup/diagnostic code harder to inspect.

Application's new ownership API already requires a non-empty string child name. Using the same rule for named Regions gives agents and users one predictable ownership vocabulary without removing valid prototype-collision strings.

## Evidence

- V3/v4 documentation and current documentation show string names.
- app-frontend and marionette.toolkit use string names.
- Targeted public GitHub searches found no Marionette consumer using `addRegion(Symbol...)`, `getRegion(Symbol...)`, `addRegion([...)`, or object-valued names; the only `getRegion([...)` result was Marionette's own v5 coercion test.
- Declarative Region maps already enumerate string keys; Symbol entries are not a coherent supported path.
- `modules/view-region.js` currently needs `getOwnRegion` exception swallowing and conditional diagnostic-label logic partly to preserve these unevidenced shapes.
- Application rejects empty and non-string child names with `MN0031`.

## Contract

This is a deliberate v5 break from incidental v4 property coercion, not a compatibility alias. Empty string is rejected even though #324 fixed its cleanup behavior; #324's broader ownership invariants remain intact.

Keep lifecycle-safe no-ops unchanged: once destruction begins, mutating operations that already no-op must not start validating caller arguments.

## Validation

Add focused source and packed-package coverage for valid collision names, invalid name types, lifecycle ordering, and unchanged ownership cleanup. Update current docs, migration guidance, diagnostics, generated distributions, size evidence, and representative app-frontend/toolkit migration checks.

paulfalgout:
Closes #339

## What

- Require named Region APIs to receive a non-empty string while preserving valid collision names such as `constructor`, `toString`, and `__proto__`.
- Use one stable diagnostic for invalid Region names across registration, lookup, display, detach, and removal operations.
- Remove object/property-key coercion branches and document the deliberate v5 contract.

## Why

V3/v4 inherited non-string name behavior from JavaScript property coercion rather than a designed Marionette API. Verified consumers use string names, and Application ownership already uses the same explicit non-empty-string contract. Narrowing the surface now makes ownership queries deterministic and inspectable without changing lifecycle-safe no-ops after destruction begins.

## Scope

This affects the Region name accepted by `addRegion`, `addRegions`, `removeRegion`, `hasRegion`, `getRegion`, `showChildView`, `detachChildView`, and `getChildView`. Region ownership, lifecycle ordering, and valid prototype-collision names remain unchanged. No compatibility alias or coercion fallback is added.

## Deployment Impact

No deployment or forms redeploy is required. This is library runtime, generated distribution, documentation, diagnostics, and test coverage only.

## Testing

- `npm run lint`
- Full unit suite with 100% coverage
- Source and generated-distribution validation
- Packed-package smoke validation on supported fixtures
- Documentation and browser fixture validation
- Bundle-size validation

Coverage includes valid collision names, rejected empty/non-string names, lifecycle-safe no-ops, and unchanged Region cleanup behavior.

---
## Summary by cubic
Requires named View Region operations to accept only non-empty string names, removing the previous property-key coercion fallback. This makes ownership lookups deterministic and adds a stable `MN0032` error for invalid names across registration, lookup, display, detach, and removal.

**Migration**
- Pass the intended string name directly; arrays, objects, symbols, and empty strings now throw instead of being coerced or ignored.
- Prototype-collision names like `constructor`, `toString`, and `__proto__` remain valid when explicitly registered.

<sup>Written for commit 7cc28124976ba01e739aa17763671e1d787cce53. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/343?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
Claude CLI exact-head review: **APPROVE**

Reviewed commit `7cc28124976ba01e739aa17763671e1d787cce53` in a bounded design pass plus a small follow-up delta.

The first pass supported the explicit non-empty primitive-string contract and identified one valid side-effect risk: invalid child-operation names rendered the parent before throwing `MN0032`. This head validates the name before rendering, removes one redundant downstream assertion, and directly proves the View stays unrendered.

The follow-up found no blocker. I independently confirmed the only two `getRequiredRegion` call sites are already preceded by `assertRegionName` through `getRegionForChild` or `getOwnRegion`. Required CI, Cubic, and the local 1,827-test suite are green; there are zero unresolved threads.

Issue #191, opened 2026-08-28
v5: verify explicit state-source composition and ownership


paulfalgout:
## Current status — September 9, 2026

Repository readiness work is merged in [#500](https://github.com/marionettejs/marionette/pull/500) at [master `70eaeaf5`](https://github.com/marionettejs/marionette/commit/70eaeaf582a3ae062b5ff48650cfda12cfa3bf9f). It is available in source and verified tarballs; no new npm version or tag was published. The registry remains beta.1 on `latest`.

The provider/state acceptance mapping is integrated. [Current timing and direct collection evidence](https://github.com/marionettejs/marionette/actions/runs/34333478421) covers shared sources, owned disposal and collection after owner destruction with sources still alive. Real-world tasks, agent outcomes and final stable-candidate evidence remain open.

## Canonical contract
State is an owner-to-source relationship. The former concrete core `State` class and implicit `new State(definition)` path were removed by the accepted design; retaining them is no longer a requirement.

- `state` supplies an exact borrowed source; a supplied function is also a source, not an inferred factory.
- `createState(options)` explicitly creates an owned source. Without supplied state or a custom factory, the first `getState()` lazily creates an ordinary non-observable object.
- `getState()` returns the exact source. Plain objects are valid and are never implicitly made observable.
- StateApi selection is explicit and separate from model/collection DataApi. `subscribe` returns cleanup; optional `disposeOwned` runs for factory-owned sources after subscriptions are released.
- The source owns mutation and event semantics. Marionette forwards native payloads and owns only its registrations and explicitly owned disposal.
- View/CollectionView render and Application stop/restart preserve state. Region does not compose state. Unused owners allocate no state record or subscription.
- Optional `@mnjs/data` Model and provider adapters supply observable choices; they do not restore a concrete core State class.

## Delivered
- [x] Explicit borrowed/factory-owned composition, exact source identity, lazy plain-object default, and adapter selection.
- [x] Successful subscription cleanup independent of public `off()`, with supported ordinary ownership/destruction behavior. Historical constructor/partial-registration rollback claims from #406/#430 are superseded by #470; synchronous adapter/callback failures abort without rollback or attempt-all cleanup.
- [x] Borrowed-source survival, subscription-before-owned-disposal ordering, shared-source observation and rerender persistence.
- [x] Public source/factory/adapter declarations and packed-package composition checks (#429).
- [x] Updated state and migration documentation, including plain-object and provider examples (#431).

[...]

Issue #190, opened 2026-08-28
v5: verify Application lifecycle and ownership


paulfalgout:
## Current status — September 9, 2026

Repository readiness work is merged in [#500](https://github.com/marionettejs/marionette/pull/500) at [master `70eaeaf5`](https://github.com/marionettejs/marionette/commit/70eaeaf582a3ae062b5ff48650cfda12cfa3bf9f). It is available in source and verified tarballs; no new npm version or tag was published. The registry remains beta.1 on `latest`.

The Application model and acceptance mapping are integrated. [Current timing and direct collection evidence](https://github.com/marionettejs/marionette/actions/runs/34333478421) covers cancellation, restart, hosting and teardown. Reference-application/corpus experience and the final stable candidate remain open.

## Canonical contract
Application is the first-class non-renderable root or nested lifecycle scope. It is independent of MnObject, has no element/render method, and never becomes a Region-renderable object.

`start`, `stop`, `restart`, and `destroy` return `Promise<boolean>` with the readiness, invalidation, rejection and terminal-destruction rules in the public guide. Child registration is synchronous ownership registration; it does not implicitly start a newly registered child. Subsequent owner operations propagate in order. Removing a child releases ownership without secretly stopping or destroying it.

Application may coordinate a root View through a constructed/owned or supplied/borrowed Region. Region owns View mounting. External replacement clears stale View ownership without implicitly stopping Application or allowing later teardown to empty an unrelated replacement View. State-source composition is #191. General resource registration is the evidence-dependent candidate #136; canonical cleanup uses lifecycle callbacks.

## Delivered
- [x] Root/nested lifecycle tables, Promise returns, readiness cancellation and documented repeated/overlapping operations.
- [x] Child ownership, named lookup, ordered lifecycle propagation, removal and teardown without Toolkit lifecycle-control flags.
- [x] Root View hosting and constructed-versus-borrowed Region ownership.
- [x] Public hierarchy access and explicit Application-to-root-View communication guidance.
- [x] Accepted adopted child-stop corrections (#413), public typing (#429), and refreshed examples (#431). Historical construction rollback from #412 is superseded by the supported synchronous failure contract in #470; it is not a current guarantee.

[...]

Issue #140, opened 2026-08-26
v5: publish a compact agent-oriented framework reference


paulfalgout:
## Current status — September 9, 2026

Repository readiness work is merged in [#500](https://github.com/marionettejs/marionette/pull/500) at [master `70eaeaf5`](https://github.com/marionettejs/marionette/commit/70eaeaf582a3ae062b5ff48650cfda12cfa3bf9f). It is available in source and verified tarballs; no new npm version or tag was published. The registry remains beta.1 on `latest`.

The generated compact reference covers the stated static topics, diagnostics, canonical examples and package surfaces. It ships with the source documentation and is drift-checked. Public agent-benchmark usefulness remains unmeasured.

## Goal
Publish a token-dense framework reference generated or checked against public contracts, types, examples, and diagnostic rules.

## Agent-development failure addressed
Agents need a compact source of truth but must not be taught a second Marionette API.

## Public behavior contract
The reference is documentation only and mirrors canonical public behavior.

## Runtime cost category
Documentation/static generation only.

## Acceptance criteria
- [x] Cover lifecycle, plain-class versus optional-MnObject versus Application roles, nested Application ownership and hosting, state sources versus domain data, Regions, Behaviors, communication, lifecycle-based resource cleanup, errors, and package surfaces.
- [x] Link stable diagnostic codes and canonical examples.
- [x] Generate or drift-check public API facts.
- [ ] Verify usefulness through the public agent benchmark.
- [x] Exclude private or unpublished context.

## Rollback or deprecation condition
If generated content diverges from human docs, fix the shared source or generator instead of maintaining two truths.

## Reconciled scope

The shared public API manifest is locally implemented under #137 and awaits integration. Use it with canonical guide examples and diagnostic codes; do not equate the friendly README with the complete compact reference. Source-owned state and ordinary lifecycle cleanup replace references to the removed concrete State class or speculative resource helpers.

PR #355, opened 2026-09-01
docs(v5): make application and state guidance executable

paulfalgout:
## Linked issue

- Related #190
- Related #191
- Related #140

## What

- Document explicit child Application ownership and direct Application/root-View communication.
- Add executable View, Application, and Behavior State ownership examples.
- Validate all five examples against the freshly packed package, including lifecycle order, stale-start cancellation, and teardown.

## Why

The v5 runtime contracts are implemented, but the remaining stable-release work requires canonical patterns that agents and humans can execute instead of inferring ownership and lifecycle behavior from source.

- Makes ownership direction, lifecycle propagation, State placement, cancellation, and teardown observable through copyable public-API examples.
- Prevents agents from treating child registration as a dormant service registry or finding an Application through DOM/private ownership state.

## Public behavior contract

- Canonical behavior after this PR: construct child Applications explicitly, register only owned children, communicate across the direct root boundary with View options/methods downward and semantic View events upward, and place local State on the owner whose lifetime it follows.
- Superseded behavior removed, if any: none; this documents existing v5 behavior.
- Compatibility exception and removal condition, if any: none.

## Runtime cost

- Bundle: 0 B across all production artifacts and canonical consumer scenarios.
- Startup/hot path: no production change; hosted timing remained reporting-only with no warning.
- Allocations/retention: no production change; the packed fixture verifies owner-driven teardown.

## Scope

This changes Application and State documentation plus one packed-package fixture. It does not change production runtime behavior, package exports, generated distributions, performance budgets, or the benchmark corpus.

## Deployment Impact

No application redeploy or package publication is required. Documentation publication remains governed by the existing post-beta release process.

## Validation

Commands:

```sh
npm run docs:check
npm run lint:ci
npm run test:source
npm run test:fixtures
```

All four commands passed locally. PR CI also passed documentation build, Node 24, Node-current advisory, CodeQL, bundle size, package smoke on macOS and Windows, verified-tarball smoke, performance reporting, and promotion dry run.

## Package and browser impact

- Entry points or install fixtures affected: no entrypoint changes; adds one packed-package documentation fixture.
- Browser contracts affected: validates root-View communication and State behavior in JSDOM; no runtime browser behavior changes.
- Production/dev/test module-graph impact: no production graph change; one test-only fixture.

## Rollback or deprecation

Revert the documentation and its fixture together if a referenced v5 contract changes before stable. Do not leave examples that no longer pass the packed-package fixture.

## Review notes

- The packed fixture extracts the exact marked code fences rather than duplicating their implementation.
- Bounded Claude review received the actual diff. It led to explicit AbortSignal guidance, a documented stale-start race, executable private-Behavior versus shared-View State ownership, and teardown assertions for both State instances.
- CodeRabbit and Copilot generated no actionable inline comments.

### DIFF
--- a/docs/marionette.application.md
+++ b/docs/marionette.application.md
@@ -20,6 +20,7 @@ The `Application` `cidPrefix` is `mna`.
 * [Instantiating An Application](#instantiating-an-application)
 * [Application Lifecycle](#application-lifecycle)
 * [Application Ownership](#application-ownership)
+* [Application and root View communication](#application-and-root-view-communication)
 * [Application State](#application-state)
 * [Application Region](#application-region)
 * [Application Region Methods](#application-region-methods)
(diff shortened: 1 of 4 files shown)


paulfalgout:
**Approved** at `b6bfacef` (the SHA I fetched). Every check passes, and `dist/`, `modules/`, `mixins/`, `runtime/`, `utils/`, and `index.js` are all untouched — worth stating explicitly now that the aggregate sits 16 B under the 75,799 B ceiling, where any shipped-byte change becomes a budget question.

## The lifecycle ordering assertion is the valuable part

```js
assert.deepEqual(ownership.lifecycle, [
  'root:before:start:owner',
  'search:before:start:owner',
  'search:start:owner',
  'root:start:owner',
  ...
```

Pinning the **whole sequence** with `deepEqual`, rather than checking each event was called, is what turns this into a contract: any reordering fails it, including reorderings that leave every individual assertion true.

And the sequence encodes something a consumer would otherwise have to guess — and could plausibly guess either way. The parent's `before:start` opens, the child starts **completely** (both its readiness and its completion), and only then does the parent's `start` fire. So a parent Application's `start` event means *"I and my children are up"*, not *"I have begun bringing my children up."* Stop mirrors it exactly, and destroy runs child-then-parent. Those are three separate ordering claims that all had to be decided; having them executable beats a prose paragraph that reads correctly under either interpretation.

## Verified the documented APIs exist

Docs asserting an API surface is exactly where drift hides, so I checked each against master rather than trusting the examples compile:

| API | Defined in |
| --- | --- |
| `getChildApp`, `getParentApp`, `getView` | `modules/application.js` |
| `getState` | `mixins/state.js` |

All four present.

## One thing worth stating so the two records do not look contradictory

#330 recorded that `onStart` / `onStop` / `onDestroy` are **non-awaited completion notifications**. This fixture shows a parent's completion waiting for its child's completion, which reads like a conflict on first pass and is not one: the child Application's startup is part of the parent's own start *work*, not a readiness hook the parent awaits. The parent still does not await its own `onStart` listeners.

Someone reading the Application reference and the roadmap contract side by side will hit that seam. A sentence in the child-ownership section naming it — child startup is owned work, not an awaited hook — would save the reconciliation.

Nothing blocking.

PR #357, opened 2026-09-02
refactor(runtime): remove global feature flags

paulfalgout:
## What

- Remove the module-global feature registry and the `setEnabled` and `isEnabled` exports.
- Preserve canonical behavior through per-View `childViewEventPrefix` and per-trigger `preventDefault` and `stopPropagation` options.
- Update the v5 roadmap and migration contract, retire `MN0027`, and remove obsolete feature documentation and fixtures.

## Why

The remaining flags did not stage future Marionette behavior. They retained older global behavior or stored application-owned values that Marionette never interpreted. No verified app-frontend or Marionette Toolkit consumer uses the global API, and the existing local View and trigger contracts provide explicit ownership without a process-wide registry.

## Scope

This changes the root runtime API, View child-event prefix fallback, trigger DOM-event defaults, generated distributions, diagnostics, documentation, and tests. Trigger behavior remains prevent-and-stop by default, and `childViewEventPrefix` remains disabled by default. Radio and other module-global configuration surfaces are intentionally unchanged.

## Deployment Impact

No application deployment is required beyond a normal Marionette package release. Consumers upgrading to v5 must remove feature imports and calls, configure child event prefixes on their owning Views, configure trigger exceptions per trigger, and move application-owned values to Application State or explicit configuration.

## Testing

- `npm test -- --reporter=dot` — 99 files, 1,910 tests passed.
- `npm run lint:ci`
- `npm run check:diagnostics`
- `npm run docs:check`
- `npm run test:source`
- `npm run test:dist`
- `npm run test:fixtures`
- `npm run test:performance` — 106 tests passed.
- `npm run size` — cumulative Brotli 75.17 kB / 75.80 kB.
- `git diff --check`

---
## Summary by cubic
Removes the module-global feature registry and its `setEnabled`/`isEnabled` exports, which is a breaking API change for code using the global feature API. The remaining flags weren't staging future behavior, so canonical defaults now come only from per-View `childViewEventPrefix` and per-trigger `preventDefault`/`stopPropagation`; trigger defaults stay prevent-and-stop, and child-event prefix stays disabled by default.

**Migration**
- Remove feature imports and calls from application and test code.
- Set `childViewEventPrefix` on the owning View; disabled trigger flags had no global replacement, so set `preventDefault: false` or `stopPropagation: false` per affected trigger.
- Move application-owned feature values to Application State or explicit configuration.
- No deployment beyond a normal Marionette package release is required.

<sup>Written for commit 199756401a7745b2dfb569760d531b580052a344. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/357?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approved** at `bd1301e8` (the SHA I fetched). Nothing failing; a few checks were still running.

**The bundle shrinks 617 B** — aggregate 75,166 B against the 75,799 B ceiling, taking headroom from 16 B to **633 B**. That is the first real relief in a while and materially changes the budget picture I flagged on the last two PRs.

## MN0027 is retired, not deleted — the #309 machinery held on its first real test

`status: "retired"`, joining MN0028 and MN0029 as reserved identities. Worth calling out because the equivalent moment in #308 went the other way: a policy prohibiting deletion already existed then and the codes were deleted anyway. The difference now is that the validator rejects emission of a retired code rather than the policy merely saying so — and the first occasion where that mattered, it produced the right outcome without anyone having to remember the rule.

## The preserved defaults are genuinely preserved

I checked these rather than taking the description's word, since "behavior preserved" is the easiest claim to get subtly wrong when collapsing a conditional:

**Triggers.** The flags shipped enabled, and the enabled path was `triggerDef.preventDefault !== false`. The replacement is that expression verbatim — prevent-and-stop by default, explicit `false` opts out. Character-identical, not merely equivalent.

**`childViewEventPrefix`.** The flag defaulted to `false`, so `defaultPrefix` resolved to `false`, and the new fallback is `false`. Same.

No `isEnabled` or `setEnabled` references remain anywhere in `modules/`, `mixins/`, `runtime/`, `utils/`, or `index.js`.

## What is actually lost, stated precisely

The migration row says trigger prevention "remains configurable per trigger," which is true but understates the change by one step. What disappears is the **opt-in** semantics — `!!triggerDef.preventDefault`, where a trigger had to ask for prevention rather than opt out of it. That path was only reachable by explicitly calling `setEnabled('triggersPreventDefault', false)`, since the flags shipped enabled.

So a consumer loses the ability to globally *invert* the trigger defaults, not the defaults themselves and not per-trigger control. Anyone who did flip those flags has no direct replacement: they now have to set `preventDefault: false` / `stopPropagation: false` on each trigger that wanted the old inverted behavior. That is a real migration task rather than a no-op, and it is the one case where "remains configurable per trigger" could be read as reassurance when it is actually the required work.

Worth a sentence in that row naming the inversion explicitly — the flags' *disabled* state is what has no equivalent.

paulfalgout:
**Approval stands** at `19975640`. Documentation only; no source or `dist/` change, so the −617 B figure is unchanged.

This is the caveat I asked for, and it names the actual work rather than gesturing at it:

> The disabled `triggersPreventDefault` and `triggersStopPropagation` flags globally inverted trigger defaults and **have no global replacement**; set `preventDefault: false` or `stopPropagation: false` on each affected trigger.

"No global replacement" is the part that matters. The surrounding sentence — prevention "remains configurable per trigger" — is true and reads as reassurance, and a consumer who had flipped those flags would have skimmed past it into a silent behavior change. Stating that the inverted default is gone, and that the replacement is per-trigger opt-out, converts that from a surprise into a checklist item.

Pointing `DEV_MODE` at cataloged diagnostics is a useful addition too, since that flag never did anything in v5 and someone migrating would otherwise be looking for its equivalent.

Issue #375, opened 2026-09-02
Make MnObject and Application owned cleanup explicit


paulfalgout:
## Problem

MnObject and Application establish owned Radio and State resources before all fallible construction work completes, but failed construction does not consistently roll them back. A failed instance can remain retained and callable through the module-global Radio channel.

Normal cleanup is also wired through removable public `destroy` listeners:

- `RadioMixin#_initRadio` registers `_destroyRadio` with `this.on('destroy', ...)`.
- `StateMixin#getState` registers `_destroyState` the same way.
- Calling `off()` before `destroy()` removes those internal listeners, leaving Radio requests or owned State alive after the owner is destroyed.

Radio event bindings additionally rely on the later generic `stopListening()` path, so constructor rollback must cover both request/reply ownership and event-listener ownership.

## Required contract

- Failed MnObject and Application construction attempts cleanup for every resource established before the failure.
- Preserve the original construction error even when cleanup also throws.
- Remove Radio event and request registrations from failed instances so the global channel cannot retain or invoke them.
- Destroy owned State and release its owner reference after failed construction.
- Destroy an Application-owned Region as part of the same attempt-all rollback.
- Normal MnObject and Application destruction invokes owned Radio and State cleanup directly; public `off()` cannot disable it.
- Keep cleanup idempotent and preserve the public `before:destroy` and `destroy` lifecycle signatures and settled state timing.
- Do not introduce protected event listeners, a second teardown registry, Radio factories, or Application injection as part of this fix.

## Coverage

- MnObject and Application failures after Radio registration, during State creation, during `initialize()`, and during `stateEvents` setup as applicable.
- Radio events and constant/function requests do not invoke a failed owner.
- A cleanup error does not replace the original construction error or prevent remaining cleanup.
- `off(); destroy()` still removes Radio requests/events and destroys owned State for both MnObject and Application.
- Application-owned Region rollback remains complete.
- Repeated/reentrant destruction does not repeat owned cleanup.

## Release status

Treat this as a v5 release-blocking ownership issue before the first integration candidate. Keep constant `replyOnce` identity and empty `listenTo` ledger cleanup as separate focused changes.

PR #380, opened 2026-09-02
docs(roadmap): define state source ownership

paulfalgout:
Related #191

## What

- Replace the roadmap's native-State ownership model with an explicit owner-to-state-source composition contract.
- Define supplied sources as borrowed and owner-factory results as owned, with `getState()` preserving exact source identity.
- Align Phase 1, TypeScript, integration, allocation, adapter, and stable-release gates with explicit per-owner observation and evidence-gated native State.

## Why

The roadmap still treated Marionette's native `State` mutation and event shape as universal and silently transferred ownership of supplied State instances. That conflicts with the selected boundary in #191: external stores own their native storage and mutation semantics, while Marionette owns only explicit source/factory selection, observation, subscription release, and owned-result disposal.

## Scope

This is a roadmap-only contract clarification. It does not implement the new composition API, change current runtime behavior, rename an adapter subpath, publish a package, or alter repository settings. Native `State` remains an explicitly constructed dependency-free candidate until reference-task and performance evidence decides whether it remains public.

## Deployment Impact

No deployment or package publication is performed. No consumer redeploy is required.

## Testing

No runtime tests were run because only `ROADMAP.md` changed.

---
## Summary by cubic
Revises the roadmap's native `State` ownership model to match the selected boundary in #191: external stores own their storage and mutation semantics, while Marionette owns only source selection, observation, subscription release, and owned-result disposal.

- Replaces the universal `State` contract with an explicit owner-to-state-source composition contract, defining supplied sources as borrowed and owner-factory results as owned.
- `getState()` now preserves exact source identity; plain objects never trigger implicit native `State` construction.
- Aligns Phase 1, TypeScript, integration, allocation, adapter, and stable-release gates with per-owner observation and evidence-gated native `State`.
- Roadmap-only clarification; no runtime behavior, package publication, or consumer redeploy is involved.

<sup>Written for commit 118b47ef2ca3281cefd76a8373df8c0d90345292. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/380?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -120,14 +120,17 @@ budget.
 - Regions own and mount Views. An Application hosted in a Region coordinates a root
   View; the Application itself never becomes a second Region-renderable object
   category.
-- State is a first-class, opt-in object composed into MnObject, View, CollectionView,
-  Behavior, or Application. Owners without State allocate no state store or
-  subscriptions.
-- State is deliberately small and synchronous. It does not implicitly render Views,
-  schedule work, compute derived values, run effects, or persist data.
+- State composition is a first-class, opt-in relationship between MnObject, View,
+  CollectionView, Behavior, or Application and an explicit state source. `getState()`
+  returns that source unchanged; owners without one allocate no state source or
+  subscription.
+- The source owns storage and mutation semantics. Marionette owns source selection,
+  declarative observation, subscription release, and factory-result disposal without
+  implicitly constructing a store from a plain object.
 - Application is Marionette's only asynchronous lifecycle and orchestration surface.
-  View, Region, CollectionView, rendering, templates, Events, Radio, State mutation,
-  and destroy callbacks remain synchronous and never auto-await callback results.
+  View, Region, CollectionView, rendering, templates, Events, Radio, state-source
+  callbacks, and destroy callbacks remain synchronous and never auto-await callback
+  results.
 - Radio retains its module-global channel registry for v5. Applications may use Radio,
   but do not own or isolate its channels.
 - Behavior scope, UI resolution, event delegation, dependencies, and teardown are
@@ -155,8 +158,8 @@ budget.
 - Documentation names canonical patterns and counterexamples.
 - Examples are run in CI or otherwise verified against the shipped package.
 - A compact agent-oriented reference describes lifecycle, ownership, regions,
-  Applications, State, behaviors, communication, and teardown without inventing a
-  separate API.
+  Applications, state-source composition, native `State`, behaviors, communication, and
+  teardown without inventing a separate API.
 - The package ships compact, version-aligned, non-runtime agent material generated
   from the same public metadata: API and lifecycle tables, diagnostics, migration
   guidance, and canonical examples and counterexamples.
@@ -226,11 +229,11 @@ budget.
 
 ## Architecture boundaries
 
-Core owns the essential View, Region, Behavior, Application, CollectionView, State,
-lifecycle, event, and error contracts and preserves MnObject as an optional
-convenience. Additive APIs are justified when they expose information the runtime
-already maintains or make an existing responsibility explicit without adding work to
-unused instances.
+Core owns the essential View, Region, Behavior, Application, CollectionView,
+state-source composition, lifecycle, event, and error contracts and preserves MnObject
+as an optional convenience. Additive APIs are justified when they expose information
+the runtime already maintains or make an existing responsibility explicit without
+adding work to unused instances.
 
 `MnObject` remains an optional minimal convenience for passive, non-renderable,
 evented objects whose lifetime ends at destroy. It has no start, stop, restart, child
@@ -244,10 +247,10 @@ owned child Applications. An Application without a parent is the root of one
 composition tree; independent roots may coexist on a page. Root status is topology,
 not a reason for an Application subtype.
 
-Application, MnObject, View, CollectionView, State, Region, and Behavior do not form a
-public inheritance hierarchy. They compose first-class collaborators and may satisfy
-small shared protocols or reuse internal implementation without exposing inheritance
-as the application architecture.
+Application, MnObject, View, CollectionView, the native `State` candidate, Region, and
+Behavior do not form a public inheritance hierarchy. They compose first-class
+collaborators and may satisfy small shared protocols or reuse internal implementation
+without exposing inheritance as the application architecture.
 
 An Application may coordinate one root View through a Region it constructs and owns
 or a borrowed host Region it receives from its owner. The Application shows that View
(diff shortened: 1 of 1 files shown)


paulfalgout:
**Approved** — verified at fetched SHA `4f836ccf` (base `8cb2b6a2`). ROADMAP-only, and the scope statement is accurate: nothing here changes runtime behavior.

The direction is right, and the strongest argument for it is one the PR does not make: **borrowed-versus-owned is not a new idea being introduced to state, it is an existing shipped pattern being generalized.** #379 already implements exactly this split for Regions — `_ownedRegion` is destroyed at Application teardown, a borrowed host Region is left alone, and the transactional destroy guard keys specifically on the owned one. State was the outlier that treated every source as owned. Bringing it in line makes the two ownership stories one story.

### What this designates as wrong is what merged two commits ago

Worth recording explicitly, because #379 merged roughly an hour before this PR opened and it *pinned* the semantics this contract replaces. Four elements of current master contradict it:

| `mixins/state.js` on master | New contract |
|---|---|
| `definition instanceof State ? definition : new State(definition)` | "A plain state object never silently selects and constructs a store." |
| `state._owner = this` on a supplied instance | "A supplied source is borrowed." |
| `throwStateOwnershipConflict()` / **MN0035** — *"must be live and unowned before composition"* | Borrowed sources remove the premise; multiple borrowers are legal. |
| `_destroyState()` destroys the supplied state | "Marionette releases only its own subscriptions." |

The last one is pinned by assertion: `owned-cleanup.spec.js` builds `const state = new State()`, passes it as `new TestApplication({ state })`, and asserts `state.isDestroyed()` is **true** after destroy — a caller-supplied instance destroyed by the framework. Under this contract that assertion inverts.

None of that is a defect in this PR; a roadmap describing a target is supposed to diverge from current code. I am flagging it as sequencing, because two consequences are easy to lose:

1. **MN0035 will need retiring, not deleting.** #309 added the `retired` status precisely so a diagnostic code cannot be silently reused. When supplied sources become borrowed, MN0035's message is no longer true, and it has to go through that mechanism.
2. **Some of BA0009's bytes are for machinery this removes.** The `_owner` assignment, the conflict check, and the unconditional destroy in `_destroyState` are part of the +309 B just authorized and shipped. If borrowed semantics lands, part of that comes back out — which is a pleasant direction for the ceiling, but means the 79,779 figure is not a stable floor for planning.

### One gap

The contract says a supplied source is borrowed and Marionette "releases only its own subscriptions," and separately that supplying an already-owned source no longer conflicts. What it does not say is what `getState()` returns when **two owners borrow the same source** and one is destroyed. Subscription release is per-owner, so this should be safe by construction — but "safe by construction" is exactly the kind of claim that deserves to be stated, since the old model made it impossible by fiat and the new one makes it ordinary. A sentence confirming that borrowing is many-to-one and that one borrower's destruction does not disturb another's observation would close it.

Everything else — the DataApi separation, the native-State-as-candidate framing, the allocation and stable-release gate rewording — is consistent, and I could not find a place where the old ownership-transfer language survives.

paulfalgout:
**Approval stands** at `647371a2`. The gap I raised is closed, precisely:

> Borrowing is many-to-one: destroying one owner releases only that owner's observation and does not alter the shared source or another borrower's observation.

That is the sentence the contract needed — it states the property rather than leaving it to be inferred from per-owner release, and it is the line a future reader will check when someone proposes a shared-source optimization.

The rest is backtick consistency for the native `State` class against the general concept of state composition, which matters more here than typography usually does: this PR's whole point is that `State` is now one candidate source among many rather than the universal model, and the two readings appear in adjacent sentences.

One thing worth calling out in the stable-release gate, since it settles an ordering question the prose left open:

> factory results are disposed exactly once after subscription release

Release-then-dispose, and *exactly once* — consistent with the LIFO teardown discipline the codebase just standardized on through `disposeAll`, and testable as written rather than aspirational. Together with "supplied sources survive owner destruction," that gate is now specific enough to fail against the current implementation, which is what a gate should do.

My sequencing notes from the previous review are unchanged: MN0035 goes through #309's `retired` mechanism rather than being deleted, `owned-cleanup.spec.js` inverts its supplied-state assertion, and part of BA0009's authorized bytes come back out when the ownership machinery goes.

paulfalgout:
**Approval stands** at `fae4ba4e`. The additions are all improvements, and dropping "ownership record" from both allocation lists is the right consistency catch — with `_owner` gone there is no ownership stamp left to allocate, so listing one as a thing tests must prove absent would have outlived its own model.

The strengthened release gate is good and testable as written: *"public `off()` cannot disable later subscription release or owned-source disposal"* generalizes the #375 property to state sources, and splitting the startup-invalidation sentence into Marionette's obligation versus the owner-hook's obligation makes both checkable instead of one blurred claim.

### The new lifetime rule trades enforcement for documentation

> Supplying a factory result to another owner does not transfer ownership or extend its lifetime; consumers must not arrange for a borrower to outlive that source's owner.

This closes the hole I was circling, and the scoping is exact — a consumer-created supplied source is borrowed by everyone and owned by no one, so nothing disposes it and no hazard exists; the risk is specific to factory results, which Marionette *does* dispose. Naming only that case is right.

What I want on the record is what changed underneath it. Today this arrangement is **mechanically impossible**:

```js
if (state._owner !== undefined || state.isDestroyed()) { throwStateOwnershipConflict(); }
```

MN0035 rejects supplying an already-owned State outright. The new model deliberately makes that legal — borrowing is many-to-one — and in doing so removes the check that made the lifetime error unreachable, replacing it with a "must not" that nothing verifies. The invariant did not get weaker by accident; it got relocated from the runtime to the consumer's discipline.

That is a defensible trade, and I am not arguing against the direction. But the failure mode is worth pricing: owner A creates a source through its factory, hands it to B as a supplied source, A is destroyed, the source is disposed, and B continues observing a disposed source. B's subscriptions are still live and its own release still works, so nothing in Marionette notices — the symptom surfaces as stale or absent updates in whatever B renders, arbitrarily far from the destroy that caused it. That is the same silent class as the `observeCollection()` noop disposer that #377 committed to making explicit.

Since the roadmap is already committing to dev-mode diagnostics elsewhere, the natural question is whether this earns one — a check at subscription time that a borrowed source is not a factory result belonging to a different owner, or at minimum a diagnostic when observation is registered against an already-disposed source. That would keep the mechanical guarantee MN0035 provides today without reinstating single-ownership. If the answer is "no, this is consumer discipline," that is a fine answer, but it should be a decision rather than a consequence of removing the old check.

Related sequencing point from my earlier review: this is the second reason MN0035 needs #309's `retired` treatment rather than deletion. Its message stops being true *and* the condition it guards becomes legal, so anything reusing the code later would be asserting the opposite of what it once meant.

paulfalgout:
**Approval stands** at `b91c3144`.

The restoration is the notable change, and it is a good self-catch. The original diff dropped this constraint entirely:

> State is deliberately small and synchronous. It does not implicitly render Views, schedule work, compute derived values, run effects, or persist data.

I missed that in my first review — I checked that no *ownership-transfer* language survived the reframing and did not notice that a genuine restriction had left with it. It is back, now scoped to the native `State` candidate rather than asserted of every source, which is where it always belonged: "no derived values, no effects, no persistence" is a real design commitment for a dependency-free store Marionette ships, and an untrue claim about an XState actor or a Redux store. Losing it would have quietly removed the reason native `State` stays small.

Generalizing the lifetime rule to *"A borrower must not outlive its source; borrowing never extends a source's independently owned lifetime"* is better than the factory-specific version. It now also covers a consumer-managed source the consumer destroys early, which the previous phrasing left out.

Collapsing the per-class lifetime enumeration into one rule — *"Every eligible owner releases its state-source subscriptions at destruction and also disposes a factory result it owns"* — is a real simplification rather than a reword. The old sentence listed which classes end a source at destruction, which invited the reading that the rule varies by class; it does not, and the gate now states it per-owner and uniformly.

The enforcement question from my last comment is unaddressed — the rule is still a "must not" that nothing checks — but I have made that case once and it is a decision for you rather than something to keep restating. Flagging only so it does not disappear by attrition.

paulfalgout:
Exact-head review refreshed at `b91c3144433d1b3fa4b99c23c16dcba6991bce49`.

Claude reviewed the complete ROADMAP diff against base `8cb2b6a2`, including the amended shared-borrower, teardown, cancellation, and native `State` constraints. Final verdict: **Approve — no blocking issues.** The prior CollectionView teardown omission is now covered by the blanket eligible-owner destruction rule. Remaining notes are implementation-time wording/protocol questions for #191, not blockers for this roadmap-only PR.

Local validation at this head:
- `git diff --check` — passed
- `npm run check:docs-examples` — 7 tests passed; 22 examples validated
- `npm run docs:check` — executable-example phase passed, then local build could not start because this temporary worktree does not have the `marked` dev dependency installed; the fresh CI Documentation job is the authoritative full docs gate.

paulfalgout:
**Approval stands** at `118b47ef`, and this corrects something I endorsed too broadly one push ago.

I called the collapse at `b91c3144` "a real simplification rather than a reword." Half right. The **destruction** rule genuinely is uniform — every eligible owner releases subscriptions and disposes what it owns — and stating it once was an improvement. But **persistence** is not uniform, and collapsing both into one sentence quietly dropped two facts the pre-PR text carried: a Behavior's source survives its owning View's render, and an MnObject's lasts the object's lifetime. Restoring those while keeping the destruction rule unified is the correct split, and it is a distinction I should have caught rather than praised past.

The two halves genuinely differ in kind. Persistence answers "which lifecycle events does the source survive," and that varies by owner because the owners have different lifecycles — render for Views and Behaviors, stop/restart for Applications, nothing at all for MnObject. Destruction answers "who releases what at the end," and that is the same everywhere by design. One sentence covering both invited the reading that destruction varies by class too, which is exactly the ambiguity the collapse was trying to remove.

No other changes; everything from my earlier reviews stands, with the enforcement question still the one open item.

paulfalgout:
Exact-head approval refreshed at `118b47ef2ca3281cefd76a8373df8c0d90345292`.

Claude compared the direct `b91c3144` → `118b47ef` snapshots and reviewed Cubic discussion `r3916351106`. Verdict: **Approve** — the narrow delta restores both missing guarantees (Behavior state survives its owning View render; MnObject state persists for its lifetime) with no other content drift.

Local checks at this head remain green: `git diff --check` and `npm run check:docs-examples` (7 tests; 22 examples).

PR #390, opened 2026-09-03
chore(performance): remove adapter transition bridges

paulfalgout:
## Linked issue

Related to #389.

## What

- Remove the CI-only resource-loader and consumer-fixture transition bridges used while #389 was based on pre-relocation code.
- Restore direct exact-base comparison of the unmodified candidate performance report.
- Keep a contract assertion that the retired bridge names do not return to the authority step.

## Why

PR #389 documented these bridges as one-PR compatibility exceptions with an explicit removal condition: delete them once the adapter relocation became the exact pull request base. That condition is now satisfied on `master`.

- Temporary compatibility machinery can otherwise become an undocumented permanent path and conceal future performance-contract drift.

## Public behavior contract

- Canonical behavior after this PR: the exact-base authority loads the relocated adapter and compares the current consumer fixture directly.
- Superseded behavior removed: candidate-owned resource import interception and fixture-revision normalization.
- Compatibility exception and removal condition: none remain.

## Runtime cost

- Bundle: no production artifact or contract change; the measured candidate remains 89,862 Brotli-11 bytes.
- Allocations/retention: unchanged; the exact-base deterministic resource report passes directly.

## Scope

- Affects only performance CI orchestration and its contract tests.
- Does not change runtime source, package exports, declarations, fixtures, release policy, or performance thresholds.

## Deployment Impact

- No package publication, release, tag, deployment, or redeploy is required.

## Validation

Commands:

```sh
npm ci
npm run test:performance -- --reporter=dot  # 118 tests passed
node --test test/performance/contract.test.mjs --test-reporter=dot  # 34 tests passed
npm run lint:ci
```

I also ran the merged #389 bundle-size script from an exact detached `fa885eb2` base checkout against this candidate without either bridge. Artifact and resource validation passed, the raw consumer fixture revision matched, the report had no violations, and its growth-approval status was `Not required`.

## Testing

- Performance contract, deterministic resource, consumer-bundle, exact-base report, and lint checks passed locally.
- Browser and package suites were not rerun because this PR changes no runtime, package, fixture, or build artifact.

## Package and browser impact

- Production/dev/test module-graph impact: deletes five temporary performance support/test files; production graphs are unchanged.

## Rollback or deprecation

- Revert this commit only if the exact base regresses to a pre-#389 checkout that cannot load the relocated adapter or recognize the migrated fixture revision.

## Review notes

- Exact base: `fa885eb27b576d98f2bbc3fd6591d22569c6c9d0`.
- Candidate: `05a853c467217c6db2641db46f96eb86de53a2a2`.

---
## Summary by cubic
Removes the temporary transition bridges that let performance CI compare against the pre-relocation base, so the exact-base authority now compares the relocated adapter directly.

- Deletes the resource-loader and consumer-fixture bridge scripts and their test.
- Restores the authority report to compare `bundle-size-base.json` with `bundle-size-authority.json` directly.
- Adds a contract assertion that the retired bridge names do not return to the CI step.

<sup>Written for commit 05a853c467217c6db2641db46f96eb86de53a2a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/390?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approved** — verified at fetched SHA `05a853c4` (base `fa885eb2`).

**The removal condition is structurally satisfied, not merely asserted.** Both base and candidate now carry fixture SHA-256 `3bfbbe21…`, so `base.fixtureRevision !== current.fixtureRevision` no longer fires and the consumer comparison passes on its own. `packages/adapters/dist/` exists on master, so the exact-base resource probe resolves the relocated path without interception. Both bridges are dead because the world moved, not because the check was relaxed — which is the only good reason to delete a compatibility shim.

The CI step is an exact revert to its pre-#389 shape, and the guard is well constructed:

```js
assert.doesNotMatch(
  authorityStep,
  /base-resource-backbone|bridge-consumer-fixture-transition|exact-base-resource-loader|comparison_report/
);
assert.match(
  authorityStep,
  /--report ".*bundle-size-base\.json" ".*bundle-size-authority\.json"/
);
```

Naming all four retired identifiers *and* asserting the direct path is restored is the right pairing. A negative-only assertion would also pass if someone broke the authority step entirely; requiring the direct report keeps "no bridge" from being satisfied by "no comparison."

This closes the note I raised on `cdd9fe74`. My concern was specifically that the fixture bridge — unlike the resource loader — carried no removal condition, and would sit as permanently dead code with nothing recording it was finished. Removing both in the very next PR is a better answer than the comment I suggested, and the reintroduction guard is better still: the removal condition is now enforced rather than remembered.

The stated rationale is exactly right, and worth having on the record:

> Temporary compatibility machinery can otherwise become an undocumented permanent path and conceal future performance-contract drift.

That is the failure mode. A bridge that outlives its transition does not announce itself — it just quietly normalizes whatever it was built to normalize, and the next genuine drift of the same shape passes through it.

Scope: I verified the removal condition, the CI revert, and the contract assertions. I did not re-run the performance job.

PR #391, opened 2026-09-03
refactor(radio): isolate request methods to channels

paulfalgout:
Related #41

## What

- Keep request/reply methods on Radio channels instead of composing `RequestsMixin` into every `CommonMixin` consumer.
- Let `EventsMixin` be the single source of `triggerMethod` and move the constant-or-function callback wrapper into `mixins/requests.js`.
- Delete the obsolete helper module, rebuild the checked-in distributions, and verify the removed methods across every public framework class and output format.

## Why

The Phase 1 roadmap requires request/reply adaptation to remain private to Requests. Framework objects unintentionally inherited `reply`, `replyOnce`, `stopReplying`, and `request`, even though application-owned request bindings already use explicit Radio channels through `getChannel()`, `bindRequests()`, and `radioRequests`.

- Prevents agents and consumers from treating arbitrary framework instances as request registries.
- Gives `triggerMethod` one composition source instead of a duplicate CommonMixin entry.
- Keeps the callback wrapper next to the only subsystem that owns its removal semantics.

## Public behavior contract

- Canonical behavior after this PR: `Application`, `Behavior`, `CollectionView`, `MnObject`, `Region`, and `View` expose event methods but not Radio channel request methods. Radio channels and the top-level Radio facade continue to expose the complete request API.
- Superseded behavior removed, if any: implicit `reply`, `replyOnce`, `stopReplying`, and `request` methods on CommonMixin-based framework objects.
- Compatibility exception and removal condition, if any: none; v5 is pre-stable and the obsolete path is removed without an alias.

## Scope

- Changes CommonMixin composition, the private Requests helper boundary, generated core bundles, and source/dist contract tests.
- Does not change `bindRequests`, `unbindRequests`, `radioRequests`, Radio channel behavior, package entrypoints, optional adapters, or `@marionette/data`.

## Runtime cost

- Bundle: exact-base cumulative Brotli-11 decreases 6 B to 89,856 B. ESM decreases 11 B, CJS decreases 23 B, minified UMD decreases 17 B, and unminified UMD increases 45 B (0.19%); approval is not required.
- Startup/hot path: no new path or work; four inherited prototype methods are removed from framework classes.
- Allocations/retention: exact-base deterministic resource comparison reports no change.

## Deployment Impact

No deployment, package publication, release, or tag is performed. A future normal Marionette package release will include the changed core bundles; the adapters and data packages are unchanged.

## Validation

Commands:

```sh
npx vitest run test/unit/mixins/common.spec.js test/unit/requests.spec.js test/unit/object-application-composition.spec.js test/unit/radio.spec.js --reporter=dot
npm test -- --reporter=dot
npm run lint:ci
npm run test:performance -- --reporter=dot
npm run size
npm run test:source
npm run test:dist
npm run test:fixtures
npm run test:browser
npm run docs:check
```

Results:

- Focused contract: 4 files, 78 tests passed.
- Full unit suite: 110 files, 2,047 tests passed.
- Performance contract: 118 tests passed.
- Source, built distribution, packed ESM/CommonJS fixtures, and Chromium/Firefox/WebKit browser validation passed.
- Documentation build validated 22 executable examples, 56 HTML files, and 503 internal links.
- Exact-base bundle/resource reporting is refreshed by CI for each amended head; no growth approval is expected because the runtime artifacts are unchanged from the previously measured commit.

## Package and browser impact

- Entry points or install fixtures affected: root `marionette` runtime bundles only; no export-map change.
- Browser contracts affected: built UMD and minified UMD now omit the request methods from all six public framework class prototypes; browser behavior otherwise remains unchanged.
- Production/dev/test module-graph impact: removes `utils/make-callback.js` from the root production graph (50 to 49 internal modules).

## Rollback or deprecation

Revert this PR before v5 stabilizes if a verified consumer demonstrates that arbitrary framework instances, rather than Radio channels, must own request registries. No compatibility shim should be added without that evidence and an explicit removal condition.

## Review notes

- A bounded Claude review received the source-level change summary, affected files, validation results, and open questions. It correctly requested negative coverage for all CommonMixin consumers; the dist validator was expanded from `MnObject`/`Application` to all six public classes.
- The review also asked for callback-identity and Radio-composition evidence. Existing focused tests already remove `replyOnce` by its original callback, preserve constant replies, assert Radio channel composition from `Events` plus `Requests`, and exercise every forwarded Radio request method, so no redundant runtime change was made.
- The Phase 0 performance baseline remains immutable; the exact-base report records the deleted helper as a production-graph removal.


paulfalgout:
**Approved** — verified at fetched SHA `b2d656e5` (base `0327f693`). Core aggregate **88,400**, essentially flat against master's 88,406.

Worth noting *why* it is flat, since removing four methods from six classes sounds like it should save more: the implementations still ship for Radio channels, so only the composition changed, and inlining `make-callback.js` into `mixins/requests.js` offsets the rest. A 6-byte drop is the right order of magnitude here — this is a boundary change, not a code deletion.

### The dist-level assertion is the right kind of proof

```js
for (const className of ['Application', 'Behavior', 'CollectionView', 'MnObject', 'Region', 'View']) {
  for (const methodName of ['reply', 'replyOnce', 'stopReplying', 'request']) {
    assert.strictEqual(methodName in Marionette[className].prototype, false, …);
```

Using `in` rather than `hasOwnProperty` is the detail that makes this work — the methods were arriving through mixin composition onto the prototype chain, and an own-property check would pass while an inherited method was still reachable. Running it across every output format catches a build that composes differently than source. Same discipline as the `DataApi.items` negative assertions in #386.

### The gap: nothing records this removal

`changelog.md`, `upgradeGuide.md`, and `docs/migration-from-v4.md` are all untouched, and none of them mentions the removal.

That breaks a pattern this project has otherwise held consistently. The changelog carries lines like *"Removed `children.findByModelCid`; `findByModel` uses the configured DataApi key"*; #381's `State` removal, #385's `getParentApp`/`getRootApp`, and #386's `DataApi.items` each landed with guidance. This one removes four methods from six public classes with no record anywhere.

I could not determine from this repository whether v4 exposed these on framework objects — there are no v4 tags here — so I cannot say whether this belongs in `migration-from-v4.md` or in `upgradeGuide.md` as an alpha-to-alpha removal. But it belongs in one of them, and in the changelog regardless.

The `docs/migration-from-v4.md` row that mentions `replyOnce` is about override dispatch on Radio channels and is unaffected, so there is no active contradiction to fix — only an absence.

The remedy is small, and the substitute is already available and idiomatic: `getChannel().reply(…)` and `bindRequests()` / `radioRequests`. Unlike #385's upward-lookup removal, this one has a direct one-line replacement, which makes the missing note cheaper to write and more useful to whoever hits `view.reply is not a function`.

Scope: I verified the aggregate, the dist assertions, and the documentation state. I did not review the `CommonMixin`/`EventsMixin` `triggerMethod` consolidation in detail.

paulfalgout:
**Approval stands** at `4c79f255`. Docs only; core aggregate remains **88,400**.

Both gaps are closed, and the entries answer the question I could not settle from the repository: **"alpha-only"** in both the changelog and the guide establishes these methods never shipped in v4, which is exactly why `docs/migration-from-v4.md` correctly stays untouched. That was the piece I was missing — no v4 tags here to check against — and stating it in the text means the next reader does not have to reconstruct it either.

Placing the guide entry directly after the existing `Radio.Requests` mixin paragraph is the right spot: a reader arriving because of the v4 mixin removal is the same reader most likely to have adopted the alpha instance methods, and now finds both halves together.

The example covers the substitution properly — explicit channel, `radioRequests`, and `bindRequests` — and this line earns its place:

> owned cleanup then removes only that owner's replies

That explains *why* `bindRequests(channel, bindings)` is preferable to calling `channel.reply(…)` directly from an owner, rather than presenting them as interchangeable spellings. Given how much of this milestone went into making owned cleanup structural, pointing consumers at the path that participates in it is worth the extra sentence.

Everything I raised on this PR is addressed.

PR #397, opened 2026-09-03
perf(collections): reduce bulk cleanup allocations

paulfalgout:
Related #127

## What

- Replace bulk child-view cleanup closure arrays with allocation-free, forward attempt-all loops while preserving first-error precedence.
- Use one canonical `@marionette/data` binding-replacement path that skips Set/filter work for empty inputs and records rollback progress by index.
- Add regression coverage for cleanup error precedence and failed empty-reset rollback, plus comments tying the specialized loops to the `disposeAll` contract.

## Why

Clearing a 1,000-row CollectionView paid avoidable bulk and per-child allocation costs before performing the required lifecycle cleanup. Exact-head browser measurements against `master` (`17cdb841`) showed the initial cleanup candidate (`069cd456`) reducing median Clear script time from 24.3 ms to 22.3 ms, approximately 6.4%. The mean improvement was approximately 5.6%, with a bootstrap interval of roughly 3.2–7.9%.

A follow-up 15-sample comparison included the canonical binding-replacement implementation now in this PR and found no statistically measurable Clear regression from `069cd456`; all confidence intervals crossed zero. Experimental create-path changes in that comparison also produced no measurable create improvement and are intentionally excluded from this PR.

The exact amended head (`4d07af45`) has a cumulative eight-artifact Brotli-11 size of 90,332 bytes, +205 bytes from the PR base and below every approval threshold. The canonical consumer scenario remains +78 bytes at 18,968 bytes.

## Scope

This changes internal cleanup implementation for CollectionView children and `@marionette/data` collection binding replacement. Existing forward child-destruction order, attempt-all cleanup, rollback behavior, and first-error precedence remain unchanged.

No public API changes are included. Container append, `_set()` self-array handling, child-view option construction, `_bindModels()`, rendering, the broader `View.destroy()` pipeline, DOM-detach behavior, and renderer strategy are intentionally unchanged.

## Deployment Impact

No separate deployment or consumer redeploy is required. The change will be included through the normal Marionette package release process.

## Testing

- `npx vitest run test/unit/data-package/collection.spec.js test/unit/collection-view/collection-view-lifecycle.spec.js --reporter=dot` — 45 tests passed.
- `npm run coverage -- --reporter=dot` — 2,058 tests passed across 111 files with 100% statements, branches, functions, and lines; 19 agent-benchmark contract tests also passed.
- `npm run test:performance` — 118 tests passed.
- `npx eslint modules/collection-view.js packages/data/src/collection.js test/unit/collection-view/collection-view-lifecycle.spec.js --max-warnings=0` — passed.
- `npm run build` followed by `node scripts/checks/dist.mjs` — passed on the committed tree.
- Exact-head CI passed Node 24, Node 26 advisory, Linux coverage, macOS and Windows package smoke, verified tarballs, release-package smoke, bundle, CodeQL, documentation, performance-contract, and promotion dry-run checks.
- Controlled browser A/B used 15 samples per compared implementation; keyed, CSP, and CPU plausibility checks passed.


paulfalgout:
**Approved** — reviewed at fetched SHA `069cd456` (base `17cdb841`).

### The translations are faithful — I checked the orderings rather than the descriptions

**`_removeChildViews`.** The old line was a double negation worth being rid of:

```js
disposeAll(cleanups.reverse());
```

`cleanups` was built forward, `.reverse()` flipped it, and `disposeAll` iterates in reverse — so the net execution order was forward. The new explicit forward loop produces the same order and no longer requires the reader to compose two reversals to work that out. Attempt-all, first-error-wins via `if (!hasError)`, and the rethrow are all preserved.

**`_removeChildView`.** The old `disposeAll([stopListening, detachOrDestroy])` executed **detach/destroy first** because of that reverse iteration, and the new code runs them in that same order explicitly. The first `catch` assigns unconditionally and the second guards on `!hasError`, which reproduces "first executed disposer's error wins" exactly. Easy to get backwards; it is right.

**The `@marionette/data` empty path.** The rollback bound is the detail that matters: on a throw at index `i`, `attemptedIndex` is still `i` because the increment runs after the body, and the rollback loop restores `i` down to `0` — including the model that failed. That is correct, because `_unbindModel` forcibly unbinds and drops ownership before rethrowing, so the thrower *is* partially unbound. It also matches the general path's existing behaviour, which restores every attempted model including the failing one. Swallowing rollback errors under the original matches `disposeAll(disposers, error)` seed semantics.

### Two notes

**The gated figure is missing.** You cite 18,890 → 18,968 B (+78), which is a consumer-scenario measurement. The eight-artifact aggregate this PR is gated on moves **90,127 → 90,332, +205 B**. Both are real, but the ceiling binds on the second, and recent PRs have quoted it. Still comfortably inside 100,000 with no artifact near the one-percent threshold.

**Teardown semantics are now hand-rolled at three sites.** This milestone spent considerable effort making `disposeAll` the single primitive with tested attempt-all, first-error, arity, and falsy-skip behaviour. These loops correctly reimplement a subset of it — today. What they do not inherit is `disposeAll`'s tests: if someone later adds a third step to `_removeChildView`, nothing forces them to notice the `if (!hasError)` guard, and the failure mode is a *later* error masking the first, which no test here would catch.

The measurement justifies the trade — median 24.3 → 22.3 ms with a bootstrap interval that excludes zero is real evidence, not a micro-benchmark hunch. But a one-line comment at each site naming the contract these loops must preserve would keep them tied to it, the same way the `_isDestroyed`-ordering and `_replaceModel` key-stability comments do elsewhere.

Scope: I verified both translations, the rollback bounds, and the aggregate. I did not review the browser timing methodology or reproduce the measurements.

paulfalgout:
**Approval stands** at `4d07af45`. Gated aggregate unchanged at **90,332**.

The comments landed at both sites, and they name the thing that matters — *"Preserve disposeAll's attempt-all, first-error contract without closures"* — so the next person adding a step to `_removeChildView` meets the constraint before the guard rather than after.

### The `_replaceBindings` restructure is better than what I reviewed

The empty-path special case is gone, replaced by narrowing only the allocation:

```js
let added = currentModels;
let removed = previousModels;
if (previousModels.length && currentModels.length) {
  // Set + filter diff
}
```

Three improvements over the branch I approved:

**It covers both directions.** The special case only skipped the Sets when `currentModels` was empty — the clear path. This also skips them when `previousModels` is empty, which is the initial-populate path, and both degenerate cases are trivially correct: an empty `previous` means every current model is added, an empty `current` means every previous model is removed.

**There is one rollback instead of two.** The previous version duplicated the restore loop inside the empty branch. Duplicated recovery logic is the worst kind to duplicate, because the copy is only exercised when something has already gone wrong.

**The duplicate had been quietly incomplete.** The shared path also unwinds `added` on failure:

```js
for (let index = added.length; index--;) { try { this._unbindModel(added[index]); } catch {} }
```

The empty-branch copy omitted that. Harmless in isolation — `added` is empty on the clear path — but it meant the two paths did not actually implement the same contract, and only the input made them agree. I checked the ordering against the original `disposeAll` form from #384: restore `removed` in reverse including the thrower, then unbind `added` in reverse. Matches.

Nothing else changed, and the fast path aliases the caller's arrays without mutating them, so there is no copy-elision hazard.

PR #398, opened 2026-09-04
perf(collections): reduce create-path allocations

paulfalgout:
## Linked issue

- No linked issue.

## What

- Append child views with `push()` when no insertion index is needed.
- Avoid copying the child-view buffer and allocating default child options on common paths.
- Replace successful `@marionette/data` batch-binding bookkeeping with an index while preserving reverse best-effort rollback.

- Removes transient arrays and callback closures identified while profiling collection creation, without turning an unmeasured micro-optimization into a public speed claim.
- Makes the `_set(this._views, ...)` aliasing case explicit and covers it directly so removing the defensive copy cannot empty the container.

## Public behavior contract

- Canonical behavior after this PR: collection child order, indexes, model ownership, rollback order, lifecycle behavior, and errors remain unchanged.
- Superseded behavior removed, if any: none; only equivalent internal allocation paths are replaced.
- Compatibility exception and removal condition, if any: none.

## Runtime cost

- Bundle: the binding eight-artifact cumulative Brotli-11 gate is 90,332 → 90,441, +109 B (+0.12%). The string-renderer benchmark application is +32 compressed bytes.
- Startup/hot path: a 15-sample exact-head A/B at merged cleanup `fcc6fe7d` versus `2e21b5c0` found no demonstrated creation-speed improvement: create 1k script mean -0.4% (95% bootstrap interval -2.3% to +1.6%); create 10k -0.6% (-1.7% to +0.4%).
- Allocations/retention: removes an append-path `splice`, the redundant `_views` copy, default child-options merge/allocation, and the successful batch-binding tracking array and rollback closures. All three 15-sample retained-memory intervals crossed zero; the deterministic resource harness reports no retained-resource proxy change and does not measure transient allocations.

## Validation

Commands:

```sh
npm run coverage
npm run lint:ci
npm run test:source
npm run test:performance
npm run build
npm run test:browser
npm run test:fixtures
npm run docs:check
npm run test:dist
npm run test:release-promotion
npm run check:release-promotion
node scripts/checks/dist.mjs
```

Results: 2,060 unit tests and 19 agent benchmark contract tests passed with 100% coverage; 118 performance tests passed; Chromium, Firefox, and WebKit passed; ESM, CJS, TypeScript, and Vite package fixtures passed; 23 executable docs examples and docs link validation passed; distribution and dry-run release-promotion checks passed with publication disabled. Exact benchmark artifacts also passed keyed and CSP checks.

## Package and browser impact

- Entry points or install fixtures affected: generated core and `@marionette/data` distributions change; package fixtures passed.
- Browser contracts affected: none; the browser suite passed in Chromium, Firefox, and WebKit.
- Production/dev/test module-graph impact: production module count and external imports are unchanged; tests add focused aliasing and rollback coverage.

## Deployment Impact

- No deployment is required. The changes would be included in the next normal Marionette package release; this PR does not publish or tag a release.

## Rollback or deprecation

- No migration or deprecation is involved. Reverting this commit restores the prior allocation paths.

## Review notes

- The macrobenchmark does not support presenting this as a measured speed improvement.
- Two initially slower guardrails were rerun with the framework order reversed. Across 30 samples, replace was +1.0% (95% interval -0.1% to +2.1%) and partial update was +2.5% (-0.7% to +5.8%); neither reproduced as a regression.
- All nine CPU cases, all three available memory cases, and size used the ordinary string renderer. Copilot review was unavailable because its requester quota was exhausted; CodeRabbit reported no actionable code comments.


paulfalgout:
**Approved** — reviewed at fetched SHA `2e21b5c0` (base `fcc6fe7d`). All three changes are behaviour-preserving; I verified each rather than reading the summary.

### The `_set` change is the interesting one

```diff
- this._views.length = 0;
- this._views.push.apply(this._views, views.slice(0));
+ if (views !== this._views) {
+   this._views.length = 0;
+   this._views.push.apply(this._views, views);
+ }
```

`views.slice(0)` was not a stray copy — it was a defensive one. Without it, `_set(this._views, …)` would clear the source array before pushing from it, emptying the container. Replacing an implicit copy-on-every-call with an explicit identity guard is a legibility improvement independent of allocation: the aliasing hazard is now named in the code rather than silently defended against, and a reader no longer has to work out why a fresh array was being copied again.

I checked that `shouldReset` still runs when the guard short-circuits — it sits outside the `if`, so index rebuilding is unaffected — and that no partial-aliasing case escapes, since `push.apply` only reads from `views` and a distinct array cannot be disturbed by clearing `this._views`.

`splice(len, 0, view)` → `push(view)` is exactly equivalent at the append index, and skipping `assignOwn` when `childViewOptions == null` produces the same object it would have returned.

### The gated figure is absent again, and it matters more here

> The benchmark app's compressed bundle increased by 32 bytes.

The eight-artifact aggregate this PR is gated on moves **90,332 → 90,441, +109 B**. That is the second consecutive PR quoting a non-gated size metric — #397 cited a consumer-scenario delta of +78 while the gate moved +205.

It matters more on this one because of how straight you were about the performance result: create 1k at -0.4% with a bootstrap interval of -2.3% to +1.6%, create 10k at -0.6% spanning zero. Declining to claim a speedup when the intervals do not support one is exactly right, and rarer than it should be. But it means the whole visible trade is cost against unmeasured benefit — and stated as +32 bytes, that trade reads roughly three times cheaper than the gate records.

I am not arguing against the change. Removing allocations is principled, and the `_set` guard earns its bytes on clarity alone. The ask is only that the number quoted be the number that binds, so the trade is legible to whoever revisits this.

Worth noting the current resource harness would not settle it either way: `deterministicResources` measures retained per-instance shapes, not transient allocation over a workload, so "fewer allocations" has no direct evidence path today. That is a gap in the tooling rather than in this PR, but it is why the byte cost is the only measured quantity in the trade.

paulfalgout:
Updated the PR description to lead with the binding gate: cumulative Brotli-11 is 90,332 → 90,441 (+109 B). The +32 B figure is now explicitly identified as the benchmark application bundle rather than the release gate.

Issue #376, opened 2026-09-02
v5: close collection-source integration and evidence gaps


paulfalgout:
## Current status — September 8, 2026

The source, public typing, and bounded CollectionView/data integration closeout are delivered. No speculative provider adapters or exception-recovery framework is required before the next alpha.

The bounded source audit is complete on [master `95d1955e`](https://github.com/marionettejs/marionette/commit/95d1955e7d1ed2bba4ce666968b0fd6a4047c28f), including #444–#453. This is a verified source candidate for the maintainer's next local alpha. Public stable-release, portable benchmark, and agent-evaluation work remains in #127/#147 and the corresponding agent issues; this closeout does not declare stable-v5 readiness.

## Accepted boundaries
- Core provides neutral DataApi, plain-object access, static arrays, and one CollectionView reconciliation path.
- `@marionette/data` is the optional Model/Collection package; it shares `@marionette/utils`. It deliberately stays small. Models emit their own attribute events; collection notifications describe structure.
- `@marionette/adapters` exports Backbone, XState actors, and jQuery/Morphdom/Lit-html DOM integrations through explicit entrypoints with optional provider peers. There is no root barrel.
- Published Redux, Zustand, and XState Store adapters were removed in #436. A real Redux Toolkit consumer fixture proves that public DataApi can support stable entity IDs without core changes. Additional packaged adapters require evidence of a good integration, not a provider checklist.
- Native and Backbone observation translates ordinary synchronous events. Native observation has no extra coalescing/replay queue. Finish a structural mutation before starting another in those event sources; listener errors propagate. The former #415 observer-coordination design is superseded by #436.
- Core still captures and queues valid nested DataApi notifications, including their source ordering. This is not a promise to make an arbitrary event source transactional or to recover a partially failed View.
- No constructor/subscription rollback, arbitrary before-hook destruction/reparenting recovery, universal mutation API, or adapter-specific CollectionView reconciliation is required.

## Delivered
- [x] Strict source-generated public declarations and packed consumer tests (#429, #432).
- [x] Native Model/Collection and public DataApi integration, simplified observer semantics, and Redux consumer proof (#436).
- [x] Ordinary multi-observer delivery, caller-owned actor lifetime, pre-render mutation, repeat render, cleanup and stale-callback tests.
- [x] Backbone merged model rendering belongs to child model events, avoiding duplicate rendering and rendering unchanged merged rows (#444).
[...]

PR #401, opened 2026-09-04
feat(adapters): add XState actor integration

paulfalgout:
Related #376

## What

- Add the `@marionette/adapters/xstate` ESM, CommonJS, and type-declaration subpath for XState v5 actors.
- Use actor-reference identity, synchronous snapshot context reads, optional keyed child-actor selection, explicitly named snapshot observation, and unchanged XState emitted-event names.
- Document and test borrowed versus Marionette-owned actor lifecycle, packaging, browser behavior, types, and 1,000/10,000-actor reconciliation.

## Why

The v5 DataApi roadmap selects XState actors as a first-party external collection and state source. This adapter keeps XState-specific observation outside core while preserving Marionette ownership rules: supplied actors are borrowed, and only actors returned from `createState()` are stopped during owned-state cleanup.

## Scope

This adds one optional `@marionette/adapters` subpath and a direct development/test dependency on XState 5.32.6. It does not add XState to Marionette production graphs, change core runtime behavior, expose the private keyed-snapshot helper, or stop caller-owned parent or child actors.

## Deployment Impact

No application deployment is required. A future package release must publish the updated `@marionette/adapters` artifact; consumers install `xstate` only when using this optional subpath.

## Testing

- `npm test -- --reporter=dot` — 2,091 tests passed across 113 files.
- `npm run test:performance` — 127 tests passed, including 1,000/10,000 actor reconciliation and unchanged-selection coverage.
- `npm run test:fixtures` — packed ESM, CommonJS, Vite, and TypeScript fixtures passed.
- `npm run test:browser` — Chromium, Firefox, and WebKit package/lifecycle checks passed.
- `npm run docs:check` — 55 pages built; 505 internal links validated.
- `npm run lint:ci`
- `npm run test:source`
- `npm run check:release-promotion`
- `npm run size` — core remains 89.06 KB / 100.00 KB; new optional artifacts measure 1.40 KB ESM and 1.41 KB CommonJS; package-specific data and adapters checks passed.

---
## Summary by cubic
Adds the optional `@marionette/adapters/xstate` subpath so XState v5 actors can be used as Marionette DataApi sources, related to #376. Core runtime behavior is unchanged, and XState is not added to Marionette’s production dependencies.

**New Features**
- Reads each child actor’s `snapshot.context` for model data and keys models by actor reference, so a stopped and respawned actor is a new identity even with the same `id`.
- Requires `select` only when the parent actor snapshot feeds a CollectionView; without it, the adapter exposes model/state methods only.
- Reserves `snapshotEvent` for `actor.subscribe()` snapshots and passes every other event name to `actor.on()` unchanged.
- Treats supplied actors as borrowed and stops only actors returned by an owner’s `createState()`, after releasing their subscriptions.

**Migration**
- Consumers using the subpath must install `xstate` separately; no application deployment is required.
- The next `@marionette/adapters` release must publish the new subpath.

<sup>Written for commit d8214a74ee131ac575a0a2bf9864807a8786d1e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approved** — reviewed at fetched SHA `d8214a74` (base `ad2eead5`). Core `dist/` is untouched, so core holds at **89,065 / 100,000**; the adapters package is now twelve artifacts totalling **11,166** bytes.

This is the first adapter to land under #400's scoping, and it demonstrates why that change was right: twelve artifacts of optional provider code, none of which a core-only consumer installs, no longer press on the core ceiling.

### The ownership implementation answers a question I raised on #389

```js
disposeOwned(actor) {
  actor.stop();
}
```

On #389 I flagged that Backbone's `disposeOwned` had become a no-op, making owned and borrowed sources indistinguishable at teardown and the stable-release gate — *"factory results are disposed exactly once after subscription release"* — vacuously true for that adapter. XState is the counterexample: actors have a real disposal operation, so the distinction is observable here.

The adapter also correctly implements only *how* to dispose, not *whether*. `mixins/state.js` gates the call on `_ownsState`, which is set only in the `createState()` path, so a supplied actor never reaches this method. That is the right division: the framework owns the ownership decision and the adapter owns the mechanism.

Together the two adapters now make the point concrete — disposal is adapter-dependent, and consumers get materially different teardown depending on which they configure. That is a stronger argument than I had on #389 for stating it where the ownership contract is defined rather than only in `optional-backbone.md`, because it is no longer a Backbone quirk but a property of the contract.

### Details worth noting

**`Object.assign(createKeyedSnapshotDataApi({…}), ActorApi)`** puts `ActorApi` second, so its `get`/`has`/`serialize` override the keyed-snapshot versions. That is correct rather than accidental: with `select`, both the parent collection and the selected children are actors, so context-reading applies uniformly to both — the helper's plain-object accessors would be wrong for the children.

**`snapshotEvent` is opt-in and must be named.** The adapter does not guess which event means "snapshot changed"; without it, every `subscribe` goes to `actor.on(eventName, …)` and XState's own emitted-event names pass through unchanged. Declining to invent a magic event name is the right call for a source whose event vocabulary belongs to the consumer.

**Child actors are deliberately left alone.** Stopping only the owned parent is correct, since XState spawns and owns its children — Marionette stopping them would reach past its own ownership boundary into the machine's.

### Still open

The packages remain without an aggregate byte backstop: `adapters-package-size.mjs` enforces external-import and module-boundary invariants, not a ceiling. At 11,166 bytes across twelve artifacts that is comfortable, and this PR does not change the argument — I note it only because each new subpath makes the gap slightly more real.

Scope: I reviewed the adapter source, its ownership semantics, and the artifact totals. I did not review the 1,000/10,000-actor reconciliation tests, the browser fixtures, or the type declarations.

paulfalgout:
```json
{
  "schemaVersion": 1,
  "headSha": "d8214a74ee131ac575a0a2bf9864807a8786d1e2",
  "issueUrl": "https://github.com/marionettejs/marionette/issues/127",
  "approvedPaths": [],
  "approvedNewSubpaths": [
    "@marionette/adapters/xstate"
  ],
  "approvedNewArtifacts": [
    {
      "path": "packages/adapters/dist/xstate.cjs",
      "size": 1411
    },
    {
      "path": "packages/adapters/dist/xstate.js",
      "size": 1402
    }
  ],
  "evidenceUrls": [
    "https://github.com/marionettejs/marionette/issues/127#issuecomment-5545060846"
  ]
}
```

PR #408, opened 2026-09-05
fix(events): preserve registry during multi-event dispatch

paulfalgout:
## What

- Keep the current event registry for the duration of a multi-name or mapped `trigger` call.
- Cover clearing, removing, adding, and replacing subscriptions during dispatch, including once handlers and mapped values.
- Explain the effect of wholesale versus targeted removal during a multi-event call.

## Why

When an `alpha` handler calls `off()`, `trigger('alpha beta')` currently rereads an undefined registry and throws before dispatching `beta`. Capturing the existing registry reference restores supported Backbone-compatible dispatch behavior. Changes within that registry still take effect; a replacement registry is used by subsequent trigger calls. This is a longstanding mismatch, not claimed as a regression introduced this week.

## Scope

Only `mixins/events.js`, its public parity tests, and the events guide. One local reference replaces repeated property reads; there is no registry copy, allocation, additional guard, or new lifecycle-cancellation promise. Generated distributions are built and checked but are not committed.

## Deployment Impact

Library change only. No forms or app-frontend bundles need redeployment. Consumers receive the fix through a future package adoption; this PR publishes no package release.

## Testing

- On current master, the initial six added cases produce four failures and two passing controls. A subsequent nested-dispatch regression also passes with the fix.
- `npm test -- --reporter=dot test/unit/events-parity.spec.js test/unit/events-iteration.spec.js test/unit/mixins/events.spec.js test/unit/requests.spec.js test/unit/radio.spec.js test/unit/radio-composition.spec.js test/unit/mixins/radio.spec.js test/unit/common/bind-events.spec.js`: 163 tests pass in eight files.
- `npm run prepare`, `npm run test:source`, `node scripts/performance/bundle-size.mjs --json`, targeted ESLint and `git diff --check` pass.
- Full suite, browser/package checks and exact-base performance comparison run in CI.

Independent Claude review found no blocker. Added its nested-dispatch coverage suggestion; verified registry replacement directly. Mapped-value assertions follow Marionette's existing documented extension rather than incorrectly demanding Backbone parity.

`npm run docs:check` passes after the documentation follow-up (55 pages and 505 internal links).

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -76,6 +76,13 @@ emitter.trigger({
 });
 ```
 
+During a multi-name or mapped `trigger` call, calling `off()` from a handler
+removes subscriptions for subsequent calls but does not cancel the remaining
+event names in the current call. For example, `off()` inside a `start` handler
+still allows the existing `stop` handlers in `trigger('start stop')` to run.
+Calling `off('stop', handler)` inside `start` instead removes that handler before
+`stop` is dispatched. A nested `trigger` call uses the current subscriptions.
+
 `once` registers its generated callback through the object's overridable
 `on` method, and `listenToOnce` registers through overridable `listenTo`.
 This preserves the extension points used by event-lifecycle mixins. Likewise,
--- a/mixins/events.js
+++ b/mixins/events.js
@@ -286,14 +286,15 @@ const Events = {
   // (unless you're listening on `"all"`, which will cause your callback to
   // receive the true name of the event as the first argument).
   trigger(name, ...args) {
-    if (!this._rdEvents) { return this; }
+    const events = this._rdEvents;
+    if (!events) { return this; }
 
     if (name && typeof name === 'object') {
       const names = getKeys(name);
       for (let index = 0, length = names.length; index < length; index++) {
         const key = names[index];
         triggerApi({
-          events: this._rdEvents,
+          events,
           name: key,
           args: [name[key]],
         });
@@ -306,7 +307,7 @@ const Events = {
       for (let index = 0, length = names.length; index < length; index++) {
         const n = names[index];
         triggerApi({
-          events: this._rdEvents,
+          events,
           name: n,
           args,
         });
@@ -315,7 +316,7 @@ const Events = {
     }
 
     triggerApi({
-      events: this._rdEvents,
+      events,
       name,
       args,
     });
--- a/test/unit/events-parity.spec.js
+++ b/test/unit/events-parity.spec.js
@@ -51,6 +51,87 @@ describe('Events parity with Backbone.Events', function() {
     });
   });
 
+  const dispatchChanges = [
+    ['clears all handlers', emitter => emitter.off()],
+    ['removes the next event', emitter => emitter.off('beta')],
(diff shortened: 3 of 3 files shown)


paulfalgout:
**Approved** — reviewed at fetched SHA `858d2a61` (base `93d19e61`).

The crash is real and the fix is the right shape: `trigger('alpha beta')` re-read `this._rdEvents` per name, so an `alpha` handler calling `off()` left the second iteration passing `undefined` into `triggerApi`. Capturing once at the top matches how Backbone's `trigger` reads `this._events` a single time before entering `eventsApi`, so "Backbone-compatible dispatch" is accurate rather than aspirational.

### I verified the semantic claim on both sides, because the two `off` forms differ

Your description — *"Changes within that registry still take effect; a replacement registry is used by subsequent trigger calls"* — is exactly right, and it took reading both paths to confirm:

- **Wholesale `off()`** sets `this._rdEvents = void 0` (line 202). The old registry object is untouched and still held by the captured reference, so remaining names in the same `trigger` call **still dispatch**.
- **Targeted `off(name, callback)`** runs through `offReducer`, which mutates in place — `events[key] = remaining`, then `delete events[key]` when empty — and returns the same object. The captured reference therefore **does** observe that removal, so the remaining names **do not** dispatch to it.

So the two forms behave oppositely mid-dispatch, and that falls out of an implementation detail — whether the registry is replaced or mutated — rather than from a stated intent.

### That asymmetry deserves documenting

A handler calling `off()` to stop everything will still see `beta` delivered in the same `trigger('alpha beta')`; a handler calling `off('beta', fn)` will not. The first is the more intuitive-sounding call and the less intuitive outcome, which is the combination worth writing down.

Only `mixins/events.js` and the parity spec changed here. The tests pin the behaviour well — clearing, removing, adding, and replacing during dispatch, including `once` and mapped values — but a passing test does not tell a consumer which of the two outcomes to expect when they reach for `off()`. Given how much of this milestone has gone into stating contracts that were previously implicit, this one seems worth a sentence in the events documentation.

### Note on verification

This is the first PR since #407 where I cannot independently recompute artifact sizes, as `dist/` is no longer in the tree. The change is five lines of source and the size impact is self-evidently negligible, so nothing turns on it here — I mention it only because my earlier reviews carried independently verified byte figures and this one does not.

paulfalgout:
**Approval stands** at `314b57e0`. The documentation covers the asymmetry and one case beyond what I raised.

Both halves are stated with the contrast made explicit rather than left to inference — `off()` inside `start` still lets `stop` run, `off('stop', handler)` inside `start` does not — and using `start`/`stop` rather than abstract names gives the reader a reason a handler would plausibly call `off()` mid-dispatch, which is what makes the rule stick.

The third sentence is the addition I had not thought to ask for:

> A nested `trigger` call uses the current subscriptions.

That is the necessary complement. The capture is per-call, so a `trigger` invoked *from inside* a handler reads the registry fresh and sees the effect of an `off()` that the enclosing call is ignoring. Without that line a reader could reasonably conclude the captured registry persists for the whole synchronous dispatch tree, which would be the wrong mental model for exactly the nested case your tests cover.

Everything I raised on this PR is addressed.

PR #410, opened 2026-09-05
fix(collection-view): accept concise child view factories

paulfalgout:
## What

- Accept concise methods, arrow functions, and native class methods as `CollectionView.childView` factories.
- Add three public rendering regressions and fix the invalid `BarView` documentation example.

## Why

A documented factory can be written as `childView(model) { return RowView; }`, but these functions have no `prototype`. The constructor check currently dereferences that missing property and throws before invoking the factory. Checking for a prototype's render method allows the existing factory path to run.

## Scope

The shared view-class check and CollectionView factory coverage/documentation. Existing view constructors and factory invocation/context behavior are preserved. `emptyView` keeps its separate existing class predicate. No new API or lifecycle guard is introduced.

## Deployment Impact

No forms or application bundles need redeployment for this library-source change. Consumers receive the fix when they choose a future package version; no package publication is part of this PR.

## Runtime cost

The existing constructor check gains one nullish property check per view-class resolution. It adds no allocation, retained state, imports, or module-graph entries. Against the exact base build, Brotli changes are +7 B ESM, +5 B CJS, +3 B UMD and +6 B minified UMD; optional artifacts are unchanged. All size and graph checks pass. Timing will be reported by CI; no local timing result is claimed.

## Testing

- Before the fix: all three new factory cases throw `Cannot read properties of undefined (reading 'render')`; 44 existing cases pass.
- Independent public runtime probe — 12 constructor/factory cases pass, including bound factories and corresponding empty-view controls.
- `npm run prepare` — build and distribution validation pass.
- `npm test -- --reporter=dot test/unit/collection-view test/unit/utils/is-view-class.spec.js` — 394 tests in 10 files pass.
- `npm test -- --reporter=dot test/unit/common/view.spec.js` — 23 shared view-helper tests pass.
- `npm run lint:ci` and `npm run test:source` pass.
- `npm run docs:check` — executable examples and 55-page documentation build pass; 505 internal links validated.
- `node scripts/performance/bundle-size.mjs --json` and exact-base report comparison pass.

CI owns the fresh full coverage, browser, package-consumer, and timing runs. The root package entrypoints, public exports, diagnostics, and optional-package declarations are unchanged. Reverting this focused change would restore the factory syntax defect.

---
## Summary by cubic
Fixes `CollectionView` so concise methods, arrow functions, and native class methods work as `childView` factories. These factory forms previously threw `Cannot read properties of undefined (reading 'render')` because the shared class check dereferenced `prototype` on functions that lack it; optional chaining now lets the existing factory path run.

- The shared `isViewClass` check gains one nullish property check; `emptyView` keeps its separate predicate.
- The `BarView` docs example now defines a valid `template`.

<sup>Written for commit 3e589e061a2eec34c2b53456fe9618dfa266c5ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/410?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved view validation so missing prototypes no longer cause errors.
  - Confirmed support for defining collection child views using concise methods, arrow functions, and native class methods.

- **Documentation**
  - Corrected the collection view example to properly render the intended text.

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -363,7 +363,7 @@ const FooView = View.extend({
 });
 
 const BarView = View.extend({
-  bar
+  template: _.template('bar')
 });
 
 const MyCollectionView = CollectionView.extend({
--- a/modules/common/view.js
+++ b/modules/common/view.js
@@ -4,7 +4,7 @@ export function isView(view) {
 }
 
 export function isViewClass(ViewClass) {
-  return ViewClass.prototype.render && (ViewClass.prototype.destroy || ViewClass.prototype.remove);
+  return ViewClass.prototype?.render && (ViewClass.prototype.destroy || ViewClass.prototype.remove);
 }
 
 export function renderView(view) {
--- a/test/unit/collection-view/collection-view.spec.js
+++ b/test/unit/collection-view/collection-view.spec.js
@@ -201,6 +201,44 @@ describe('CollectionView', function() {
       });
     });
 
+    it('resolves a concise childView method for each model', function() {
+      const ChildList = CollectionView.extend({
+        childView(child) {
+          expect(this.collection).to.equal(collection);
+          expect(child).to.equal(model);
+          return MyChildView;
+        }
+      });
+      const view = new ChildList({ collection }).render();
+
+      expect(view.children.first()).to.be.instanceOf(MyChildView);
+      view.destroy();
+    });
+
+    it('resolves an arrow childView factory', function() {
+      const view = new CollectionView({
+        collection,
+        childView: child => child === model ? MyChildView : MyBbChildView
+      }).render();
+
+      expect(view.children.first()).to.be.instanceOf(MyChildView);
+      view.destroy();
+    });
+
+    it('resolves a native class childView method', function() {
+      class ChildList extends CollectionView {
+        childView(child) {
+          expect(this.collection).to.equal(collection);
+          expect(child).to.equal(model);
+          return MyChildView;
+        }
+      }
+      const view = new ChildList({ collection }).render();
+
+      expect(view.children.first()).to.be.instanceOf(MyChildView);
+      view.destroy();
+    });
+
     describe('when childView is not a valid view', function() {
       it('should throw InvalidChildViewError', function() {
         const myCollectionView = new CollectionView({


paulfalgout:
**Approved** — reviewed at fetched SHA `3e589e06` (base `7b2f975e`).

The bug is real and the diagnosis is exact: arrow functions, concise object methods, and class methods have no `prototype`, so `ViewClass.prototype.render` threw a `TypeError` before the factory path could run. `?.` on the first access is the minimal correct fix, and the second access stays safe because it is only evaluated once `prototype?.render` proved truthy.

The documentation fix is a genuine one too — `View.extend({ bar })` was shorthand for an undefined identifier, so that example threw a `ReferenceError` if anyone ran it. Same class as the broken snippets corrected in #386 and #399.

### Two things worth noting

**The sibling predicates in this file now differ in rigour.** #393 hardened `isView` to explicit function checks after a truthy-but-invalid value slipped through:

```js
export function isView(view) {
  return typeof view?.render === 'function' &&
    (typeof view.destroy === 'function' || typeof view.remove === 'function');
}

export function isViewClass(ViewClass) {
  return ViewClass.prototype?.render && (ViewClass.prototype.destroy || ViewClass.prototype.remove);
}
```

`isViewClass` would still accept a class whose prototype carries a truthy non-function `render`, which is the exact shape #393 rejected one function above. Not reachable in practice — a class with a non-callable `render` on its prototype is pathological, and the only caller is `collection-view.js:639` — so this is consistency rather than a defect. But two adjacent predicates answering the same question at different strictness invites the assumption that both are equally strict, and this PR is the natural moment to align them.

**This will conflict with #409.** That PR moves `modules/common/view.js` to `src/modules/common/view.js` as a 100%-similarity rename, and this PR edits the file at its old path. Whichever merges second needs the other's change reapplied — trivially, since both are one-liners, but worth knowing before it happens rather than discovering it in a conflict. #409 is currently held on a separate predicted gate failure, so the ordering is not settled.

paulfalgout:
Thanks for checking both cases. I am retaining the existing constructor predicate in this focused fix: checking a missing prototype is required for ordinary supported factory syntax, while validating non-callable prototype methods would change accepted-constructor behavior for a separate edge case. There is no demonstrated consumer failure here that warrants expanding the hot-path check.

The source move is being coordinated with this PR and the prerequisite in #411. The later branch will merge the current base normally so the one-line fix follows the moved file; no forwarding source file is needed.

PR #416, opened 2026-09-05
refactor(source): clarify internal names and lifecycle comments

paulfalgout:
Related #329

## What

- Give the internal Events unbinding callback and Radio channel-log helper descriptive names.
- Clarify three source comments, including the once wrapper's unbind-before-callback ordering.
- Correct Application `getView()` documentation: a root shown while stopped, including before startup, remains available until its ownership is released.

## Why

Keep terminology and comments consistent with what the source does. The existing getView prose incorrectly treated stopped state alone as releasing the coordinated root.

## Scope

Three core source files and Application documentation. No runtime algorithm, lifecycle guard, public API, type, configuration or dependency changes. The later Events TypeScript conversion is separate.

## Deployment Impact

No forms or applications require redeployment for this library source/documentation polish. No release or publication is included. The minified core bundle and all optional executable outputs remain byte-identical.

## Testing

Fresh against master `3702ea29`:

- Base and candidate `npm run prepare` passed, including TypeScript/private consumers, build and distribution checks.
- `vitest run test/unit/events-parity.spec.js test/unit/events-iteration.spec.js test/unit/mixins/events.spec.js test/unit/radio.spec.js test/unit/radio-composition.spec.js test/unit/mixins/radio.spec.js test/unit/requests.spec.js test/unit/utils/once-wrap.spec.js test/unit/application-root-view.spec.js --reporter=dot`: 173 tests passed across nine files.
- `eslint src/mixins/events.js src/modules/radio.js src/utils/once-wrap.js --max-warnings=0`, `npm run docs:check`, and `git diff --check` passed. Documentation validation covered 23 executable examples and 505 internal links.
- Exact-base comparison: 15 of 18 executable outputs and all 15 declarations are byte-identical. The three unminified core files differ only in the intended comments/internal names (+27 raw bytes each); normalized comparison confirms no other code differences.
- Full coverage, browser, package fixture, release and timing suites were not repeated for comment/internal-name changes.

---
## Summary by cubic
Renames the Events unbinding callback and Radio channel-log helper to match what they do, and fixes `Application#getView()` docs: a root shown while stopped now stays available until ownership is released. Related to #329. No runtime or API changes — the minified bundle stays byte-identical.

<sup>Written for commit d9610782f7a057fbd95acfd2baede83311e543a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approved** — reviewed at fetched SHA `d9610782` (base `3702ea29`).

The `getView()` correction is the substantive part, and I checked it against the implementation rather than the description. `getRootView` contains **no lifecycle-state test at all**:

```js
function getRootView(application) {
  const view = application._view;
  if (view && application._region?.currentView !== view) {
    clearRootView(application);
    return;
  }
  return view;
}
```

The only thing that yields `undefined` is the host no longer holding that View — emptied, detached, replaced — or `_view` having been cleared by teardown. So the old sentence, *"or the Application is stopped or destroyed, this method returns `undefined`"*, was wrong about `stopped` and right about `destroyed` only incidentally, because destruction releases the reference rather than because the state is consulted.

The replacement attributes the outcome to the right cause — *"or Application teardown releases it"* — and separates the case the old prose obscured: a View shown **while stopped, including before startup**, is coordinated and returned. That is the reading someone would have gotten wrong, and expensively: believing `getView()` is state-gated invites either a defensive `isRunning()` check that is never needed, or the more damaging assumption that a stopped Application has already let go of a root View it is still holding.

Splitting one sentence into two also separates "what is returned" from "when it is not," which is worth the extra line in a method whose answer depends on host state rather than its own.

The internal renames and the once-wrapper comment about unbind-before-callback ordering are consistent with what the code does; no runtime change in any of them.

PR #421, opened 2026-09-05
fix(diagnostics): inspect erased TypeScript error expressions

paulfalgout:
## What

- Inspect erased TypeScript wrappers around diagnostic error constructors and native throw expressions.
- Add regression coverage for missing and uncataloged codes, valid cataloged codes, and wrapped native errors.

## Why

A cast such as `new (FrameworkError as unknown as new (options: object) => Error)(...)` hid the constructor from diagnostic validation. The checker could accept missing or unknown diagnostic codes, and similarly miss deliberately thrown native errors. Type annotations do not change the executed error expression, so they should not bypass the existing rules.

## Scope

Only `scripts/diagnostics/catalog.mjs` and its unit tests. The checker unwraps `as`, angle-bracket assertions, non-null assertions, `satisfies`, and type instantiation at the existing constructor and throw boundaries. Runtime source, the diagnostic catalog, public APIs and build configuration are unchanged. Existing indirect/dynamic constructor detection limitations remain outside this fix.

## Deployment Impact

No forms, bundles or application deployment required. This affects repository validation only; it does not change shipped runtime code or require a library release.

## Testing

Freshly validated on base `7f7684fa` with candidate `7398f7e4`:

- Before-fix reproduction on previous base `bb369961`: 4 failing groups / 38 passing tests. The refreshed base leaves the checker unchanged.
- `node_modules/.bin/vitest run test/unit/config/diagnostic-catalog.spec.js --reporter=dot`: 42 passed after the fix.
- `npm run check:diagnostics`: validated all 39 production catalog entries.
- `npm run test:source`: passed.
- `node_modules/.bin/eslint scripts/diagnostics/catalog.mjs test/unit/config/diagnostic-catalog.spec.js --max-warnings=0`: passed.
- `git diff --check marionettejs/master HEAD`: passed.

Full runtime coverage and browser tests were not run for this validation-only change. The reviewed patch was transferred without functional changes.

---
## Summary by cubic
Previously, TypeScript casts and assertions around error constructors and throw expressions hid diagnostic errors from the catalog checker, so missing or unknown codes and native throws could pass. The checker now unwraps `as`, angle-bracket assertions, non-null assertions, `satisfies`, and type instantiation before validating constructors and throws. This is validation-only; no runtime code or catalog entries change.

- Adds regression tests for missing and uncataloged codes, valid cataloged codes, and wrapped native errors.
- Changes only `scripts/diagnostics/catalog.mjs` and its unit test.

<sup>Written for commit 7398f7e431ae42eaea19339f2c52736b2bcc7757. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/421?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/diagnostics/catalog.mjs
+++ b/scripts/diagnostics/catalog.mjs
@@ -174,6 +174,14 @@ function parseJavaScript(contents, path, errors) {
   return sourceCode.ast;
 }
 
+function unwrapTypeScriptExpression(node) {
+  while (node && ['TSAsExpression', 'TSTypeAssertion', 'TSNonNullExpression',
+    'TSSatisfiesExpression', 'TSInstantiationExpression'].includes(node.type)) {
+    node = node.expression;
+  }
+  return node;
+}
+
 function addRuntimeSourceErrors(runtimeSources, diagnosticsByCode, errors) {
   for (const { contents, path } of runtimeSources) {
     const ast = parseJavaScript(contents, path, errors);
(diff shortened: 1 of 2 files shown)


paulfalgout:
**Approved** — reviewed at fetched SHA `7398f7e4` (base `7f7684fa`).

I checked whether this was closing a live bypass or a latent one: no cast of that shape exists in the converted `.ts` sources today, so nothing is currently escaping validation. That makes this preventive — and mid-conversion is exactly when to do it, because every remaining slice is an opportunity for such a cast to appear, and its failure mode is silence rather than an error. A diagnostic constructor that stops being recognised does not fail the checker; it simply stops being checked.

**The `while` loop is load-bearing**, not defensive styling:

```js
while (node && ['TSAsExpression', …].includes(node.type)) { node = node.expression; }
```

The idiom in your own example — `FrameworkError as unknown as new (options: object) => Error` — is **two** nested `TSAsExpression` nodes, because `as unknown as T` is the standard way to force an otherwise-rejected cast. A single unwrap would have handled `x as T` and missed precisely the case that motivated the fix.

The node list looks complete for type-only expression wrappers: `as`, angle-bracket assertion, non-null `!`, `satisfies`, and instantiation expressions cover everything TypeScript erases at an expression position. Since they all erase, unwrapping them is not an approximation — the checker now sees the expression that actually executes, which is the correct invariant to hold.

Applying it at **both** boundaries matters too: `node.argument` and `argument?.callee` for throws, `node.callee` for construction. Fixing only the `MarionetteError` side would have left `throw new (TypeError as …)(…)` able to slip a native error past the no-native-throws rule, which is the same class of hole from the opposite direction.

Scoping out indirect and dynamic constructor detection — `const E = MarionetteError; throw new E(…)` — is honest, and a different problem: that one needs binding analysis rather than node unwrapping.

PR #424, opened 2026-09-05
perf(region): skip teardown lookup during live empty and reset

paulfalgout:
## What

- Skip the teardown-authorization WeakMap lookup in live Region `empty` and `reset` calls.
- Preserve the existing authorization and public override paths during Region destruction.

## Why

Teardown authorization is created only after a Region begins destruction. A live Region cannot hold that authorization, so looking it up on every live empty/reset call does unnecessary work. Existing lifecycle flags already distinguish the two cases.

## Scope

Only `src/modules/region.js`, with 8 additions and 6 deletions. This is the previously reviewed optimization without additional guards or behavior changes. DOM element resolution, reset semantics, lifecycle ordering, public empty/reset overrides and terminal reentry/error handling remain unchanged. It does not introduce before:show or before:render cancellation.

## Deployment Impact

No forms or bundles require redeployment for this library-only change. Consumers receive it through a future Marionette upgrade; this PR performs no release or deployment. Optional package output is unchanged.

## Testing

Fresh on base `26aee8ce`, candidate `f47d122e`:

- Vitest on the existing Region lifecycle/Region, View lifecycle/children/dynamic Regions, CollectionView lifecycle/empty, and Application root/ownership suites: 360 tests across nine files passed, including override/reentry/throw/retry behavior.
- `npm run build` on the exact base and refreshed candidate, including source/private and generated consumer type checks: passed. Reused the owner-release candidate build after proving its whole Git tree matches merged master and all 31 previously recorded output hashes remain unchanged, then extending verification recursively to the four nested jQuery files.
- `node_modules/.bin/eslint src/modules/region.js --max-warnings=0`, `npm run test:source`, `npm run test:dist`, and `git diff --check 26aee8ce HEAD`: passed. Full lint passed on the immediately preceding refresh; the Region patch remains identical.
- Recursive exact-base comparison of 35 distribution files: all 14 optional executables (including dom/jquery ESM/CJS) and 15 optional declarations identical. Core CJS/ESM raw deltas -11/-11 bytes; minified/unminified UMD +2/-7. Brotli quality11 deltas are +17/+1/-4/+8 bytes respectively.

No fresh timing run was performed. Earlier isolated lookup timings are historical; mixed composite timings do not support a real-application speed claim. Browser, full core coverage and package-fixture suites were not repeated for this unchanged source optimization.

---
## Summary by cubic
Skips the teardown-authorization lookup in live Region `empty` and `reset` calls, since that authorization is only created once destruction begins. The lookup previously ran on every call; it now runs only when the region is destroying or destroyed, preserving the existing destruction paths and public overrides.

This is a library-only change with no behavior or API impact, so no redeployment is needed; consumers pick it up in a future Marionette upgrade.

<sup>Written for commit f47d122e5e54d1c612d1030e85ec415937c0d035. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approved** — reviewed at fetched SHA `f47d122e` (base `26aee8ce`).

I worked the four cases rather than the description, since this converts an unconditional call into a short-circuited one and `consumeDestroyTeardown` **has a side effect** — it deletes the token it matches.

| Region state | Before | After |
|---|---|---|
| live | consume called (no entry, returns false), `false \|\| true` → proceed | `canMutateRegion` true → short-circuits, consume never called → proceed |
| destroying, token matches | consume returns true → proceed | consume evaluated and returns true → proceed |
| destroying, no match | consume false, cannot mutate → return | consume false → return |

Equivalent throughout, and the one-shot token is still consumed on the path that needs it. The `reset` variant preserves the later `if (authorized)` behaviour too: a live Region leaves `authorized` false, so `destroyTeardown.set(this, 'empty')` stays skipped exactly as it was.

**The invariant the optimization rests on is real.** I checked the ordering in `destroy` rather than taking it: `_isDestroying = true`, then `_isDestroyed = true`, and only afterwards `destroyTeardown.set(this, 'reset')`. The second `set(this, 'empty')` in `reset` is likewise reachable only under `authorized`, which requires a token, which requires destruction already underway. So a live Region genuinely cannot hold authorization, and skipping the lookup for it cannot change behaviour.

### Worth a note where it is easy to break

That invariant is now load-bearing rather than incidental: any future `destroyTeardown.set` occurring while a Region is still live would be **silently ignored** by `empty` and `reset`, because they no longer look. The old code would have honoured it. That is the same shape as the `_isDestroyed`-before-`disposeAll` ordering I flagged on #389 — correct today, and quiet if a reasonable-looking edit inverts it.

The `destroy` method's flag assignments are far from the `canMutateRegion` helper that now depends on their ordering, so a line at one end pointing at the other would make the coupling visible to whoever edits either.

PR #430, opened 2026-09-05
fix(behaviors): preserve cleanup during initialization

paulfalgout:
Related #429

## What

- Stop Behavior construction from attaching DOM callbacks after initialization or a synchronous state subscription destroys the Behavior.
- Register live Behaviors in the host's current collection so earlier destruction cannot lose later siblings or rollback cleanup.
- Add seven public regressions covering self-destruction, sibling and nested ownership, subscription cleanup, and constructor failure.

## Why

A Behavior could reacquire DOM listeners after destruction. Its destruction also replaced the host collection while parsing still used the old array, leaving live siblings unreachable during host cleanup. Clicking a retained root after host destruction could still call those handlers. This predates the typing conversion in #429.

## Scope

Two core Behavior construction files and one regression suite. Nested definitions remain part of the host's flat Behavior collection, even when their declaring Behavior destroys itself. The fix adds two checks per constructed Behavior, using the existing destroyed flag. Render, update and attachment loops are unchanged; this does not establish general cancellation semantics for lifecycle callbacks.

## Deployment Impact

No forms or application redeploy is required for this repository change. Package consumers receive it with a future library version. No release, tag, or publication is included.

## Testing

- `npm run check:types` and focused ESLint on the changed source and regression suite.
- `npm run build`.
- Focused Behavior tests: 102 tests passed.
- `npm run coverage`: 2,179 tests passed, with 100% statements, branches, functions and lines; 19 agent benchmark contract tests passed.
- Five new regressions fail against the unchanged base; the ordinary-order control already passes. The additional synchronous state regression fails before the constructor check is moved after state initialization. An independent public StateApi reproduction confirms the callback leak and its correction.
- Independent source/test review verified retained-root DOM cleanup and the nested ownership contract.

Fresh full-core output comparison against master: Brotli ESM -20 bytes, CJS -9 bytes, UMD -13 bytes, minified UMD +6 bytes. The earlier two-check prototype had overlapping construction timings within 1%; that is bounded prototype evidence, not an exact-final-commit speed claim.

Fresh CI will validate distributions, browsers, installed packages and deterministic size against the current master base.

---
## Summary by cubic
Fixes Behavior cleanup during initialization so a destroyed Behavior can no longer rebind DOM listeners, and live siblings stay reachable when the host is destroyed.

- Checks the destroyed flag after initialization and state event setup, and only registers live Behaviors in the host's current collection.
- Adds regression tests covering self-destruction, sibling and nested ownership, constructor failure, and synchronous state subscription cleanup.
- Changes apply to Behavior construction and parsing only; render, update, and attachment loops are unchanged.

<sup>Written for commit 85b28d29058fd25841ad2178115003c8c9a92c98. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/430?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Behaviors that destroy themselves during initialization are no longer rebound or synchronized.
  * Nested behaviors are handled correctly even when a parent behavior is destroyed during setup.
  * Remaining active behaviors continue to initialize and clean up reliably after a sibling is destroyed or fails during construction.

### DIFF
--- a/src/mixins/behaviors.ts
+++ b/src/mixins/behaviors.ts
@@ -4,6 +4,7 @@ import disposeAll from '../utils/dispose-all.ts';
 import getValue from '../utils/get-value.ts';
 
 export interface BehaviorInstance {
+  _isDestroyed?: boolean;
   behaviors?: unknown;
   destroy(options?: unknown): unknown;
   _syncElement(): unknown;
@@ -59,29 +60,29 @@ function getBehaviorClass(options: BehaviorDefinition) {
   });
 }
 
-function addBehavior(view: BehaviorContainer, behaviorDefinition: BehaviorDefinition, allBehaviors: BehaviorInstance[]) {
+function addBehavior(view: BehaviorContainer, behaviorDefinition: BehaviorDefinition) {
   const { BehaviorClass, options } = getBehaviorClass(behaviorDefinition);
   const behavior = new (BehaviorClass as BehaviorConstruction)(options, view);
-  allBehaviors.push(behavior);
+  if (!behavior._isDestroyed) {
+    view._behaviors!.push(behavior);
+  }
 
-  parseBehaviors(view, getValue(behavior, 'behaviors'), allBehaviors);
+  parseBehaviors(view, getValue(behavior, 'behaviors'));
 }
 
 // Iterate over the behaviors object, for each behavior
 // instantiate it and get its grouped behaviors.
 // This accepts a list of behaviors in either an object or array form
-function parseBehaviors(view: BehaviorContainer, behaviors: unknown, allBehaviors: BehaviorInstance[]) {
+function parseBehaviors(view: BehaviorContainer, behaviors: unknown) {
   if (Array.isArray(behaviors)) {
     for (let index = 0, length = behaviors.length; index < length; index++) {
-      addBehavior(view, behaviors[index], allBehaviors);
+      addBehavior(view, behaviors[index]);
     }
   } else {
     eachOwn(behaviors, (behaviorDefinition: BehaviorDefinition) => {
-      addBehavior(view, behaviorDefinition, allBehaviors);
+      addBehavior(view, behaviorDefinition);
     });
   }
-
-  return allBehaviors;
 }
 
 function eachBehavior(behaviors: BehaviorInstance[] | undefined, iteratee: (behavior: BehaviorInstance) => unknown) {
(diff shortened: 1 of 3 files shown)


paulfalgout:
**Approved** — reviewed at fetched SHA `85b28d29` (base `d7a94c93`).

### The guard placement is the detail that matters

```ts
this._initStateEvents();
if (this._isDestroyed) { return; }

this._syncElement();
```

Placing the check **after** `_initStateEvents` rather than only after `initialize` is what makes it complete. A state subscription can destroy its owner synchronously before returning — that is exactly the window #406 closed for `StateMixin` — so a guard sitting between `initialize` and `_initStateEvents` would have caught the self-destroying `initialize` and missed the self-destroying subscription. Both paths reach `_syncElement`, which is what attaches the DOM callbacks, so both had to be covered.

The symptom is worth noting because it is not a quiet one: a retained root element still firing handlers on click after the host was destroyed. That is a leak with user-visible behaviour rather than just retained memory.

### The captured-array bug is real, and line 126 is the proof

```ts
_rollbackBehaviors() {
  rollbackBehaviors(this._behaviors || []);
  this._behaviors = [];      // ← replaces the array
}
```

The old `parseBehaviors(view, behaviors, this._behaviors)` captured that array once and pushed into it for the whole traversal. If a Behavior's construction triggered a rollback partway through, `this._behaviors` became a **fresh** array while every subsequent sibling was still being pushed into the orphaned one — unreachable to `eachBehavior`, `disposeBehaviors`, and every later cleanup pass. Reading `view._behaviors` at each push resolves the live array instead.

Same shape as the stale-snapshot problems in #415 and #423: a reference captured before user code runs, invalidated by that user code, then used anyway.

I checked the non-null assertion at `view._behaviors!.push(...)`, since that is the one place this could throw: `_behaviors` is assigned `[]` at lines 114 and 126 and reassigned to arrays elsewhere, never deleted or set undefined, so it is always an array by the time `parseBehaviors` runs.

### The nested-behavior decision

Skipping registration of a self-destroyed Behavior while still parsing its nested definitions is deliberate and documented, and it follows from the flat host-owned collection model — nested Behaviors belong to the host, not to the Behavior that declared them, so a parent destroying itself should not silently drop them. Worth having stated, because the opposite reading is equally plausible from the code alone.

PR #431, opened 2026-09-05
docs(guides): lead with application tasks and a consistent voice

paulfalgout:
Related #147

## What

- Give the main README the marketing site's purpose-first voice and restrained humor, with direct routes from application tasks to Marionette classes.
- Refresh current guide introductions and replace the terminology rules table with a practical glossary. Add a checked TypeScript installation example.
- Correct stale package, Application, Region-name, template-data and Underscore claims; identify the existing v3 starter projects as historical references.

## Why

The documentation should help readers find an approach and understand the code they are about to write. It currently foregrounds integration vocabulary and release machinery before ordinary application work. This pass brings its tone closer to the approved marketing copy while keeping API contracts precise and agent-development claims limited to the project's direction.

## Scope

README and 19 current Markdown guides, plus link corrections in the Events and View lifecycle guides. Historical upgrade guides and precise release, performance and diagnostic references remain unchanged. API references outside the existing generated-docs set link to their canonical GitHub source, so navigation works both in the repository and in the rendered guides. No library source, package configuration, website implementation or deployment setting changes.

## Deployment Impact

No forms or application redeploy is required. This changes repository and generated documentation plus the README included in a future package. No package release or site deployment is included; the existing documentation publication gate remains disabled.

## Testing

- `npm run docs:check`: 7 example-contract tests passed; 23 marked examples mapped to 13 fixture validators; 55 pages built; 56 HTML files and 506 internal links validated.
- Executed the two changed introductory JavaScript examples with Node 24/JSDOM: rendering produces `<h1>Contents</h1>`; the Behavior example shows the revised message and removes its host View.
- The installation TypeScript snippet exactly matches the example verified against the installed package with TypeScript 6.0.3 and 7.0.2 under ESM, CommonJS and bundler resolution. This checks `this.el` inside `initialize` and `this.options` inside the View method. Invalid options produce the expected errors; its rendered greeting was verified.
- Existing marked executable examples are unchanged. No runtime suite was rerun locally for prose-only changes; normal PR CI remains enabled.
- Independently reviewed the 20-file voice pass and its reference-link follow-ups for contract accuracy and unsupported claims; verified all five historical starter manifests target `backbone.marionette ~3.0.0`.
- `git diff --check` passed.

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -1,8 +1,9 @@
 # Common Marionette Concepts
 
-This document covers the shared configuration patterns used by Marionette's
-classes. Class-specific pages remain the authority for when a particular value
-is read and whether it is read again.
+Learn the configuration patterns once, then use them across Marionette's classes.
+Each class's reference explains when it reads an option and whether it reads it
+again. For checked application options, see the
+[TypeScript example](./installation.md#typescript).
 
 ## Documentation Index
 
@@ -24,9 +25,9 @@ const view = new View();
 const app = new Application();
 ```
 
-V5 has no default namespace export. The package also exposes explicit optional
-integration subpaths; see [Installing Marionette](./installation.md) for the
-current entrypoints and peer-dependency boundaries.
+V5 has no default namespace export. The separate `@marionette/adapters` package
+provides optional integration subpaths; see [Installing Marionette](./installation.md)
+for the entrypoints and their dependencies.
 
 Existing no-bundler applications may serve the published
 `dist/marionette.umd.js` artifact. It exposes the named API on the global
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -1,19 +1,17 @@
 # Marionette Classes
 
-Marionette follows Backbone's [pseudo-class architecture](./basics.md#class-based-inheritance).
-This documentation is meant to provide a comprehensive listing of those classes so that
-the reader can have a high-level view and understand functional similarities between the classes.
-All of these classes share a [common set of functionality](./common.md).
+Each Marionette class has a job: render a piece of interface, manage where it goes,
+repeat it, share an interaction, or coordinate a feature. Start with the job you
+need, then follow the reference for its options and lifecycle.
+
+The classes share [configuration and inheritance patterns](./basics.md#class-based-inheritance)
+and a [common set of methods](./common.md).
 
 ### [Marionette.View](./marionette.view.md)
 
-A `View` is used for managing portions of the DOM via a single parent DOM element or `el`.
-It provides a consistent interface for managing the content of the `el` which is typically
-administered by serializing a `Backbone.Model` or `Backbone.Collection` and rendering
-a template with the serialized data into the `View`s `el`.
-
-The `View` provides event delegation for capturing and handling DOM interactions as well as
-the ability to separate concerns into smaller, managed child views.
+A `View` owns a piece of interface through its root element, `el`. It renders a
+template, handles DOM interactions, and can divide a screen into Regions for child
+Views. Plain objects and function templates work with the default configuration.
 
 `View` includes:
 - [The DOM API](./dom.api.md)
@@ -29,10 +27,10 @@ A `View` can have [`Region`s](#marionetteregion) and [`Behavior`s](#marionettebe
 
 ### [Marionette.CollectionView](./marionette.collectionview.md)
 
-A `CollectionView` like `View` manages a portion of the DOM via a single parent DOM element
-or `el`. This view manages an ordered set of child views that are shown within the view's `el`.
-These children are most often created to match the models of a `Backbone.Collection` though a
-`CollectionView` does not require a `collection` and can manage any set of views.
+A `CollectionView` manages an ordered set of child Views inside its root element.
+Use it for rows, cards, or other repeated content. A plain array supplies a static
+collection; an observable data integration can notify it of changes. You can also
+manage child Views directly without supplying a collection.
 
 `CollectionView` includes:
 - [The DOM API](./dom.api.md)
@@ -48,8 +46,8 @@ A `CollectionView` can have [`Behavior`s](#marionettebehavior).
 
 ### [Marionette.Region](./marionette.region.md)
 
-Regions provide consistent methods to manage, show and destroy views in your
-applications and views.
+A `Region` gives a View a place to appear. Showing a new View renders and attaches
+it; replacing or emptying the Region destroys its current View by default.
 
 `Region` includes:
 - [Class Events](./events.class.md#region-events)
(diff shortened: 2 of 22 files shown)


paulfalgout:
**Approved** — reviewed at fetched SHA `ecca76fe` (base `6c2d09c7`).

A voice pass over documentation is easy to get wrong by rounding precise claims into friendly ones. This one didn't. I spot-checked the load-bearing new and rewritten claims against the source rather than the prose, and they hold.

### Verified against source

**`template: false` preserves text set during initialization** — `render()` returns at `src/modules/view.ts:585` on `template === false`, before `_renderTemplate`. The example's `view.render()` is a genuine no-op, so `.el` still carries the `textContent` from `initialize`.

**`this.el` is available in `initialize`** — the constructor calls `setElement(this._getEl())` before `initialize`, so the example isn't relying on an accident of ordering.

**The readiness / notification distinction survived the rewrite intact**, which is the claim I most expected to get blurred. "The corresponding `before:*` event listeners are synchronous notifications; their return values are not awaited" is exactly right, and for a non-obvious reason: `triggerMethod` returns *only* the `onBeforeStart` method's result and discards what `trigger` gives back, so `await application.triggerMethod('before:start', …)` awaits the hook and not the listeners. Two things dispatched by one call, one of them awaited. Worth keeping stated as precisely as it is here.

### On "a checked TypeScript installation example"

Mostly earned. `test/fixtures/core-types/consumer.mts:14-17` is the same pattern under `strict: true`, and the `@ts-expect-error` assertions at lines 58 and 60 encode precisely the two claims the new section makes — missing options and wrong-typed values are compile errors.

One construct in the example has no fixture analogue: the docs use `this.el` **inside `initialize`**, and the fixture's `initialize` only ever touches its `options` parameter, never `this`. `this` typing inside a method of an `.extend` literal is the fragile part of that pattern, so a fixture line covering it would make the "checked" claim complete rather than nearly complete.

### Dead links in two files this PR touches

Pre-existing, not introduced here — but the stated scope includes making navigation work in both the repository and the rendered guides, and these are in the touched set:

`docs/events.class.md` — eight links to `./collectionview.md`. No such file; the correct name is `./marionette.collectionview.md`, used correctly at line 318 of the same file, a few lines below the broken ones. Fixing the filename repairs `#adding-a-child-view` and `#removing-a-child-view` (headings at 815 and 881) and `#defining-the-viewcomparator` / `#defining-the-viewfilter` (961, 1157).

`#collectionviews-children` needs a target regardless — no heading generates that anchor, so it is broken in the five wrong-filename links *and* in the correctly-spelled one at line 318.

`docs/marionette.region.md:556` — `./upgrade.md#changes-to-regionshow`. There is no `docs/upgrade.md`, and no heading producing that anchor in `upgradeGuide.md`, `docs/upgrade-v3-v4.md`, or `docs/upgrade-v2-v3.md`. The nearest surviving material is `docs/upgrade-v3-v4.md:175`, on `preventDestroy` rather than on leftover HTML, so this one needs a decision about intent rather than a mechanical retarget.

None of these block the merge; they are the kind of thing worth catching while the files are already open.

paulfalgout:
The reference links are corrected in a847e7be: all 10 CollectionView targets now resolve to existing headings, and the Region note retains its current behavior explanation without pointing at a nonexistent upgrade page. For the typing evidence, the exact documentation snippet—including this.el inside initialize and this.options in message()—was separately compiled against the installed package on TS 6.0.3 and 7.0.2 in ESM, CJS and bundler modes, then executed in JSDOM. The existing fixture is not the only evidence behind that claim; the PR description now makes the exact-snippet coverage explicit.

paulfalgout:
**Still approved** — re-reviewed the update at fetched SHA `a847e7be` (previously `ecca76fe`).

All nine links are resolved. I re-swept every relative `.md` link in `readme.md` and `docs/` and there are now zero broken file targets. `#accessing-a-child-view` is a real heading (`docs/marionette.collectionview.md:617`), and it is a better target than the `#collectionviews-children` anchor I was chasing — that one never existed, so retargeting rather than inventing a heading was the right call.

Two things I did not flag that you fixed anyway: "removed **to** the `children`" → "removed **from**", and the `marionette.region.md` link, where dropping the link and keeping the prose is the right resolution. That link's anchor described leftover HTML while the nearest surviving upgrade material is about `preventDestroy` — there was no correct target to point at, so pointing at nothing is honest.

### Seven dead anchors remain (all pre-existing)

Since the file-level sweep is now clean, the anchor-level one is what's left. I checked each against the base: **none were introduced by this PR**, and two of the four files are outside your touched set.

| Link | In | Actual heading |
| --- | --- | --- |
| `./marionette.view.md#showing-a-view` | `events.class.md`, `view.lifecycle.md` | `### Showing a Child View` (`:232`) |
| `./events.entity.md#model-events` | `events.md`, `marionette.behavior.md` | file was restructured — no per-entity headings survive |
| `./events.entity.md#collection-events` | `events.md`, `marionette.behavior.md` | same |
| `./marionette.state.md#owned-state` | `marionette.application.md` | `## Borrowed and owned sources` (`:26`) |

The two `marionette.view.md` and the `marionette.state.md` ones are mechanical retargets. The `events.entity.md` pair is not: that file now organizes around `Handler ownership and arguments`, `Event-map names`, and `Backbone entities` rather than by entity type, so both links need a judgment call about what they were meant to point the reader at.

Entirely reasonable to leave all of these for a separate pass — `events.md` and `view.lifecycle.md` are outside this PR's scope, and splitting the remainder would just scatter the fix across two PRs.

paulfalgout:
Included the seven remaining anchors in 45d01fb5 so this link cleanup can close out together. showChildView and owned-state links now use the actual current headings. Both modelEvents and collectionEvents link to the shared Entity events guide, which now explains their common subscription and handler contract. The two extra files contain link-only corrections. The target headings and documentation build/link checks pass.

paulfalgout:
**Approved** — re-reviewed at fetched SHA `45d01fb5` (previously `a847e7be`).

All seven anchors are resolved. I re-ran both sweeps across `readme.md` and every file in `docs/`: **zero broken file targets and zero dead anchors**, where the starting point was nine broken links and seven dead anchors.

The `events.entity.md` pair was the one that needed a judgment call, and dropping to the file root is the right answer. Inventing `#model-events` / `#collection-events` headings to satisfy the old links would have re-imposed the per-entity structure that file deliberately moved away from — the reader arriving from `modelEvents` now lands on a document organized around handler ownership and event-map names, which is what actually explains the option. A slightly less precise link beating a precise link to a structure that no longer exists.

`#showing-a-child-view` and `#borrowed-and-owned-sources` are exact matches for their targets.

Nothing outstanding from my side.

PR #433, opened 2026-09-06
fix(perf): scope consumer peers to measured scenarios

paulfalgout:
Related #127

## What
- Keep the frozen v1 consumer benchmark external allowlist at Backbone and jQuery while permitting unrelated optional peers in package manifests.
- Reject missing canonical peers, extra required peers, modified v1 authority, and renderer imports leaking into existing benchmark scenarios.

## Why
Adding a separately imported optional renderer currently invalidates all 18 existing consumer measurements solely because its peer is declared. Unused optional integrations must not change the measured scenarios or their authority.

## Scope
The consumer benchmark validation and its regression tests. No production runtime, frozen fixture, baseline, ceiling, compiler, or reporting format changes.

## Deployment Impact
No application or form redeployment. No package publication or release tags. This prepares validation for optional renderer subpaths.

## Testing
- `node --test test/performance/consumer-bundles.test.mjs`: 11 passing tests, including a real injected Morphdom import rejected by measured graph checks.
- Targeted ESLint and `git diff --check` pass.

---
## Summary by cubic
Scopes consumer peer validation to the frozen v1 benchmark scenarios so unused optional integrations no longer invalidate all 18 measurements. Previously, declaring any additional peer dependency failed every consumer bundle check; now only the canonical `backbone` and `jquery` peers are enforced, while unrelated optional peers are allowed as long as they aren't imported into measured scenarios.

**Validation**
- Rejects missing canonical peers, extra required peers, modified v1 peer authority, and renderer imports in benchmark scenarios.
- Adds regression tests covering optional peer declarations and rejected imports.

<sup>Written for commit a503d2d22cb51e2783f172dd50be42098fc4a29a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/433?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consumer bundle validation to consistently enforce the supported peer-dependency contract.
  - Optional peer dependencies can now be added without affecting established bundle scenarios.
  - Required dependencies outside the supported contract are now flagged, including undeclared or unexpectedly imported dependencies.

- **Tests**
  - Added coverage for missing required dependencies, unsupported dependencies, optional peers, metadata mismatches, and unexpected external imports.

### DIFF
--- a/scripts/performance/bundle-size.mjs
+++ b/scripts/performance/bundle-size.mjs
@@ -356,8 +356,22 @@ export function validateConsumerBundleContract(
     .filter(peer => !peer.startsWith('@types/'))
     .filter(peer => !internalPackages.has(peer))
     .sort();
-  if (!sameStringInventory(contract.peerExternalImports, runtimePeers)) {
-    violations.push(`Consumer bundle peer externals must be ${runtimePeers.join(', ')}`);
+  if (!sameStringInventory(contract.peerExternalImports, consumerPeerExternalImports)) {
+    violations.push(`Consumer bundle peer externals must be ${consumerPeerExternalImports.join(', ')}`);
+  }
+  const missingPeers = difference(consumerPeerExternalImports, runtimePeers);
+  if (missingPeers.length) {
+    violations.push(`Consumer bundle peers are not declared runtime peers: ${missingPeers.join(', ')}`);
+  }
+  // Other integrations may add optional peers without changing the frozen v1
+  // scenarios. Exact measured graph checks still reject their runtime imports.
+  const additionalRequiredPeers = [...new Set(packageJsons.flatMap(manifest => {
+    return Object.keys(manifest.peerDependencies || {}).filter(peer =>
+      runtimePeers.includes(peer) && !consumerPeerExternalImports.includes(peer) &&
+      manifest.peerDependenciesMeta?.[peer]?.optional !== true);
+  }))].sort();
+  if (additionalRequiredPeers.length) {
+    violations.push(`Runtime peers outside consumer bundle v1 must be optional: ${additionalRequiredPeers.join(', ')}`);
   }
   if (!isDeepStrictEqual(contract.toolchain, consumerToolchain)) {
     violations.push('Consumer bundle toolchain metadata is not canonical');
--- a/test/performance/consumer-bundles.test.mjs
+++ b/test/performance/consumer-bundles.test.mjs
@@ -162,6 +162,74 @@ describe('consumer bundle measurements', () => {
     }
   });
 
+  test('keeps canonical peers declared and rejects required peers outside v1', async() => {
+    const { brotliQuality, contract, fixture, fixtureRevision, packageJson, packageJsons } = await canonicalInputs();
+    const adapters = packageJsons[1];
+    delete adapters.peerDependencies.backbone;
+    assert.match(validateConsumerBundleContract(
+      contract, fixture, packageJson, brotliQuality, fixtureRevision, packageJsons,
+    ).join('\n'), /not declared runtime peers: backbone/);
+
+    adapters.peerDependencies.backbone = '^1.4.0';
+    adapters.peerDependencies.morphdom = '^2.7.8';
+    delete adapters.peerDependenciesMeta.morphdom;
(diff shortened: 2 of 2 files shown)


paulfalgout:
**Approved** — reviewed at fetched SHA `cc300dfb`, diffed against merge-base `066d46f4` on `refactor/finish-typescript` (not `master`; this is stacked on #432).

The change replaces a derived equality with a frozen constant, which is normally a downgrade — a hardcoded allowlist drifts from reality silently. Here it isn't, because the two new checks pin the constant from both directions:

- `missingPeers` rejects a canonical peer that no manifest actually declares, so `['backbone', 'jquery']` cannot outlive the declarations it describes.
- `additionalRequiredPeers` rejects any *required* runtime peer outside v1, so the frozen set cannot quietly acquire a neighbor.

That pairing is what makes freezing safe, and it is the right shape for the actual problem: declaring an optional peer is not evidence that the measured scenarios use it, so it should not invalidate 18 measurements.

### The third assertion in the first test is the sharp one

```js
adapters.peerDependenciesMeta.morphdom = { optional: true };
packageJson.peerDependencies = { ...packageJson.peerDependencies, morphdom: '^2.7.8' };
```

`canonicalInputs()` returns `packageJsons: [packageJson, adaptersPackageJson]` — the same object — so this asserts that optional in the adapters manifest while required in the root manifest is *still* rejected. That is the correct semantics (any manifest requiring it makes it a real requirement) and it is exactly the case a reader skims past, since the first two assertions already look like full coverage.

### The third test earns the claim rather than asserting it

Injecting a real `import 'morphdom'` into the built `packages/adapters/dist/dom/jquery.js` and re-measuring is what distinguishes this from a validator unit test. It shows the permissive path staying permissive only while the frozen scenarios genuinely do not import the peer — 18 artifacts, `['backbone', 'jquery']`, no violations — and then both the fixture-metadata and non-peer-external violations firing once it does.

Worth noting the layering that leaves no gap here, since it is easy to look at `additionalRequiredPeers` and think it should also police `dependencies`: it governs *declared peers* only, an undeclared import surfaces as `non-peer external imports` in the graph check, and a bundled dependency surfaces as size against the ceilings. Three separate layers, so a fourth check inside this one would be redundant rather than defensive.

### Merge ordering

Since CI runs the *base* checkout's validator, this permission has to land before the renderer work that consumes it — which is what "prepares validation for optional renderer subpaths" says, and matches the base-authority pattern from the earlier budget PRs. Order is #432, then this, then the renderer change.

One thing to watch: the merge-base is `066d46f4`, so this branch predates #432's latest head (`9e96fe32`). Nothing in that update touches `scripts/performance/`, so there is no conflict — just worth a rebase before merge so the stack measures against what actually lands.

`Build and validate` and `Release preflight` pass at this SHA; the rest were pending when I looked, with no failures.

paulfalgout:
**Still approved** — re-reviewed at fetched SHA `a503d2d2` (previously `cc300dfb`).

This is the rebase onto `master` now that #432 has landed, and it came out exactly right: the merge-base is now `6c01b573` itself, and the two-dot and three-dot diffs are identical — both showing only `scripts/performance/bundle-size.mjs` and `test/performance/consumer-bundles.test.mjs`. The seven files that appear when diffing against the old head are #432's tail arriving through the rebase, not new work, so nothing carried forward and nothing can revert master on merge.

The validator and test changes I reviewed at `cc300dfb` are unchanged. `Bundle size` now passes at this SHA, which is the check that most directly exercises the change; `Build and validate`, `Analyze`, and `Performance timing` also pass, with the rest pending and no failures.

One consequence worth flagging: **#434 is based on `fix/consumer-peer-scope` and still points at the pre-rebase commit**, so it now carries a duplicate of #432's tail relative to master. I verified those seven files are byte-identical to what landed, so there's no divergence risk — but the same rebase there would make its diff as clean as this one is now.

PR #436, opened 2026-09-07
refactor(data): simplify sources and share common utilities

paulfalgout:
## What

- Simplify native Model/Collection: remove `touch`, `swap`, `replace`, model-owner tracking, ID freezing, and observer coalescing. Keep `move` for explicit ordering, use stable `cid` keys, and expose plain-data copies through `toObject` / `toArray`.
- Share option, binding, inheritance, assignment, and error helpers through typed `@marionette/utils`. Core and data use the same module implementation; standalone UMD remains self-contained. Update package, release, and bundle-measurement tooling for the fourth package.
- Remove Redux, Zustand, and XState Store adapters and simplify the retained actor snapshot helper. Prove immutable Redux integration through the existing seven public DataApi methods, including falsy source references and retained editable child Views.

## Why

The native package had accumulated behavior beyond the simple Backbone-like source it needs to be. Ordinary synchronous events provide the required notifications without a second observer system. The removed store adapters recreated children on normal immutable edits and did not provide satisfactory model observation; the maintained public-contract example proves extensibility without shipping those implementations.

Shared helpers remove duplicated implementations and give generic components a supported import surface. Rendering reads current attributes independently of the plain-data copy methods.

## Scope

Core source imports and opaque-source handling; native data; optional data adapters; shared utilities; declarations, tests, documentation, package/release tooling, and performance accounting. DOM adapter contracts are unchanged. No compatibility aliases remain for removed alpha APIs. Nested structural mutation during synchronous collection notification is outside the supported contract, matching the observed Backbone limitation.

## Deployment Impact

No forms require redeployment. This PR changes library packages and their build/release tooling; no npm version, tag, or release is published. Publication remains disabled. Future publication processes utils before core, data, and adapters.

## Testing

- `npm run build` — all four packages, source checks, generated declarations, and core consumer typing pass.
- `npm run coverage` — 1,998 tests across 117 files; 100% statements, branches, functions, and lines. Also passes 19 agent-benchmark contract tests.
- `npm run test:browser` — Chromium, Firefox, and WebKit, including packed data/adapters and all DOM adapter contracts.
- `npm run test:dist` and `npm run test:source` — module/standalone formats and source graph pass.
- `npm run test:fixtures` — all 35 packed consumers pass CI on Linux, macOS, and Windows, including TypeScript 6/7, shared core/utils error and inheritance identity, and isolated DOM-adapter consumers.
- `npm run test:performance` — 63 tests; real consumer measurements include utils bytes, with no arbitrary alpha size ceiling.
- `npm run test:release-promotion` — 11 tests; release profile and promotion preflight pass with publication disabled. Full clean-checkout release artifact construction runs in CI.
- `npm run docs:check` — 23 executable examples, 56 HTML files, 506 links. Diagnostic catalog validation passes.
- `npx eslint . --ignore-pattern '.claude/**' --max-warnings=0` and `git diff --check` pass. The ignore excludes unrelated nested Claude checkouts from the local lint walk.

Claude reviewed the plan and actual diff. Its valid destroy-options finding was fixed with regression tests. The serialization concern was checked against the existing Backbone adapter's direct-attributes contract; the questioned utils publication output is generated by the updated release target script.


paulfalgout:
**Approved** — reviewed at fetched SHA `78da050d` (base `8f1aa2a7`). 172 files; I read the contract changes, the data semantics, and the upgrade guide rather than every file.

### This closes a measurement gap that has been open a long time

`packages/data/dist/index.js` and `.cjs` were **never in the performance contract** — I checked master, and the count is zero. This PR adds them alongside the new `packages/utils` artifacts and registers both as production graphs. The data package has been shipping unmeasured through every size discussion in this series, so having it in the inventory is worth more than the bytes it reports.

Removing the Redux, Zustand, and XState Store entries is clean by comparison: these are deletions rather than moves, so there is nothing for a relocation record to express, and the artifacts, graphs, and package exports all come out together.

### Removing the id guard is a replacement, not a deletion

This is the part I'd flag as easy to misread. The native Model used to throw `'@marionette/data cannot change a Model id while it belongs to a Collection.'`, and that guard is gone. Read alone it looks like dropped safety — but:

> Native DataApi keys are model `cid` values, so application ids can change without changing child identity.

The guard existed because children were keyed by id, which made an id change silently rekey a child. Keying by `cid` makes id mutation harmless, so the constraint isn't relaxed, it's made unnecessary. That's the better shape: the invariant is enforced by the key choice instead of by a runtime throw plus the owner-tracking bookkeeping (`modelOwners`, `addModelOwner`, `releaseModelOwners`) that existed to know when to apply it. All of that goes away together, which is the tell that it was one mechanism rather than several features.

The guide's follow-on — "Keep application ids unique for unambiguous collection lookup" — is the right caveat, since `get(id)` still exists and is now the only place id collisions bite.

### Notes

`toJSON()` leaving the native package in favour of `toObject()` / `toArray()` is the other user-visible break, and the guide is explicit that callers serialize plain values themselves rather than the method changing shape quietly.

The `@marionette/utils` extraction doesn't add an external import to core — the `.` graph entry is untouched, so utils is bundled rather than imported, and the structural checks will confirm that against the built output.

"Collection notifications now follow ordinary synchronous events. They do not combine nested mutations or recover missed notifications after a listener throws" is consistent with the coalescing removal in #415's successor work, and the guide gives the actual remedy (schedule structural mutations after the notification returns) rather than just naming the limitation.

### Still open on master

`delete this._isDestroying` remains absent from `src/`, now with three set sites after `region.ts:561` joined `destroy.ts:19` and `view.ts:215`. Unrelated to this PR's scope — noting only because #435 merged with it and there's no longer an open PR carrying the thread.

paulfalgout:
**Approval carries over** — reviewed the update at fetched SHA `5af496b6` (previously `78da050d`). Housekeeping, and the one change that could have gone wrong is self-checking.

```ts
// @ts-expect-error The assignment helpers copy enumerable string keys, not symbols.
-ownStrings[symbol];
+void ownStrings[symbol];
```

Wrapping an expression that a `@ts-expect-error` depends on is the kind of edit that quietly voids an assertion. It doesn't here — the symbol index still errors inside the `void`, so the directive is still satisfied. Worth noting the failure mode is loud rather than silent: had `void` suppressed the error, TypeScript would report the now-unused `@ts-expect-error` directive and the fixture would fail. That's a property of `@ts-expect-error` over `@ts-ignore` doing real work.

`adapters-package-size.mjs` drops three `[]` entries from `expectedExternalImports`, which is a positional array, so this is only correct if the alignment holds. It does: five subpaths remain — backbone, `dom/jquery`, xstate, `dom/morphdom`, `dom/lit-html` — against `[], ['jquery'], [], ['morphdom'], ['lit-html']`. The xstate `[]` is right too; that adapter imports only the internal keyed-snapshot helper and duck-types the actor shape rather than importing `xstate`.

A positional list like this is worth keeping an eye on generally — nothing in the file ties an entry to the subpath it describes, so a future insertion in the wrong slot would shift every later expectation silently. Not a change request, just the reason this particular three-line deletion deserved checking rather than skimming.

CI: 3 passing, 7 running, nothing failing.

PR #442, opened 2026-09-07
refactor(application): use the Region current View

paulfalgout:
## What
Make Application's View its Region's `currentView`. Remove separate `_view` tracking, the Region empty listener, and synchronization helpers. Update lifecycle tests and the Application/migration documentation.

## Why
The extra View ownership policy duplicated Region state. For example, after `app.getRegion().show(replacement)`, `app.getView()` now returns that replacement and stopping the Application removes it.

## Scope
Application View reads and stop/restart/destroy cleanup. Region lifetime ownership remains unchanged: Application destroys Regions it constructs and preserves borrowed Region instances.

One authoritative current View removes the need to reason about independently tracked Application and Region contents.

## Public behavior contract
Views shown directly through the Region, including before Application construction, are Application's current View. Teardown removes the current View. When no View is present, Application stop leaves unmanaged HTML alone. The former external-replacement exemption is removed without a compatibility alias.

## Runtime cost
Removes stored View references and the additional Region listener. Cleanup checks `currentView` before calling `empty`, avoiding unrelated raw-HTML removal or DOM target resolution. No timing claim.

## Deployment Impact
Library source changes only. No forms or application bundles need redeployment as part of this PR. Consumers opting into a future release must account for the explicit View ownership change; no package is published here.

## Testing
- `npm run check:types`
- Application, composition, and State-owner Vitest suites.
- Tests cover externally shown/replaced/re-shown Views, prepopulated Regions, stop/restart/destroy, borrowed Region lifetime, unmanaged HTML, missing DOM targets, and existing lifecycle reentrancy.
- `npm run docs:check`: 55 pages, 506 links, 23 executable examples on the combined follow-up docs.
- Combined follow-up build and coverage passed; CI verifies the isolated PR.

## Package and browser impact
No new public methods, entry points, dependencies, or browser APIs.

## Rollback or deprecation
Revert this PR to restore the former tracking policy. No dual ownership mode is retained.

---
## Summary by cubic
The Application now treats its Region's `currentView` as the single source of truth for the root View, replacing the separate `_view` tracking, the Region empty listener, and synchronization helpers. Previously, Views shown directly through the Region (or externally replaced) were not claimed or emptied by the Application; now `app.getView()` returns whatever the Region shows, and stopping or destroying the Application empties that View. Teardown no longer touches unmanaged HTML when the Region has no View, and borrowed Regions are still preserved on destroy.

**Behavior changes**
- Views shown via `app.getRegion().show(view)`, including before Application construction, are now returned by `app.getView()` and destroyed on stop/restart/destroy.
- The former external-replacement exemption is removed without a compatibility alias.
- Docs (`marionette.application.md`, `migration-from-v4.md`) and lifecycle tests updated to cover externally shown, re-shown, prepopulated, and detached Views.

<sup>Written for commit 0ab835f9d65e399e0ed10549dab0030754dd1cfe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/442?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approved, with one question about dispatch** — reviewed at fetched SHA `a26af81d` (base `55fbd24b`).

Removing `_view` in favour of `region.currentView` eliminates a genuine duplication: `getRootView` existed only to detect that the two had drifted, and `clearRootView` plus the `empty` listener existed only to keep them from drifting. All three go away because the question they answered can no longer be asked. `getView()` can't return a stale View now, and each Application carries one fewer subscription.

`showView` dropping its `region.currentView === view` check is right for the same reason — the old code needed it to decide whether to trust `_view`, and with the region authoritative there's nothing to decide.

### Teardown now dispatches through the overridable `getRegion()`

`emptyView` reads `application.getRegion()`, where the previous `emptyRootView` and `getRootView` both read `application._region` directly. `getRegion()` is a public, overridable method, so teardown now follows a subclass override where it previously didn't.

That's defensible as consistency, but it splits from ownership. The description keeps ownership on construction — "Application destroys Regions it constructs and preserves borrowed Region instances" — which is `_ownedRegion`. So for an Application whose subclass overrides `getRegion()` to return a different Region:

- `stop()` empties the **overridden** Region
- `destroy()` still destroys the **owned** Region

Previously both followed `_region` and couldn't diverge. It's an edge, and arguably an override returning a foreign Region is asking for trouble — but View's `getRegion` overridability is explicitly contracted in the migration docs, including non-delegating overrides, so it isn't unthinkable that someone does the same on an Application.

Worth a deliberate answer: should teardown read `_region` because it's an ownership operation, or `getRegion()` because it's a public lookup? Either is fine; having emptying and destroying disagree is the part I'd avoid.

### Note on the removed exemption

"Views shown directly through the Region, including before Application construction, are Application's current View. Teardown removes the current View." That's a real behavior change — a View the Application never showed is now destroyed by `stop()` — and the description states it plainly rather than filing it as an internal cleanup. Given the alternative was Application and Region disagreeing about what the current View is, taking the simpler rule looks right.

`getRegion()` is `return this._region;` with no side effects, so `getView()` costs nothing extra despite the added call.

### Merge ordering

#441 also edits `src/modules/application.ts` — adding `preinitialize` to the constructor and the instance interface, while this removes `_view` and `_onRootRegionEmpty` from `ApplicationInternals`. Different regions of the file, so probably no textual conflict, but whichever lands second is worth a re-run rather than a fast-forward.

CI: 6 passing, 6 running, nothing failing.

paulfalgout:
Status note — #441 merged, so `master` is now `088c34ba` and this branch is a revision behind. I checked the overlap I flagged earlier rather than leaving it as a caution.

**No conflict.** `git merge-tree` reports three files changed in both and **zero conflict markers**: `src/modules/application.ts`, `docs/marionette.application.md`, and `test/unit/object-application-composition.spec.js` all auto-merge.

The interesting one is the spec, because both PRs edit the same array:

```js
const applicationFinalKeys = [
+     'preinitialize',        // #441, top of the list
      'cidPrefix',
      …
      'getRegion',
-     '_onRootRegionEmpty',   // #442, near the bottom
      'showView',
```

Different lines, so git merges them without complaint — and here that's also the semantically correct result: the merged list should gain `preinitialize` and lose `_onRootRegionEmpty`.

Worth noting that this particular file is self-guarding. `applicationFinalKeys` is an exact ordered list asserted against the real prototype's own keys, so a merge that got this wrong would fail the test rather than compile and pass. That's the reassurance I couldn't give when I raised the ordering concern — the risk exists, and this specific test is the thing that catches it.

#440 merges cleanly against the new master with no overlap at all.

Review unchanged at `a26af81d`; the `getRegion()`-versus-`_region` dispatch question from my earlier comment still stands on its own.

paulfalgout:
**Approval carries over** — re-checked at fetched SHA `0ab835f9` (previously `a26af81d`).

Pure rebase onto `28ada8b3` after #440 merged. The merge base is now master itself, and the effective diff is unchanged: the same 5 files, +79/−76. The apparent delta against the old head is #440's and #441's work arriving through the rebase, not new content here.

The `object-application-composition.spec.js` merge I checked earlier resolved as expected — `applicationFinalKeys` now carries `preinitialize` from #441 and drops `_onRootRegionEmpty` from this PR, which is the correct combined result.

Still open from my review: `emptyView` reads the overridable `getRegion()` while destruction follows `_ownedRegion`, so a `getRegion()` override makes stop and destroy act on different Regions. Worth a deliberate answer rather than a change request.

PR #444, opened 2026-09-07
fix(backbone): leave merged model rendering to child events

paulfalgout:
Related #376

## What
- Leave Backbone merged-model rendering to child model events, matching Marionette v4.
- Continue collection sorting and filtering after merges without invalidating retained children.
- Document the distinction between model events and explicit DataApi render requests.

## Why
A collection.set() merging one changed model and one unchanged model rendered children [1, 1, 2] when the children subscribed to change:render. It now renders only [1]. Backbone reports unchanged models as merged, so forwarding every merge as an in-place render request also replaced DOM unnecessarily.

## Scope
Backbone adapter observation and its tests/documentation. Core DataApi explicit invalidation and immutable replacement behavior remain intact.

## Deployment Impact
No forms require redeployment. This changes the optional Backbone adapter in the next package build; no package publication is part of this PR.

## Testing
- Added two regressions, observed both fail before the fix, then pass afterward.
- npx vitest run test/unit/backbone-adapter.spec.js test/unit/runtime/backbone-api.spec.js test/unit/collection-view test/unit/data-api-integration.spec.js --reporter=dot: 396 passed.
- npm run coverage: 1,958 unit tests and 19 agent contract tests passed, coverage requirements passed.
- npm run build: package builds and packed ESM/CommonJS TypeScript consumers passed.
- npm run docs:check: executable examples, 55 documentation pages and 506 internal links passed.
- Targeted ESLint and git diff --check passed.
- Render counts prove the removed work; no browser timing claim is made for this adapter-only change. The existing native/morphdom/lit-html/jQuery benchmark fixtures use @marionette/data, not the Backbone adapter.

---
## Summary by cubic
Fixes the Backbone adapter so merged models no longer trigger child re-renders, matching Marionette v4. A `collection.set()` merging one changed and one unchanged model previously rendered children [1, 1, 2]; it now renders only [1] since rendering is left to child `modelEvents`.

**Behavior change**
- Merges still update collection order and filtering without invalidating retained children.
- Unchanged models are no longer redrawn because Backbone reports them as merged too.
- Only the optional Backbone adapter changes; core DataApi explicit invalidation and immutable replacement behavior stay intact.
- Relates to #376.

<sup>Written for commit f0f5d3665649148dbcc9b93338c9821424ec4bcc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/444?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collection updates so only models with actual data changes re-render.
  * Preserved existing content for unchanged items, reducing unnecessary visual refreshes.
  * Collection sorting and filtering now update correctly without re-rendering retained items.
  * Removed items are correctly dropped from collection views during updates.

* **Documentation**
  * Clarified how model changes and collection updates control child rendering in the Data API and Backbone integration.

### DIFF
--- a/docs/data.api.md
+++ b/docs/data.api.md
@@ -93,6 +93,12 @@ every same-key replacement View before removing any existing child. A
 replacement-construction or rendering failure propagates to the caller. Core
 does not undo a partial update or promise recovery on the next notification.
 
+An in-place `updated` entry requests a child render. Adapters for mutable models
+with their own change events can leave `updated` empty and let child
+`modelEvents` handle rendering. The Backbone adapter follows this approach:
+merges still update collection order and filtering, without rendering children
+again after their model events have run.
+
 An immutable same-key replacement belongs only in `updated`, not in `removed`
 and `added`. Replacing a model with one that has a different stable key is a
 removal plus an addition; changing the key of a retained model is invalid. The
--- a/docs/optional-backbone.md
+++ b/docs/optional-backbone.md
@@ -47,6 +47,12 @@ Backbone's `sort`, `reset`, and `update` payloads are translated to the neutral
 records documented by [`DataApi.observeCollection()`](./data.api.md#collection-observations).
 Those Backbone-specific shapes do not enter Marionette core.
 
+As in Marionette v4, child `modelEvents` control rendering after model changes.
+For example, `modelEvents: { change: 'render' }` renders a child when its model
+changes. Collection merges still sort and filter children, but do not request
+another render. Backbone also reports unchanged models as merged, so treating
+every merge as a render request would redraw unchanged children.
+
 Sort handling follows Marionette v4: the adapter skips `sort` events carrying
 `add`, `remove`, or `merge` flags and handles those mutations through `update`.
 Explicit `collection.sort()` calls still notify the View. The observer does not
--- a/packages/adapters/src/data/backbone.ts
+++ b/packages/adapters/src/data/backbone.ts
@@ -64,7 +64,8 @@ const BackboneApi = {
         kind: 'update',
         added: changes.added,
         removed: changes.removed,
-        updated: changes.merged.map(model => ({ previous: model, current: model }))
+        // Merges retain the model; child model events own its rendering, as in v4.
+        updated: []
       });
     };
     const events = {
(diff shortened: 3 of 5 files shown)


paulfalgout:
**Approved** — reviewed at fetched SHA `f0f5d366` (base `ad547096`).

One line of source, and the reasoning behind it is the interesting part. `updated` entries request a render because an immutable replacement produces a *different* object that the existing child can't observe. A Backbone merge is the opposite: the model is mutated in place and keeps its identity, so the child's own `modelEvents` subscription already fires. Forwarding the merge as an `updated` entry asked core to render something that had just rendered itself.

Backbone reporting unchanged models as merged turns that from redundant into wrong — the `[1, 1, 2]` in the description is one duplicate render plus one entirely spurious one.

### The node-identity assertion is what makes the first test real

```js
expect(view.children.findByIndex(1).el.firstChild).to.equal(unchangedContents);
```

Checking `rendered` and `textContent` would both pass against a re-render that produced identical HTML — which is precisely the failure mode here, since the unchanged model's template output doesn't change. Holding the original child node and asserting identity is the only assertion in that test that distinguishes "didn't render" from "rendered the same thing", and it's also what proves the DOM wasn't replaced under a focused input.

The second test covers the harder half of the claim: a merge that changes sort rank and filter visibility still reorders and filters (`'31'`), while `findByModel` returns the *same* child instance and `onRender` is never called. Sorting and filtering responding to merges without invalidating children is easy to break in either direction, and that test pins both.

### The generalization is the right place to put it

`data.api.md` frames this as an adapter choice rather than a Backbone quirk — "adapters for mutable models with their own change events can leave `updated` empty and let child `modelEvents` handle rendering." That tells a future adapter author which side of the line their source falls on, which the Backbone-specific note alone wouldn't.

Worth being explicit about the accepted trade, which both documents state: a child with no `modelEvents` no longer re-renders when its model is merged. That's v4 behavior and the right default given the alternative, but it is a real behavior change for anyone who had been getting the render for free from the adapter.

PR #446, opened 2026-09-07
fix(collection-view): sort each notification from its snapshot

paulfalgout:
Related #376

## What
- Use the current collection notification's captured snapshot for default source ordering, including nested update/reorder/reset notifications.
- Perform source ordering inside the normal sort path, after before:sort, and remove the redundant default source-child partition/rebuild.
- Preserve manual-child placement, stable custom-comparator ties, disabled source ordering and explicit sort's live-source behavior.

## Why
A child-add hook can enqueue an immutable replacement before the current update renders. The old default comparator reread the live source, making the earlier update render children in a future snapshot's order. It also reordered children before before:sort fired and read Data.models twice per ordinary update. The queue now retains its processing notification until rendering ends; sorting reads that snapshot once.

## Scope
CollectionView sorting and notification dispatch, regression tests and documentation. The existing sort → filter → render flow remains canonical. Overrides that skip the parent sort now own all sorting; no fallback is injected. Custom comparators retain control of their own data reads, and the existing built-in comparator identity check continues to preserve overrides. No callback-destruction guards, new dispatch flags or rollback behavior are added.

## Deployment Impact
No forms require redeployment. This affects the next core package build; there is no package publication or application deployment in this PR.

## Testing
- Three regressions failed before the source fix: queued replacement render order, before:sort timing, and duplicate source reads. They pass after the fix.
- Additional coverage checks nested reorder/reset notifications, default/manual order, custom comparator ties and sortWithCollection:false.
- npm run coverage: 1,966 unit tests, 100% required coverage and 19 agent contract tests passed.
- npm run build: package/declaration builds and packed ESM/CommonJS consumer checks passed.
- npm run docs:check, targeted ESLint and git diff --check passed.
- node test/browser/collection-removal-survivors.mjs: all 15 configurations passed across Chromium, Firefox and WebKit.
- External deterministic audit: 8,000 mutation steps across 16 combinations of native/jQuery/morphdom/lit-html and sorting modes passed, checking DOM order, retained identity, rendered data, destruction and observer cleanup.
- Bounded Claude review completed with no blockers after clarifying the custom-comparator data-read boundary. No override interception was added.

### Matched performance check
Standard native js-framework-benchmark fixture; same row template/build tooling, headless Chrome 152, renderer accessibility explicitly disabled, lifecycle monitoring enabled. Base/candidate/candidate/base blocks, seven samples per operation per block; all 140 timed samples retained. No concurrent builds/tests during timed blocks. These AX-off numbers do not represent accessibility-enabled user performance.

| Operation | Base total median ms | Candidate total median ms | Base script ms | Candidate script ms |
| --- | ---: | ---: | ---: | ---: |
| Create 1k | 44.55 | 44.60 | 17.25 | 17.15 |
| Partial update | 41.25 | 40.40 | 9.45 | 9.55 |
| Remove | 17.65 | 17.90 | 1.60 | 1.70 |
| Append 1k | 48.55 | 48.75 | 16.20 | 16.30 |
| Clear | 30.30 | 29.80 | 27.55 | 27.10 |

This run remains close to baseline; it does not demonstrate a general speedup. The concrete work reduction is one source read instead of two plus removing the default source-child rebuild. Frozen inputs, hashes, raw samples/traces and reproduction scripts are retained in the 2026-09-08 marionette-collectionview-audit local artifacts.

---
## Summary by cubic
Fixes CollectionView sorting to order children from each notification's captured snapshot instead of the live source, so nested updates and reorders render correctly and `before:sort` fires before source order changes.

**Bug Fixes**
- Update and reorder notifications now sort from the notification snapshot rather than reading the live collection twice.
- `before:sort` now fires before any source-order changes are applied.

**Refactors**
- Kept the notification queue's head snapshot active until rendering ends, so nested notifications don't alter the current sort pass.
- Removed the redundant default source-child rebuild from update handling.
- Documented that `sort()` outside a notification uses the current source, and that custom comparators retain control of their own ordering and ties.

<sup>Written for commit b8b876010cc3c2f3a3e530f1ac988a0a903184d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/446?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -958,6 +958,17 @@ By default the `CollectionView` will maintain a sorted collection's order
 in the DOM. This behavior can be disabled by specifying `{sortWithCollection: false}`
 on initialize.
 
+Default source ordering uses each notification's captured snapshot. A nested
+notification waits for the current sort, filter, and render pass to finish.
+Calling `sort()` outside a collection notification reads the current source
+after `before:sort`. With the default comparator, manually added children whose
+models are absent from the source stay before the source children.
+
+Custom comparators still determine their own order and data reads. With
+`sortWithCollection` enabled, source order breaks ties and manually added
+children follow source children on ties. With it disabled, ties retain the
+existing child order.
+
 ### Defining the `viewComparator`
 
 `CollectionView` allows for a custom `viewComparator` option if you want your
--- a/src/modules/collection-view.ts
+++ b/src/modules/collection-view.ts
@@ -190,9 +190,9 @@ type CollectionViewInternals = CollectionViewInstance & ViewMixinHost & {
   _initChildViewStorage(): void;
   _initialEvents(): void;
   _onCollectionChange(change: unknown): void;
-  _onCollectionReorder(snapshot: Snapshot): void;
+  _onCollectionReorder(): void;
   _onCollectionReset(snapshot: Snapshot): void;
-  _onCollectionUpdate(changes: Update, snapshot: Snapshot): void;
+  _onCollectionUpdate(changes: Update): void;
   _setChildrenFromSnapshot(snapshot: Snapshot): void;
   _removeChild(view: CollectionChild): void;
   _addChildModels(models: unknown[]): CollectionChild[];
@@ -444,21 +444,20 @@ Object.assign(CollectionView.prototype, ViewMixin, {
       return;
     }
 
-    const queue: Notification[] = this._collectionChangeQueue = [];
-    let pending: Notification | undefined = notification;
+    const queue: Notification[] = this._collectionChangeQueue = [notification];
 
     try {
-      while (pending) {
-        const { change: pendingChange, snapshot } = pending;
+      while (queue.length) {
+        const { change: pendingChange, snapshot } = queue[0];
         if (pendingChange.kind === 'reorder') {
-          this._onCollectionReorder(snapshot);
+          this._onCollectionReorder();
         } else if (pendingChange.kind === 'reset') {
           this._onCollectionReset(snapshot);
         } else {
-          this._onCollectionUpdate(pendingChange, snapshot);
+          this._onCollectionUpdate(pendingChange);
         }
         this._collectionSnapshot = snapshot;
-        pending = queue.shift();
+        queue.shift();
       }
     } finally {
       delete this._collectionChangeQueue;
@@ -466,16 +465,14 @@ Object.assign(CollectionView.prototype, ViewMixin, {
     }
   },
 
-  // Internal method. This checks for any changes in the order of the collection.
-  // If the index of any view doesn't match, it will re-sort.
-  _onCollectionReorder(this: CollectionViewInternals, snapshot: Snapshot) {
+  // Follow source reorders unless the CollectionView manages its own order.
+  _onCollectionReorder(this: CollectionViewInternals) {
     if (this._isDestroying || this._isDestroyed) { return; }
 
     if (!this.sortWithCollection) {
       return;
     }
 
-    this._setChildrenFromSnapshot(snapshot);
     this.sort();
   },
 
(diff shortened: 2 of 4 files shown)


paulfalgout:
**Approved** — reviewed at fetched SHA `bd09c3b6` (base `c26763ad`).

The enabling change is small and easy to skim past:

```js
-const queue = this._collectionChangeQueue = [];
-let pending = notification;
+const queue = this._collectionChangeQueue = [notification];
```

Previously the queue held only *future* notifications, so nothing inside `sort()` could tell which pass it was running in — hence the live `Data.models()` read, and hence a pass rendering in a snapshot that hadn't happened when it started. Seeding the queue with the in-flight notification and shifting only after processing completes makes `_collectionChangeQueue?.[0]` a reliable "which notification am I serving" answer, and everything else follows from that.

This is the same family as the stale-closure problems in #415 and #423, inverted: those used a captured value after it went stale, this used live state where a captured one was authoritative. Both reduce to *which snapshot does this operation belong to*, which is now answerable rather than inferred.

### The three exclusions each earn their place

```js
if (notification && this.sortWithCollection && viewComparator !== defaultViewComparator &&
    notification.change.kind !== 'reset') {
```

- No notification → explicit `sort()`, which should read live source.
- Default comparator → it already orders by source index through `indexByModel`, so applying the snapshot first is the redundant rebuild being removed.
- Reset → `_onCollectionReset` already rebuilt children in source order.

Worth noting `_setChildrenFromSnapshot` still runs when `viewComparator` is falsy, before the `if (!viewComparator) return`. That preserves the old net behavior — source order applied with no `before:sort`/`sort` events — even though the code that does it moved.

### `before:sort` now fires earlier, which is a real change

Source ordering moved from `_onCollectionUpdate` (before `sort()` was called) into `sort()` after the event, so a `before:sort` handler now observes children in their *pre*-source-ordering arrangement. `fires before:sort before changing children to the new source order` pins it and the docs describe it, which is the right treatment for a change that's invisible until someone's handler reads `children` at that moment.

### The tests map one-to-one onto the claims

Each of the seven covers a distinct branch — queued ordering, event timing, a reorder queued *from* `before:sort`, the reset exclusion, manual-child ties parameterized by comparator kind, ties with source ordering disabled, and `uses one source snapshot to validate and sort a collection change`, which pins the "twice per update" reduction structurally rather than as a performance assertion.

### One unstated complement

> Calling `sort()` outside a collection notification reads the current source after `before:sort`.

That leaves the inside case implicit: an explicit `sort()` from a child-add hook *during* a notification will use that notification's snapshot, not live source. I think that's the correct behavior — inside a pass, the snapshot is authoritative — but the sentence states only half a rule, and the half it omits is the surprising one.

paulfalgout:
Status note — #445 merged, so `master` is now `bc89f6c5` and this branch is a revision behind. Review unchanged at `bd09c3b6`; no new commits.

**Merges cleanly.** `git merge-tree` reports `src/modules/collection-view.ts` and `docs/marionette.collectionview.md` changed in both, with **zero conflict markers** — the two PRs touch different regions of each file.

Semantically they don't interact either: #445 works in `_setupChildView` / `_removeChildView` and the MN0003 guard, while this changes the notification queue, `_onCollectionReorder`, `_onCollectionUpdate`, and `sort()`. The one place worth checking is that `_setChildrenFromSnapshot` — which this PR now calls from inside `sort()` — reorders existing children rather than removing and re-adding them, so it never crosses the ownership set/clear pairs #445 introduced.

No rebase needed on my account, though one wouldn't hurt if you want CI measured against what actually lands.

paulfalgout:
**Approval carries over** — re-checked at fetched SHA `b8b87601` (previously `bd09c3b6`).

Pure rebase onto `bc89f6c5`. Merge base is now master itself and the effective diff is unchanged: the same 4 files, +160/−23, with the queue seeding, the three `_setChildrenFromSnapshot` exclusions, and the snapshot-based default comparator all intact.

The delta against the old head is #445's ownership work arriving, which also confirms the coexistence I checked before the rebase: `_setupChildView` now carries `view._parent = this` immediately above the listener setup, while this PR's changes to the notification queue, `_onCollectionReorder`, `_onCollectionUpdate`, and `sort()` sit elsewhere in the file. No interaction, and git needed no help resolving it.

Nothing further from me; the one open item is still the half-stated rule about `sort()` inside a notification pass.

PR #449, opened 2026-09-07
fix(collection-view): remove disabled empty views

paulfalgout:
Related #376

## What
Remove an already displayed empty View when `emptyView` subsequently resolves to `false`, `null`, or `undefined`. Reuse `_destroyEmptyView()` and document the transition.

## Why
A dynamic empty-state resolver could stop requesting an empty View, but `filter()` or `render()` left its old message visible and alive. The early return also exists in v4.1.3. The documented disabled values now remove the old View; re-enabling the resolver creates a fresh one.

## Scope
CollectionView empty-state rendering: one source line plus regression tests and documentation. Existing Region cleanup preserves unrelated content when no empty View is present. No new API or error-recovery behavior.

## Deployment Impact
No forms or applications require deployment for this repository change. Consumers receive it in a later adopted library build; no publication occurs here.

## Testing
- Three disabled-result regressions failed before the fix.
- `npm run coverage`: 1,981 unit tests, 19 agent contracts, and required coverage passed. The final targeted 49-test suite additionally covers the three transitions through `render()` as well as `filter()`.
- `npm run build`, targeted ESLint, `npm run docs:check`, and `git diff --check`: passed.
- Chromium, Firefox, and WebKit: disabled/re-enabled empty Views with monitoring on/off preserve their parent input, value, focus, and selection (six cases per browser).
- Bounded Claude review found no correctness blocker; added its suggested render-path coverage. Initial disabled values and unrelated-markup preservation are also covered by existing tests.
- Matched clear-1k benchmark: exact master `ab1f4852`, headless Chrome 152, monitoring on, accessibility explicitly disabled, same harness/template/compiler, base/candidate/candidate/base with nine samples per block (36 total). Median total **29.85→30.10 ms**, script **27.15→27.40 ms**; block results overlap. Raw samples, traces, and bundle identities retained. This is not an accessibility-enabled browser measurement or a universal performance guarantee.

---
## Summary by cubic
Fixes `CollectionView` empty-state handling. Previously, when the `emptyView` resolver returned `false`, `null`, or `undefined` during `filter()` or `render()`, the old empty View stayed visible and alive; now it's destroyed and removed, and re-enabling the resolver creates a fresh one.

- Addresses #376.
- Adds regression tests covering all three disabled values through both `filter()` and `render()`.
- Documents the new behavior in the collection view docs.

No deployment impact; consumers pick this up in a later adopted library build.

<sup>Written for commit ff5c6bd13d0ce09b261f4a96fa40446dcde29ed3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/449?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -531,6 +531,9 @@ definition. A resolver may return a `View` class or `undefined`, `null`, or
 Marionette trusts the result when the collection is empty. Errors thrown by a
 resolver propagate unchanged.
 
+When the empty collection is rendered or filtered again, a disabled result also
+removes any empty View already shown.
+
 ```javascript
 import _ from 'underscore';
 import { View, CollectionView } from 'marionette';
--- a/src/modules/collection-view.ts
+++ b/src/modules/collection-view.ts
@@ -1013,6 +1013,7 @@ Object.assign(CollectionView.prototype, ViewMixin, {
     const EmptyView = this._getEmptyView();
 
     if (!EmptyView) {
+      this._destroyEmptyView();
       return;
     }
 
--- a/src/modules/collection-view.ts
+++ b/src/modules/collection-view.ts
@@ -167,7 +167,7 @@ export type CollectionViewConstructor<Props extends object = {}, Args extends un
 interface SnapshotEntry {model: unknown; key: unknown;}
 interface Snapshot {
   entries: SnapshotEntry[];
-  models: Map<unknown, SnapshotEntry>;
+  models: ReadonlyMap<unknown, SnapshotEntry>;
 }
 interface Replacement {key: unknown; previous: unknown; current: unknown;}
 interface Update {kind: 'update'; added: SnapshotEntry[]; removed: SnapshotEntry[]; updated: Replacement[];}
@@ -242,9 +242,8 @@ function throwCollectionProtocolError(message: string): never {
   });
 }
 
-function buildCollectionSnapshot(Data: DataProvider, collection: unknown, previous: SnapshotEntry[]): Snapshot {
+function buildCollectionSnapshot(Data: DataProvider, collection: unknown, previous?: ReadonlyMap<unknown, SnapshotEntry>): Snapshot {
   const models = Data.models(collection as never);
-  const previousKeys = new Map(previous.map(entry => [entry.model, entry.key]));
   const keys = new Set<unknown>();
   const modelEntries = new Map<unknown, SnapshotEntry>();
   const snapshot: SnapshotEntry[] = Array(models.length);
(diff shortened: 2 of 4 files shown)


paulfalgout:
**Approved** — reviewed at fetched SHA `c7a6686c` (base `ab1f4852`).

One line, and the asymmetry it fixes is visible right in `_renderChildren`: the non-empty branch calls `_destroyEmptyView()` unconditionally, while the empty branch delegated to `_showEmptyView()`, which had a path — resolver returns a disabled value — that neither showed nor destroyed anything. A stale empty View survived every subsequent `render()` and `filter()`, still alive and still in the DOM, because nothing in that path was responsible for it.

`_destroyEmptyView()` was already safe to call here:

```js
const emptyRegion = this.getEmptyRegion();
// Only empty if a view is show so the region
// doesn't detach any other unrelated HTML
if (emptyRegion.hasView()) { emptyRegion.empty(); }
```

That guard is what makes the fix a one-liner rather than a conditional — and it matters more than usual right after #447, since an unguarded `empty()` on a container holding template content is precisely the failure that PR was addressing. Reusing the existing method inherits that protection instead of re-deriving it.

The tests check both halves of the transition:

```js
expect(previous.isDestroyed()).to.be.true;
expect(emptyRegion.hasView()).to.be.false;
expect(myCollectionView.el.childNodes.length).to.equal(0);
```

Asserting destruction *and* region state *and* an empty DOM covers the three ways this could half-work — destroyed but still mounted, unmounted but still alive, or the region's bookkeeping disagreeing with reality. And `expect(emptyRegion.currentView).to.be.instanceOf(OtherView).and.not.equal(previous)` pins that re-enabling builds a fresh View rather than resurrecting the destroyed one, which is the behavior the docs now promise.

Parameterizing over the disabled values and over `render` / `filter` covers both entry points named in the description.

Worth noting this predates v5 as well — the early return is in v4.1.3 — so it's another longstanding bug rather than a regression from the recent rendering work.

paulfalgout:
Status note — #448 merged, so `master` is now `9d464656` and this branch is a revision behind. Review unchanged at `c7a6686c`.

**Merges cleanly.** Only `src/modules/collection-view.ts` is changed in both, with zero conflict markers. The two land in different parts of the file — #448 reworked the focus block inside `_renderChildren` and `buildCollectionSnapshot`'s key validation, while this adds a single line to `_showEmptyView`.

No semantic interaction either: the `_showEmptyView` path returns before reaching any of the focus or snapshot code #448 touched, and `_destroyEmptyView` operates on the empty Region rather than the child container.

No rebase needed on my account.

paulfalgout:
**Approval carries over** — re-checked at fetched SHA `ff5c6bd1` (previously `c7a6686c`).

Pure rebase onto `9d464656`. Merge base is now master itself and the effective diff is unchanged: the same 3 files, +30, with the `_destroyEmptyView()` call in `_showEmptyView` intact.

The delta against the old head is #448's focus-scan and snapshot work arriving through the rebase, which confirms the non-interaction I described before it landed — the two changes sit in different parts of `collection-view.ts` and git needed no help combining them.

Nothing further from me.

PR #465, opened 2026-09-08
fix(data): preserve model identity and bound bulk removal work

paulfalgout:
## What

- Preserve supplied native Model instances across Collection construction, add, and reset; correct collection factory inference, array-removal return types, partial attributes, and known-key writes.
- Make identity lookup prefer exact member, application id, then cid; resolve batch removals with a temporary Map and Set instead of repeated scans.
- Document mutation semantics and add correctness, type, and complexity regressions plus reproducible before/after measurements in `evidence/data-package-audit.md`.

## Why

A configured Collection could copy a different Model subclass's internals into its attributes, losing normal field reads. Types could promise subclass methods on plain Models and reject valid array-removal results. ID/cid collisions changed lookup results after a reorder. Removing 10,000 exact instances also performed 49,995,000 ID reads.

The local macOS arm64/Node 24 comparison measured 10,000-model removal at 531.228 ms before and 6.266 ms after (84.78x); removal by IDs improved 72.89x. These are medians of nine alternating samples after two warmups, excluding construction and cleanup. Single-removal differences were small and mixed. ESM grew 124 Brotli bytes; CommonJS grew 113 bytes. No size budget was relaxed.

## Scope

`@marionette/data` runtime, declarations, documentation, tests, and a removal benchmark. Supplied native Models are now preserved canonically; there is no conversion fallback. Core, adapters, persistence, automatic sorting, and observer scheduling are unchanged. `Collection.set()` is intentionally excluded because it introduces a separate reconciliation contract.

Native-data type corrections are intentionally breaking: `attributes` and `toObject()` now return partial attributes, and collections using the base factory no longer promise subclass-only methods. Callers must handle absent attributes and configure or narrow model factories appropriately.

## Deployment Impact

No forms or form bundles need redeploy. No shared form code is changed. This affects consumers when they adopt a subsequently published `@marionette/data` package; this PR does not publish a package. The normal master documentation workflow publishes the documentation changes after merge.

## Testing

- `npm run build` — passed, including core declaration/consumer checks.
- `node node_modules/vitest/vitest.mjs run --coverage --reporter=dot` — 2,004 tests passed; 100% lines and branches.
- Native-data type fixture installed from packed data/utils tarballs: `npm run validate` — ESM/CommonJS TypeScript 7.0.2 and legacy TypeScript 4.6.4 passed.
- `npm exec -- node test/browser/data-package.mjs` — packed runtime passed Chromium, Firefox, and WebKit.
- `npm run test:performance` — 64 passed, including bounded identity-read work.
- `npm run test:source`, `npm run test:dist`, `npm run lint:ci`, `npm run docs:check`, `npm run size:data`, and `node scripts/performance/bundle-size.mjs --json` — passed.
- `node --expose-gc scripts/performance/data-package-removal.mjs BASELINE_SOURCE_DIRECTORY` — baseline `5bb66c7f`; samples and source hashes retained in the evidence JSON.

Regressions cover mixed native Model subclasses, removal return inference, absent attributes, invalid known-key writes, mixed instance/id/cid removal, silent ID changes, identity collisions across reorders, and first-ID precedence after duplicate ID writes. The full cross-platform fixture matrix is left to CI; local fixture validation focused on the changed native-data declarations.


paulfalgout:
**Approved** — reviewed at fetched SHA `307cf2c4` (base `5bb66c7f`).

### The complexity test is the right kind

```js
Object.defineProperty(model, 'id', { get() { reads.id++; return id; } });
```

Counting property reads and asserting linear work, rather than asserting a wall-clock threshold, is what makes this claim testable in CI at all. A timing assertion would be flaky on shared runners and would drift with unrelated changes; a read count pins the actual property the optimization delivers and fails for the right reason if someone reintroduces a scan.

The `49,995,000` figure in the description is exactly `n(n-1)/2` for n = 10,000, so the before-state was computed rather than estimated — which matches what the old code did: `this.get(candidate)` scanning the array per candidate, plus `removed.includes(model)` inside the loop.

### The priority consistency is load-bearing and tested

`get()` now resolves by predicate rather than array position — exact member, then id, then cid — instead of one `find` whose three-way predicate let whichever model came first win. That's the reorder-sensitivity being fixed.

`indexModels` has to encode the same precedence in a Map, and does it through overwrite order: cids, then ids **iterating backward**, then exact instances. The reverse iteration is the subtle part — it makes the *first* model with a duplicate id win, matching `find`'s first-match semantics in `get()`. Had it iterated forward, single and bulk removal would silently disagree whenever an application temporarily assigned two models the same id.

That's exactly the invariant I'd have flagged as easy to break later, and `keeps first-id lookup consistent for single and bulk removal after duplicate id writes` tests it directly. Good.

### `_prepareModel` matches Backbone's contract now

`model instanceof ModelClass` → `model instanceof Model` fixes a genuinely nasty failure: a Model of a different subclass failed the check and was passed to `new ModelClass(otherModel)`, where its `attributes`, `cid`, `changed`, and `id` became *attribute values*. The symptom — ordinary field reads returning nothing — points nowhere near the cause.

Worth noting the consequence the description states plainly: a Collection configured with `model: SpecialModel` now retains a supplied plain `Model` rather than converting it, so `this.model` stops being a guarantee about membership. The types were updated to match (`CollectionInstance<M | Factory>`), which is the part that usually gets missed when a runtime contract widens.

### Notes

The `remove` overload reorder is the fix for array results: with the `unknown` single-identity overload first, an array argument matched it and resolved to `M | undefined`. More specific first is correct.

Guarding index construction on `candidates.length > 1` keeps the single-removal path from paying an O(n) Map build to answer one question.

Reporting +124 / +113 Brotli bytes is worth acknowledging given there is no longer a ceiling gate requiring it — self-reporting a size delta when nothing forces you to is the habit that keeps the numbers meaningful.

paulfalgout:
**Approval carries over** — reviewed the update at fetched SHA `47b1c218` (previously `307cf2c4`). Type fixtures only, across all three variants, and they pin the distinction this PR's runtime change makes necessary.

```ts
const explicitInput = new Collection<InputModel>([new InputModel()]);
// @ts-expect-error A base factory permits later base Models, even with explicit M.
explicitInput.at(0)!.inputOnly();
// @ts-expect-error Raw additions use the base factory.
explicitInput.add({ id: 2 })!.inputOnly();
const inputFactory = new Collection([new InputModel()], { model: InputModel });
const inputFactoryResult: string = inputFactory.add({ id: 3 })!.inputOnly();
```

**An explicit member generic is not a factory configuration**, and now that supplied Models are preserved rather than converted, that separation carries weight: a Collection's membership is a union of what was supplied and what the factory produces. `<InputModel>` constrains the first; `{ model: InputModel }` controls the second. Promising `inputOnly()` on `at(0)` or on a raw `add()` result would be the types guaranteeing something the runtime won't deliver.

Asserting both halves in one place is what makes it readable — the rejection and the acceptance sit four lines apart, so the difference between the generic and the option is visible rather than inferred from two distant tests.

This is the type-level counterpart of the runtime consequence I noted earlier — `this.model` no longer being a guarantee about membership — so the surface that widened and the declarations that describe it now say the same thing.

Runtime, docs, and evidence are unchanged from my review.

CI: nothing failing.

paulfalgout:
**Approval carries over** — re-checked at fetched SHA `40cdc243` (previously `47b1c218`).

Pure rebase onto `ae842653` after #455 merged. The merge base is now master itself and the effective diff is unchanged: the same 13 files, +698/−38, with the identity precedence, `indexModels` construction order, and the read-counting complexity test all intact.

The Region delta against the old head is #455 arriving through the rebase.

paulfalgout:
**Approval carries over** — reviewed the update at fetched SHA `af93fae8` (previously `40cdc243`).

`globalThis.gc?.()` is the failure mode where the output looks completely normal. Without `--expose-gc` the optional call silently does nothing, and the benchmark still prints timings — numbers whose stated method was not the method applied. Since those timings become `evidence/data-package-removal-2026-09-08.json` and the 84.78x claim, a silent no-op is the difference between evidence and a plausible-looking number.

Throwing at startup fixes it, and dropping the `?.` at the call site matters too: with the guard in place the optional call is redundant, but leaving it would let someone remove the guard later and quietly resume measuring without GC control.

The test pins the part that's easy to get wrong:

```js
assert.notEqual(result.status, 0);
assert.match(result.stderr, /Run with --expose-gc/);
assert.equal(result.stdout, '');
```

Failing isn't sufficient — it has to fail *before emitting anything*, or a caller capturing stdout could still scrape partial measurements from a run that then died. Asserting stdout is empty is what makes the guard's placement part of the contract rather than an implementation detail, and placing the check above `registerHooks` is what satisfies it.

Same discipline as the retired-workload rows and the "not comparable" consumer table: refuse to produce a number rather than produce one that can't be trusted.

CI: nothing failing.

PR #492, opened 2026-09-09
docs(readme): use the current Marionette logo mark

paulfalgout:
## What
- Replace the README's old PNG logo with the current crimson SVG mark used by the Marionette website.
- Set the displayed mark to 140 × 145 pixels and remove the obsolete PNG.

## Why
The repository README still displayed the previous logo.

## Scope
README header and repository image assets only. No runtime code, package entry points, or browser behavior changes.

## Deployment Impact
No forms or bundles need redeployment. The GitHub README image updates after merge; no separate deployment is required.

## Testing
- `git diff --check marionettejs/master...HEAD` — passed.
- `/usr/bin/xmllint --noout marionette-mark.svg` — passed.
- `cmp` against the website's `site/assets/mark.svg` — exact match.
- Python assertions confirmed the README reference, display dimensions, and removal of the old PNG.
- Runtime tests were not run locally because this change only affects documentation and an image asset. CI will run on the PR.

---
## Summary by cubic
Updates the README to use the current crimson Marionette SVG mark instead of the old PNG logo.

- Displays the mark at 140 × 145 pixels and removes `marionette-logo.png`.
- The SVG is an exact copy of the website’s `site/assets/mark.svg`.
- Docs-only change; no runtime, build, or deployment impact.

<sup>Written for commit 069ac7c5c58f90cad678ff3d4fa0314904a7a125. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/492?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Updated the Marionette logo displayed in the README to use the SVG mark with explicit dimensions.

PR #493, opened 2026-09-09
chore(deps): update Backbone and defer unsupported toolchain majors

paulfalgout:
## What
- Update the locked root Backbone development dependency from 1.4.0 to 1.6.1 and synchronize config/performance.json with the installed version. This incorporates #489 and repairs its failing performance-contract check.
- Exclude root Babel and TypeScript major updates from Dependabot until their verified tooling blockers are removed. Minor/patch updates and all consumer-fixture updates remain enabled.

## Why
#489 updates the lockfile alone, causing the performance tests and artifact checks to reject the stale Backbone pin.

#487, #488, and #490 fail npm ci with Babel peer conflicts. Even a coordinated Babel 8 update conflicts with the current/latest @rollup/plugin-babel 7.1.0 peer requirement of @babel/core ^7.0.0. Remove the exclusion when the Rollup integration supports Babel 8 and the full build is validated.

#491 fails npm ci because typescript-eslint 8.69.0 requires TypeScript >=4.8.4 <6.1.0. Declaration generation and public-test checks also use the TypeScript compiler API. Remove the exclusion after lint and compiler-API tooling support TypeScript 7. Separate TypeScript 7 consumer fixtures remain active.

The four incompatible PRs were closed with these reasons. This PR supersedes #489 after merge.

## Scope
Root development dependency lockfile, performance provenance, and Dependabot policy. No production source or public package contract changes; the existing Backbone compatibility range remains unchanged.

## Deployment Impact
No forms or form bundles require redeployment. No deployment beyond the repository's normal configured merge workflows is requested. No package publication or release tag is created.

## Testing
- npm ci --no-audit --no-fund passed, including prepare/build, declarations, consumer types, and distribution checks.
- npm run coverage passed: 125 files and 1,905 tests; existing coverage thresholds unchanged.
- npm run test:performance passed: 70 checks, including pinned-toolchain validation.
- Parsed .github/dependabot.yml using the repository's yaml dependency and asserted that the exclusions affect only root major updates, preserving fixture updates.
- git diff --check passed.
- Full PR CI and CodeQL will validate the exact branch before merge.

---
## Summary by cubic
Updates Backbone dev dependency from 1.4.0 to 1.6.1 and stops Dependabot from proposing Babel and TypeScript major upgrades that currently fail `npm ci`.

- Syncs `config/performance.json` to the new Backbone pin, fixing the failing performance-contract check.
- Supersedes the earlier lockfile-only Backbone update.
- Excludes `@babel/*` and `typescript` major updates in the root package only; consumer-fixture updates remain enabled.
- Removes exclusions once Rollup supports Babel 8 and lint/compiler tooling supports TypeScript 7.
- No production code or package contract changes; no redeployment needed.

<sup>Written for commit afd0d558d5deb6b14f18a152bd1cdf11c83fa629. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/493?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


paulfalgout:
**Approved** — reviewed at fetched SHA `000ee881` (base `b06750c5`). This resolves the whole blocked batch, and the Babel reasoning goes a step further than mine did.

### The backbone pairing is the missing half #489 needed

`config/performance.json`'s `lockedDependencies.backbone` moves `1.4.0 → 1.6.1` alongside the lockfile, which is exactly what `pins the checked-in release profile and dependency versions` was failing on. CI is green here where #489 has three failures from the same mismatch.

### Deferring by update-type is better than what I suggested

I proposed excluding the pinned set from automation. Ignoring only `version-update:semver-major` for the specific blocked packages is the sharper version: patch and minor bumps keep flowing, so the deferral doesn't silently freeze those dependencies while nobody's looking.

The comments are what keep it from becoming permanent by forgetting:

```yaml
# Remove when @rollup/plugin-babel supports Babel 8 and the build is validated.
# Remove after typescript-eslint and compiler-API tooling support TypeScript 7.
```

An ignore rule without a stated unblock condition is indistinguishable from an abandoned one a year later.

### Correction to my triage

I told #487/#488/#490 that combining the three into one branch was what made Babel 8 evaluable. That was necessary but **not sufficient**: `@rollup/plugin-babel@7.1.0` is a direct dependency here, and it doesn't support Babel 8 either. So even a combined branch would have hit a second wall — the ERESOLVE I diagnosed was the first blocker, not the only one. This PR's reason line names the real one.

### On #492

Also fine — `marionette-logo.png` is deleted and replaced by `marionette-mark.svg`, and no file in the tree still references the old name, so nothing is left pointing at a missing asset.

PR #494, opened 2026-09-09
chore(ci): move Windows validation after merge

paulfalgout:
## What
- Run the full Windows package-fixture job on pushes to `master`, while reporting the existing Windows PR check as skipped.
- Exclude Windows from release dry runs triggered by PRs; manually dispatched release validation still tests exact candidate tarballs on macOS and Windows before publication.
- Document the post-merge Windows policy and profile-validation timing.

## Why
Windows package fixtures took over 20 minutes in a recent successful run. Repeating them in both normal CI and release dry runs on PRs adds runner cost and blocks routine merges. Windows compatibility remains covered after merge and before release.

## Scope
GitHub Actions scheduling and release-profile documentation. Linux and macOS PR validation remain in place. No file-specific bypasses, runtime changes, or changes to package support. The existing Windows required-check name is preserved, so branch protection does not need a settings change.

## Deployment Impact
No forms or bundles need redeployment. No package publication or site deployment is performed by this change. Windows-specific failures can now be detected after merge; release publication still requires Windows artifact validation.

## Testing
- `npm run check:workflows` — workflow YAML and pinned actionlint passed.
- `npm run check:release-profile` — passed.
- `npm run check:release-promotion` — dry-run preflight passed.
- `npm run docs:check` — passed, including 106 HTML files and 1,577 internal links.
- `npm run test:release-promotion` — all 110 tests passed.
- Live PR CI confirms `Package smoke (Windows x64)` is skipped under its existing check name, and the release dry-run fixture matrix schedules only macOS. Post-merge Windows execution and manually dispatched Windows release validation have not been run as part of this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,18 +107,8 @@ jobs:
           if-no-files-found: warn
 
   package-smoke:
-    name: Package smoke (${{ matrix.name }})
-    runs-on: ${{ matrix.runner }}
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - name: macOS arm64
-            host: macos-arm64
-            runner: macos-15
-          - name: Windows x64
-            host: windows-x64
-            runner: windows-2025
+    name: Package smoke (macOS arm64)
+    runs-on: macos-15
 
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


paulfalgout:
**Approved, with the tradeoff stated** — reviewed at fetched SHA `772db485` (base `06751e20`).

The release gate is preserved, which is the part that matters most: `release.yml` still runs the Windows matrix entry on non-`pull_request` events, so nothing gets promoted without Windows validation. What changes is *when* a Windows regression is discovered — after merge rather than before.

That's a real cost in this repository specifically. Two Windows-portability defects surfaced in the last few days — the `path.relative` backslashes in generated documentation links (#471) and the cross-drive `isAbsolute` gap in the docs-lookup containment check (#486) — and **neither was caught by CI**; both came out of reading the code. So the PR-time Windows job wasn't the thing catching these, which weakens the argument for keeping it, but it also means the remaining safety net is thinner than the job list suggests.

The consequence worth being deliberate about: a Windows-only break now lands on `master` and is fixed by a follow-up rather than prevented by a review gate. For a project whose release story is built on certification evidence, a red master is a different kind of problem than a red PR.

Worth confirming the interaction with the concurrency group added in #470 — `ci-${{ github.event.pull_request.number || github.run_id }}` with `cancel-in-progress: true`. Push events fall to `run_id`, which is unique per run, so consecutive merges won't cancel each other's Windows validation. That's the case this change depends on and it holds.

One fragility note on `release.yml`:

```yaml
include: >-
  ${{ fromJSON(github.event_name == 'pull_request' && '[…]' || '[…]') }}
```

Embedding matrix JSON in a folded scalar works, but a typo inside those strings becomes a runtime expression failure rather than a YAML parse error — and the two branches duplicate the macOS entry, so they can drift. Not worth blocking on; worth knowing when someone next edits the matrix.

PR #503, opened 2026-09-09
docs(installation): align beta.2 package examples

paulfalgout:
## What
- Pin the native-data, Backbone-adapter, and migration installation examples to `5.0.0-beta.2`.

## Why
The final release sweep found these examples still installing beta.1. The published beta.2 documentation should install the version it describes.

## Scope
Three documentation lines and their two generated evidence hashes. No runtime or API changes.

## Deployment Impact
Recertify the merged library source before npm/GitHub publication, then import that exact snapshot into the website/MCP release. No forms or form bundles apply.

## Testing
- `git diff --check` passed.
- `node scripts/api-contracts/check.mjs --write`, followed by `node scripts/api-contracts/check.mjs`, passed; the diff changes only the two documentation hashes.
- Searched documentation, package READMEs, the upgrade guide and consumer skill: no remaining exact beta.1 version pins.
- `npm run build` and `npm run docs:check` passed:30 executable examples,115 HTML files and1,731 internal links. The initial docs check needed the generated version file supplied by the build.
- Release-candidate checks run in CI. The previous unpublished candidate was cancelled; its evidence will not certify this new source.


paulfalgout:
**Review — fetched `43e9b426`** (against master `5de3396c`).

**Approved.** Four files, three prose lines and the two digests they invalidate. The change is exactly what it claims to be.

Verified:

- **`docs/data.api.md`** and **`docs/optional-backbone.md`** are both in the inventory's `evidence` digest map, and both digests moved. **`docs/migration-from-v4.md`** is not tracked, so no third digest was owed — worth stating explicitly, since "three files changed, two hashes updated" looks like an omission until you check the map. `node scripts/api-contracts/check.mjs` passing confirms it from the other direction.
- The three edited lines are all genuine install commands for the version the page documents. No behavioral or example-semantics change rides along.
- I re-ran the residual scan across `docs/`, `packages/`, `readme.md`, `skills/`, and `upgradeGuide.md`. No install pin still names beta.1 — the sweep's stated claim holds.

Two leftovers that the "exact version pins" framing correctly excludes but the *intent* ("the published beta.2 documentation should install the version it describes") arguably covers. Both are non-blocking, and both are version-specific prose rather than commands, which is why a pin search doesn't surface them:

1. **`upgradeGuide.md:13`** — "a source checkout can contain changes absent from beta.1." The sentence exists to warn that the published prerelease lags the source. Once beta.2 is on `latest`, beta.1 is the wrong referent: the statement stays technically true and stops being the useful one.
2. **`docs/development.md:80`** — "Do not combine an unreleased starter with registry beta.1 dependencies." This one is slightly more than staleness. The hazard is combining an unreleased starter with *any* registry version; naming beta.1 reads as though beta.2 registry dependencies would be a safe pairing, which is the opposite of the rule. The same file already says at line 10 that "This npm workflow starts with beta.2," so it names both releases in two different roles.

For both, phrasing the referent version-independently — "the published prerelease", "registry dependencies" — would fix the current inaccuracy and stop the line from needing an edit every release. That is the cheaper long-term shape than another sweep next time.

This is the second instance this week of the pattern I raised on #502: version strings that live outside the digest inventory drift silently, and are found by someone reading rather than by CI. `docs/data.api.md` and `docs/optional-backbone.md` were tracked — but tracking only proves a file changed without its hash being refreshed; it can't tell that `beta.1` inside the file has gone stale. The three lines this PR fixes were caught by a human sweep, not by a check. If beta.3 is coming, a single check asserting that no documentation install command names a version other than the one in `package.json` would convert the whole class from "remember to sweep" into a build failure, and would have caught all three of these plus both leftovers above.

CI: `Build and validate`, `Release preflight`, `Analyze`, `Performance timing`, CodeRabbit, and cubic all green on this head; `Node 24`, `Package smoke (macOS arm64)`, `Build verified tarballs`, and `Bundle size` still running. Nothing in a documentation-only diff should trouble them.

PR #504, opened 2026-09-09
docs(release): base stable v5 on contracts and usability

paulfalgout:
Related #128

## What
- Make stable v5 depend on verified public contracts, public migration and application evidence, fresh-agent maintenance work, and bounded stabilization.
- Add the usability evaluation procedure and align maintainer, beta, release-checklist, issue-template, and agent guidance. Keep comparative framework research separate from release acceptance.
- Remove retired baseline task classifications and the fixed paired-task floor from uncollected benchmark metadata and development tooling; retain task isolation and prototype capability-diversity checks.

## Why
The previous gates mixed dependable-release evidence with prescribed improvements over a historical Marionette revision. The accepted direction requires demonstrated application usability without making a comparative win or two complete application rewrites prerequisites for stability. The pilot must inform a frozen policy before a separate release evaluation; existing prototypes do not establish that evaluation is complete.

## Scope
Project strategy, contributor/consumer guidance, release preparation, and benchmark development tooling. No production library source, public runtime API, package version, publication authorization, or recorded historical result changes. Existing GitHub issue acceptance text still needs reconciliation; this PR does not complete #128 or run an agent evaluation.

## Deployment Impact
No application or form redeployment is required. Documentation publication follows the normal exact-revision import/deployment process after merge. This PR does not publish packages, deploy documentation, or run paid agents.

## Testing
- `npm run test:agent-benchmark-contract` — 37 tests passed, including corpus validation, hidden-test isolation, duplicate decision rejection, selected-workspace schema rejection, and unsuccessful-attempt handling.
- `npm run docs:check` — passed; 114 pages built, 115 HTML files and 1,734 internal links validated.
- `npm run check:api-contracts` — public contract inventory matches.
- `node node_modules/eslint/bin/eslint.js scripts/agent-benchmark/harness.mjs scripts/agent-benchmark/task-contract.mjs test/agent-benchmark/harness.test.mjs test/agent-benchmark/task-contract.test.mjs --max-warnings=0` — passed.
- `npm run test:fixtures -- --fixture docs-package` — passed after correcting a CI-discovered relative roadmap link in the packaged beta guide; includes the package build and installed-documentation checks.
- Production browser/runtime suites were not run locally because no production source changed; the PR workflows provide the broader checks. These checks establish documentation/tooling consistency, not agent usability or release readiness.

## Breaking changes
No public runtime API change. Uncollected benchmark planning and capability metadata move to schema 2 and discard the retired fields; there is no compatibility path for the obsolete prototype policy.

### DIFF
--- a/.github/ISSUE_TEMPLATE/v5-task.yml
+++ b/.github/ISSUE_TEMPLATE/v5-task.yml
@@ -1,11 +1,11 @@
 name: v5 implementation task
-description: Maintainer task for the agent-ready Marionette v5 release.
+description: Maintainer task for dependable contracts and demonstrated v5 usability.
 title: "[v5]: "
 body:
   - type: markdown
     attributes:
       value: |
-        Use this template for focused stable-v5 work. The task must be implementable and reviewable using public evidence, with an explicit production-cost boundary.
+        Use this template for focused stable-v5 work under the current ROADMAP.md. The task must be implementable and reviewable using public evidence, with an explicit production-cost boundary. Identify whether it supports a release requirement or separate comparative research; a comparative win is not a stable-release gate.
 
   - type: textarea
     id: goal
@@ -18,8 +18,8 @@ body:
   - type: textarea
     id: agent-failure
     attributes:
-      label: Agent-development failure addressed
-      description: Name the observed ambiguity, mistake, or verification gap.
+      label: Consumer or agent problem addressed
+      description: Name the observed contract defect, migration problem, usability failure, or verification gap.
     validations:
       required: true
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,5 +64,5 @@ Follow the public [synchronous failure contract](docs/view.lifecycle.md#synchron
 - Keep test, lint, benchmark, and diagnostic tooling outside production import graphs. Runtime cost remains part of review.
 - Use stable diagnostic codes for invariants. Do not make agent workflows depend on exact prose or undocumented maintainer knowledge.
 - New async CLI code must await work and propagate failures. Preserve concise failure output plus detailed machine-readable artifacts.
-- Report commands actually run, source revision, failures, and gaps. Coverage and a successful reference solution do not establish agent readiness; that requires the frozen benchmark and scored evidence.
-- Do not run paid agent benchmarks without the predeclared profile, model, permissions, run count, spend and elapsed-time budget required by #128.
+- Report commands actually run, source revision, failures, and gaps. Coverage and a successful reference solution do not establish agent usability; that requires the frozen usability evaluation and public application evidence in ROADMAP.md. Comparative research is separate from release readiness.
+- Do not run paid agent pilots or evaluations without a predeclared model/runner profile, permissions, run count, spend and elapsed-time budget, and authorization for that envelope. Follow [the evaluation plan](benchmarks/agent/evaluation-plan.md).
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,17 +1,25 @@
 # Agent-ready Marionette v5
 
-Date: 2026-09-08
+Date: 2026-09-09
 Status: Governing project strategy
 
 ## Decision
 
-Marionette v5 will be released as stable only when the framework is demonstrably
-good for AI-agent development and remains a clean, fast runtime.
+Marionette v5 should let people and agents build and maintain substantial
+applications predictably, with understandable code and reasonable effort. Stable
+`5.0.0` requires dependable public contracts and demonstrated usability through
+migration, public application work, and independent agent maintenance attempts.
 
-Pre-releases may continue while that work is in progress. `5.0.0` is the public
-promise that Marionette's contracts, documentation, diagnostics, types, and tooling
-are coherent enough for agents and humans to make correct changes without hidden
-application knowledge.
+Comparative superiority over earlier Marionette versions or other frameworks is a
+separate research claim, not a condition of stability. Prereleases may continue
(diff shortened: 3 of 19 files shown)


paulfalgout:
**Review — fetched `7d129093`** (against master `13f4954c`, after #503 merged).

**Approved.** This is a strategy change, and the direction is the maintainer's call — I'm reviewing whether the new text is internally consistent, whether the tooling matches it, and whether anything that used to be enforced silently stopped being enforced. Two of those are clean; the third has one real gap.

## The direction

Moving comparative superiority out of the release gate is defensible and, I'd argue, more honest than what it replaces. The old text conditioned `5.0.0` on beating a historical Marionette revision by ≥20 points or reaching ≥95 percent, plus a 50 percent reduction in cataloged architecture violations. Those are research outcomes. A framework can have dependable contracts, good docs, and working diagnostics and still not clear them — and if it didn't clear them, the old gate said don't ship, which is the wrong answer to that question. "Comparative superiority over earlier Marionette versions or other frameworks is a separate research claim, not a condition of stability" resolves that correctly, and pushing it into a named separate project that "may report mixed or negative results for Marionette without blocking a dependable release" is the version of that move I'd trust.

The anti-gaming clauses survive the rewrite intact, which is what I'd check first on a gate that's getting less prescriptive:

- "Do not narrow frozen acceptance after seeing failures; a changed scope or policy requires a new evaluation series."
- "An uncollected policy or incomplete evaluation cannot pass this gate."
- "Successful reference solutions and green tooling checks are not agent runs."
- "attribution does not erase an unsuccessful attempt."
- In the plan: freeze before collection, withheld tests inaccessible to the agent, new series on any frozen-input change, prior results preserved, paid runs separately authorized.

That last set is the part that actually does the work. A qualitative gate with freeze-then-honor discipline is worth more than a quantitative gate someone can re-cut after seeing the numbers.

**One substantive tension worth naming**, since it's the thing the old design got right and the new one leaves open. The old text used the pilot for *sample size only* — power to detect a 15-point regression — while the acceptance *level* was fixed independently at an 80 percent Wilson lower bound with no task below 60 percent. The new text says "Select counts and thresholds from the pilot's evidence" and the plan says to use pilot findings "to justify the release evaluation's repetitions, task-level and aggregate acceptance criteria." That makes the acceptance level itself a function of observed candidate performance. The plan's own guard — "Do not count pilot attempts in that evaluation or tune acceptance to a previously observed candidate score" — sits awkwardly next to it, because a pilot *is* a previously observed candidate score. Nothing here would stop a weak pilot from justifying a weak threshold, entirely in good faith and in writing.

Not a blocker, and not something to fix by reinstating the deleted numbers. But if the frozen policy is going to be the whole gate, it's worth stating in the ROADMAP that the pilot sets *precision* (repetitions, detectable effect) while some minimum acceptance level is chosen before the pilot is read. That preserves the flexibility you're buying here and closes the one path where the gate can be satisfied by lowering it.

## The tooling

- **`harness.mjs`**: replacing the hardcoded `< 2` with `catalog.coverage.minimumIndependentTasksPerCapability` is correct, and it isn't a fail-open — `loadCorpus` calls `validateTaskContracts` before the loop, and that path runs `validateCapabilityCatalog`, which pins `schemaVersion === 2`, the exact key set, and the value `2`. The ordering dependency is implicit rather than expressed, but it holds today.
- **`capabilities.json` schema 2** is enforced end to end: exact-key check, exact-value check, and the test fixture updated in step.
- **`series-decisions.json` schema 2 is not enforced by anything.** It moved from `schemaVersion: 1` to `2`, dropped four fields, and gained six (`purpose`, `runner`, `documentationAccess`, `acceptancePolicy`, `analysisPlan`, `stabilization`) — and the only check anywhere is `loadCorpus`'s assertion that `decisions.tasks` names every corpus task exactly once. No `schemaVersion` read, no key validation, no null-vs-missing distinction. A file still on schema 1 with `phase0Revision` and `runPairing` would load without complaint; so would one that silently lost `acceptancePolicy`. That's the asymmetry: the capability catalog, whose only content is two integers, gets a strict contract; the decisions file, which is where the entire frozen policy is supposed to be recorded before collection, gets none. Given that the ROADMAP now leans on "freeze these inputs before evaluation" as the primary discipline, the file that records the freeze is the one I'd want key-validated — including a check that fields are `null` (declared, undecided) rather than absent (forgotten). This is the same shape as the private-member inventory gap I've raised before: the check catches stale entries but not missing ones.
- **`minimumTasks: 10` removal** takes the paired-task floor with it, as intended — but it was also the only *policy-declared* floor on corpus size. What remains is the literal `assert.equal(corpus.tasks.length, 13)` in `harness.test.mjs`. That's fine while the corpus is a prototype; it does mean the floor now lives in a test rather than in the config the ROADMAP points at. Same family as the `absoluteCeilingBytes` removal in #435 — worth being a recorded position rather than a side effect.
- Test rename ("without historical comparison decisions") matches what the assertions now check. No orphaned assertions left behind.

## Consistency

Downstream references were updated rather than left dangling: `readme.md`, `AGENTS.md`, the issue template, `docs/beta.md`, `docs/maintainers/*`, and the API-shape gate's "paired agent tasks" → "executable contract checks" with the explicit "Contract discrimination does not require a historical-version agent comparison." The Scope section is candid that "Existing GitHub issue acceptance text still needs reconciliation; this PR does not complete #128" — that's the right disclosure, and it's the follow-up I'd want tracked, since #128's acceptance text currently contradicts the governing strategy file.

CI: `Build and validate`, `Release preflight`, `Analyze`, `Node 26 advisory`, `Bundle size`, and `Performance timing` green; `Node 24`, `Package smoke (macOS arm64)`, and `Build verified tarballs` still running.

PR #505, opened 2026-09-10
docs(agents): improve first-draft guidance and typed starter

paulfalgout:
## What
- Make the consumer guide and packaged Marionette skill teach templates, named Regions, data ownership, complete update subscriptions, focused-input preservation, and rendered personal-app review.
- Replace imperative content updates in the integration example and typed data starter with escaped templates, UI bindings, semantic triggers, and a separate status View.
- Release the starter's owned collection during workspace destruction.

## Why
Agents following the starter should encounter the same composition and ownership patterns expected in finished applications. Working buttons alone missed architecture, direct-data updates, draft preservation, and visual-quality checks; the guidance now identifies those separately.

## Scope
Consumer documentation, the packaged skill, and `test/fixtures/data-package-starter`. No changes to `src/`, `packages/*/src/`, runtime APIs, or production allocation costs. The accepted synchronous lifecycle failure contract is unchanged; this does not introduce synchronous recovery or rollback behavior.

## Deployment Impact
No forms or bundles need redeployment and no application deployment is required. These examples run on published beta.2. The updated packaged documentation/skill and distributed starter require the next package publication (beta.3 if that is the next release); the website can publish equivalent guidance independently now. This PR does not release packages, publish tags, or deploy documentation.

## Testing
Completed locally with Node 24.19.0:
- `npm ci`
- `npm run docs:check` — 30 executable documentation examples, 114 built pages, and 1,735 internal links validated.
- `npm run test:fixtures -- --fixture data-package-starter` — package build/pack plus isolated starter typecheck, lint, public-contract tests, and Vite build passed.
- Independently copied this starter into a fresh consumer directory with exact npm `marionette@5.0.0-beta.2` and `@mnjs/data@5.0.0-beta.2`; `npm install`, `npm run validate`, and `npm run test:browser` passed. All three browser checks passed: editable-row identity/draft preservation during reorder and navigation, latest-selection wins, and departure cleanup.
- `npm run check:api-contracts` — passed after refreshing only the edited integration guide's evidence hash; the first CI attempt identified the stale hash.
- Review regression: a missing row title reproduced `TypeError: Cannot read properties of undefined (reading replaceAll)` before the fix. The new public-behavior test now verifies missing/null row and detail fields render empty text, quotes/markup remain literal, and status still updates.
- After the review fix, reran the targeted package fixture and the published-beta.2 consumer validation (typecheck, lint, two Node tests, build, and all three browser tests); all passed.
- `npm run lint` — full repository lint passed after correcting the unnecessary escape in the new fixture test; both starter Node tests also passed again against published beta.2.
- `git diff --check`

The complete library verification suite was not run locally for these docs/fixture-only changes; CI provides broader checks. This is reference-path evidence, not a scored fresh-agent benchmark or a guarantee of beautiful generated applications.

## Breaking changes
None to published runtime APIs. Examples now canonically demonstrate composed Views and template-owned content.

## Summary by CodeRabbit

- **Documentation**
  - Expanded guidance for choosing application patterns, validating behavior, and building idiomatic Marionette applications.
  - Updated integration examples to demonstrate template-driven rendering and clearer application architecture.
  - Clarified ownership, escaped content, controls, and status updates in the starter application guidance.

- **Bug Fixes**
  - Starter application examples now safely escape rendered content and handle missing note values without creating unintended markup or handlers.
  - Improved repeated view attachment and cleanup behavior.

- **Tests**
  - Added coverage for empty values, literal content, HTML escaping, and workspace cleanup.

### DIFF
--- a/config/api-contracts/inventory.json
+++ b/config/api-contracts/inventory.json
@@ -8072,7 +8072,7 @@
   },
   "evidence": {
     "config/diagnostics/catalog.json": "a044dac8d4242de8e7c6c58a6442fb58384069c0f4d76ec6fabec03f34697eac",
-    "docs/choosing-integrations.md": "24ce6295dcc8cf8f587a52a935e2b351ebab4e9721e6ed311625c611eec0bc6c",
+    "docs/choosing-integrations.md": "385bfb2c7d85a8e0db71fd795ed1fef488aa637aa5094e93210810ff91f31f12",
     "docs/data.api.md": "c9bf72e4afd8c9b774391b3e343284808705887c8a38265609bfad910193e45c",
     "docs/diagnostic-catalog.md": "1df844c7e8edbea59a8c07776d656d45f77266108babb770d7690183fe85d180",
     "docs/dom.api.md": "a697a637ae4c6cc92aa5d72850b404763f08e0417800feacc0b431ec073d3916",
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -62,6 +62,33 @@ for repeated children. Use an Application when work has an asynchronous feature
 lifecycle. A plain function or class is enough when it needs none of these
 contracts. The [class guide](./classes.md) explains the boundaries.
 
+Before expanding a small example into an application, revisit its data and ownership
+choices. A recipe demonstrating manually managed children is not a default data
+architecture for a todo application. Domain records belong in a data source;
+CollectionView children are their presentation. Do not use child View traversal as
+the application's record store. Plain arrays can remain appropriate for explicit
+snapshot updates. When records need shared observation, filtering, and coordinated
+updates, select an observable provider; for a new application without one, start
+with `@mnjs/data` and its [DataApi setup](./data.api.md).
+
+Native DOM defaults describe the integration, not a replacement for View composition.
+Render ordinary content through `template` and `templateContext`, place child Views
+through named Regions, and declare controls with `ui`, `events`, and `triggers`.
+Use `events` when the handler needs input or keyboard details; use `triggers` when
+an interaction should become a View event. Direct DOM work still belongs at real
+integration boundaries, such as focus, measurements, and an external animation or
+widget. Keep its lifetime with the owning View.
+
+Observe each field used by a template or derived display, including changes that
+do not come from its main button. Let CollectionView handle membership changes
+without a second whole-list render subscription. Keep focused editors stable
+when processing their own input.
+
+When building teaching examples, make the application runnable independently of
+narration and inspection. Prefer ordinary modules with explicit imports and exports.
+Check that the preview supports the selected packages and module structure; a tiny
+sandbox's restrictions should not silently become the recommended app architecture.
+
 Configure the selected runtime before creating its consumers. The default named
 exports share a runtime. Use [runtime isolation](./runtime-isolation.md) when
 independent configurations must coexist; do not create a runtime per View.
@@ -106,6 +133,13 @@ Use a real browser when correctness depends on focus, attachment, DOM event
 propagation, or editable state. A build or screenshot alone does not prove those
 interactions. Use documented public APIs for assertions rather than private
 framework fields.
+
+Review the architecture separately from interaction results: identify the data
+source, the Views and Regions that own the screen, and the external work each owner
+releases. Working buttons do not establish that the example teaches those contracts.
+For observable data, change the source directly and verify all intended consumers
+update. For attachment-bound resources, also detach and reattach the same View;
+replacement alone does not prove that repeated attachment releases resources.
 
 When reporting a change, name the behavior, the tested package/source, the exact
 commands or interactions performed, and any untested boundary. Keep changes
--- a/docs/choosing-integrations.md
+++ b/docs/choosing-integrations.md
@@ -81,11 +81,8 @@ import { View } from 'marionette';
 import BackboneApi from '@mnjs/adapters/backbone';
 
 const AccountView = View.extend({
-  template: () => '<span class="name"></span>',
-  modelEvents: { change: 'render' },
-  onRender() {
-    this.el.querySelector('.name').textContent = this.model.get('name');
-  }
+  template: ({ active }) => `<span>${active ? 'Active' : 'Inactive'}</span>`,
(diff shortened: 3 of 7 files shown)


paulfalgout:
**Review — fetched `a5e61eb1`** (against master `6b328554`).

**Approved.** The rewrite lands the lesson it's after: content moves into templates, controls become `ui` + `triggers`, and the status display becomes its own View with its own subscription instead of a hand-written `querySelector` write from the shell. Full CI is green, and the PR body's independent install against published beta.2 is the right evidence for a starter whose whole job is to be copied out of `node_modules`.

Things I checked rather than assumed:

- **`state: this.getState()` on the new `Status` View is correct sharing, not a leak.** `StateMixin._initState` sets `this._state` directly from the option and leaves `_ownsState` unset, so `_destroyState` runs the event cleanup but does **not** call `State.disposeOwned`. The shell keeps ownership of the state it created; the borrowed child releases only its own subscription. That is the distinction the added AGENTS.md line ("borrows shell-owned state") is claiming, and the code actually earns it.
- **`notes.destroy()` ordering is right.** `region.destroy()` first tears down Shell → List → Rows, so the CollectionView is gone before the collection it observes is destroyed. Reversing those two would have the CollectionView reacting to its source's teardown.
- **`Detail`'s plain-object `model` is a supported shape, not a misuse.** `@mnjs/data`'s `DataApi.serialize` returns non-`Model` values verbatim, and `get`/`has` fall back to `Object.hasOwn`, so POJOs are first-class here. I went looking for a trap and there isn't one — which leads to the next point.

Three findings, one of which I'd fix before merge:

**1. The rewrite dropped a nullish guard, turning a missing attribute into a throw during render.**

The old `Row` was `this.input().value = this.options.model.get('title') ?? ''`. The new one is `value="${escapeHTML(title)}"`, and `escapeHTML` calls `.replaceAll` on its argument. `serializeData()` returns `model.attributes`, so a row model constructed without a `title` yields `undefined` and the template throws a `TypeError` mid-render instead of rendering an empty input. `NoteRow` types `title` as required, but a `Model<NoteRow>` is a runtime bag — `notes.add(new Model({ id: 'x' }))` type-checks fine at the call site and blows up inside `_renderTemplate`. `Detail` has the same exposure through `loadNote`, which is consumer-supplied: a loader returning `{ title }` without `body` throws where the old code rendered (badly, but harmlessly).

This matters more here than it would in ordinary application code, because a throw inside render is the one path this framework handles least gracefully — `_isDestroying` is set at three sites in `src/` and cleared at none, so an object that throws during a lifecycle step stays wedged and every retry returns success off the first-line guard.

One line in the helper closes the entire class, for every copy of this starter:

```ts
const escapeHTML = (value: unknown) => String(value ?? '')
  .replaceAll('&', '&amp;')
  ...
```

**2. `Detail.templateContext()` is dead weight.**

```ts
initialize(options: { model: Note }) { void options; },
templateContext() { return this.options.model; },
```

`model` is in both `ViewClassOptions` and `ViewOptions`, so `this.model` is already the note; `serializeData()` → `Data.serialize(this.model)` → the same object, because it isn't a `Model`. `mixinTemplateContext` then merges that object over itself. The template receives identical data with the `templateContext` line deleted.

In a normal file that's a shrug. In *this* file it's the wrong lesson: an agent reading it learns that reaching a POJO into a template needs an explicit `templateContext` hop, when the framework's own serialization already does it. Deleting those two lines makes `Detail` a `template` and nothing else, which demonstrates the actual contract.

(`Status` genuinely needs its `templateContext` — it has no `model` or `collection`, so `serializeData()` returns `undefined` and the context is the only source. Worth using `this.getState().toObject()` there rather than `this.options.state.toObject()`: same object, but through the accessor, so it keeps working if that View ever creates its own state instead of borrowing one.)

**3. Every consumer now hand-rolls HTML escaping, and `@mnjs/utils` exports no helper for it.**

I checked `packages/utils/src/index.ts` — 21 exports, none of them an escape. The thesis of this PR is that templates own content, which means every interpolation in every consumer application is now an escaping decision. Shipping the answer as copy-paste in a starter means each project carries its own version and each one can be subtly wrong (this one is correct for text and for double-quoted attributes; it is not correct for unquoted attributes, URLs, or `<script>`/`<style>` context, and nothing in the starter says so).

A named export — even just this function — would give the guidance a thing to point at instead of a snippet to reproduce. Not this PR's job, but it's the natural follow-on to making template interpolation the default.

**On `docs/choosing-integrations.md`:** replacing `${model.get('name')}` with `${active ? 'Active' : 'Inactive'}` is the right call and worth stating explicitly, because the obvious minimal edit — moving `name` into the template string — would have shipped an unescaped model interpolation in the canonical example, teaching exactly the XSS pattern the starter now escapes against. Interpolating a literal sidesteps it. The one cost: the surrounding prose still says "a View may use Backbone data with native DOM operations and a plain template function," and after this change no example in the section shows the native DOM half. Minor, and arguably intentional given the PR's direction.

Everything else reads clean. The `triggers`/`onClickOpen` conversion is safe (`type="button"`, so default prevention is moot), `getUI('input')?.[0]` keeps the same throw-on-missing contract as the old `querySelector` guard, and the Row still has no `modelEvents`, so `model.set('title', …)` in the open handler can't re-render the row out from under a draft.

paulfalgout:
Addressed the starter findings in 49357a4e.

- `escapeHTML` now normalizes nullish values to empty text before escaping. A new public-API regression reproduced the missing-title render failure before the fix, then passed for missing/null fields and quoted markup in both row attributes and detail text.
- Removed Detail's redundant `initialize` and `templateContext`; its plain-object model uses the configured DataApi serialization directly.
- Status reads through `getState()`. Beta.2 infers the borrowed option type separately from the method receiver, so the context method declares its state accessor type explicitly; no runtime guard or allocation was added.
- Documented that the local escaping helper covers text and quoted HTML attributes, not URLs, scripts, or styles. A public utility export remains a separate API decision. No production runtime or synchronous failure/recovery contract changes are included.

Validation passed: `npm run test:fixtures -- --fixture data-package-starter`, `npm run check:api-contracts`, and independent published-beta.2 consumer validation (`npm run validate`: types, lint, two Node tests, build; `npm run test:browser`: all three passed). CI is rerunning for the new commit.

PR #522, opened 2026-09-15
fix(application): scope state events to running lifecycle

paulfalgout:
Related #190
Related #191

## What

- Deliver Application `stateEvents` only while the Application is running. Startup initialization can mutate state without superseding startup, and stopped state changes cannot restart the feature through its event map.
- Preserve state identity and subscriptions across stop/restart. Delivery follows the existing lifecycle state, including restoration after failed or canceled stop readiness; destruction still releases subscriptions before owned-state disposal.
- Add public lifecycle and Backbone integration tests, a packed-package browser regression, and matching state/Application documentation, upgrade guidance, changelog, and recorded API semantics.

## Why

A restartable feature can map a filter change to `restart` and seed that filter in `onBeforeStart`. Previously that initialization invoked the handler and superseded its own startup. The same subscription could start the feature again when its state changed after stop. Application handlers now follow activation while the source remains persistent.

## Scope

Application state-event delivery changes. Shared state subscription plumbing accepts an internal activation predicate; other state-owning classes keep their existing delivery behavior. Explicit `listenTo`, Radio bindings, child ownership, and lifecycle operation results are unchanged.

Subscriptions are installed after `initialize` and retained until destruction. Each configured Application state handler adds one wrapper and a running-state check; there are no new lifecycle flags, resubscription work on restart, or per-instance resources for Applications without state events.

Readiness notifications are suppressed, not queued or replayed. `onStart` reads current state for initial display. A filter change during a fetch does not automatically replace a request made with an earlier filter snapshot; the documentation describes explicit loading and refresh policies.

The supported synchronous failure boundary is unchanged: valid adapters and working cleanup callbacks are required. This PR adds no registration rollback, attempt-all cleanup, or synchronous recovery contract.

## Deployment Impact

No forms or bundles require redeployment as part of this library PR. Consumers receive the behavior when they upgrade to a release containing it. This prepares the contract for beta.3 without changing package versions, publishing packages, or deploying documentation.

## Testing

Passed locally on Node 24.19.0 / npm 11.17.0:

- `npm run build` — runtime builds, source types, generated declarations, and consumer type checks.
- `npm run coverage` — all 2,043 tests pass; existing coverage thresholds pass and all three changed production files have 100% statements, branches, functions, and lines.
- `npm run test:browser -- test/browser/application-state-events.spec.mjs` — Chromium, Firefox, and WebKit pass against packed packages, with real filter control interaction, readiness changes, restart, stable state identity, stopped changes, and subsequent startup.
- `npm run lint`
- `npm run check:public-tests`
- `npm run check:api-contracts`
- `npm run test:source`
- `npm run docs:check`
- `git diff --check`

Unit coverage includes native payload/context preservation, startup seeding, child readiness, canceled and failed startup, failed/canceled stop restoration, failed restart, independent observers, and one-time cleanup before owned disposal. The complete browser suite and release validation remain CI checks; only the new browser regression was run locally.

---
## Summary by cubic
Scopes Application `stateEvents` so handlers run only while the Application is running. Startup initialization and stopped mutations are retained in state but no longer invoke handlers, so a filter change can no longer supersede its own startup or restart a stopped feature.

- State identity and subscriptions persist across stop and restart, including restoration after failed or canceled stop; subscriptions release at destruction before owned-state disposal.
- Notifications suppressed during readiness or stopped time are not delivered or replayed; read current state in `onStart` for initial display.
- Explicit `listenTo`, Radio bindings, child ownership, and other state-owning classes keep their existing delivery behavior.
- Includes unit, Backbone adapter, and packed-package browser regression tests and updates docs, upgrade guide, changelog, and API contract semantics.

**Migration**

- Apps that depend on a filter change during loading to restart a fetch need an explicit policy, such as disabling controls until readiness completes or coordinating readiness cancellation.

<sup>Written for commit a3a7981d7b39feeaa04ea2ae67390325b03ad4b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/522?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Application state events now run only while the application is running.
  - State identity and subscriptions persist across stop/restart cycles.
  - Changes made during startup or while stopped are retained but not delivered or replayed.
  - State event subscriptions are released when the application is destroyed.

- **Documentation**
  - Updated application state documentation and upgrade guidance to describe lifecycle behavior and migration considerations.

- **Tests**
  - Added coverage for lifecycle transitions, readiness, restart behavior, persistent state, and Backbone integration.


paulfalgout:
**Review — fetched `a3a7981d`** (against master `0a5567df`).

**Approved.** Small, well-placed change with documentation and tests that are noticeably more careful than the diff size suggests.

**The gating window is correct, and I checked it rather than trusting the description.** My first concern was asymmetry: `before:start` runs before `_lifecycleState = RUNNING` (line 479), so startup seeding is suppressed — but `before:stop` is awaited at line 489 while `_lifecycleState = STOPPED` doesn't land until line 510, which looked like a long RUNNING window during stop readiness where a handler mapping to `restart` could fire and race its own teardown. It isn't. `beginOperation` sets `_lifecycleState = state` to the transitional state before running the operation callback, so the entire stop window is already non-RUNNING. Handlers fire only in the steady running state, never during any transition, and that is symmetric on both ends.

The lifecycle test pins exactly this: nine triggers spanning construction, `onBeforeStart`, startup readiness, `onStart`, running, `onBeforeStop`, stop readiness, `onStop`, stopped, and destroyed — with `handler.mock.calls` asserted to be exactly `[['start'], [state, 42]]`. Asserting the full delivered sequence rather than a count per phase is what makes this test able to fail if the window moves in either direction.

Three design choices worth naming because each could have gone wrong the other way:

- **The predicate reads `_lifecycleState` live at event time** rather than caching an enabled flag at subscribe time. That is why the failure and cancellation paths (`_lifecycleState = operation.failureState` at 403/411, and the restore in `beginOperation`) restore delivery with no extra bookkeeping, and why restart needs no resubscription. A cached flag would have required a parallel invalidation path through every one of those.
- **`isActive` uses the private `isApplicationRunning(application)`, not the public `isRunning()`.** Gating that a consumer could subvert by overriding `isRunning()` would be a worse contract, and refactoring `isRunning()` to delegate to the same helper keeps them consistent by default without making the gate overridable.
- **Suppressed, not queued.** Replaying startup-time changes into handlers after `onStart` would reintroduce the superseding problem with extra steps. "Read current state in `onStart`" is the right instruction and it is in both docs.

Documentation covers the parts that will actually bite. `docs/marionette.state.md` states the scope limit explicitly — "The delivery boundary affects Application handlers only; other owners of the same source continue observing it normally" — which matters because `stateEvents` now means something different on an Application than on a View, under one option name. And `docs/marionette.application.md` is candid about the case the change does not solve: a filter changed during readiness will not restart a fetch made with an earlier snapshot, with three concrete policies offered instead of pretending the framework handles it. `config/release-validation.json` registers the new browser spec, so the check is routed rather than existing only as a file.

**One narrow consequence worth a note somewhere: the StateApi adapter contract quietly got stricter.**

In the gated path, `subscribeBindings` no longer passes the user's own function to `Api.subscribe` — it passes an anonymous arrow wrapper and applies the real callback itself. For the built-in and Backbone adapters this is fine (cleanup flows through the returned function, and `backbone-adapter.spec.js` covers it). But a custom StateApi whose `subscribe` identifies handlers by reference — anything that reconstructs an `off(source, name, callback)` call rather than closing over its own cleanup — will now be handed a function the consumer never wrote, only for Applications with `stateEvents`. That is a supported extension point, and the requirement "your `subscribe` must return a working cleanup and must not assume it receives the consumer's function" is currently implied rather than stated. One sentence in the StateApi section would make it a contract instead of a discovery.

Also minor: `(...args: never[])` on the wrapper reads oddly next to `callback.apply(context, args)` — `unknown[]` describes what actually flows through. No behavioral difference.

CI green across the board, including the three-browser packed-package spec.

paulfalgout:
Closing this PR because the running-only Application stateEvents policy is superseded by the broader Application lifecycle assessment.

The startup self-restart and post-stop reaction problems are real, and the reproductions, state-identity checks, adapter coverage, and cleanup tests here remain useful evidence. However, gating stateEvents alone leaves loading-time reactions and the lifetimes of Radio handlers, explicit listeners, timers, and requests unresolved. We should prove one coherent subscription and activation pattern before establishing another default lifecycle policy.

The follow-up work will cover explicit effect activation and cleanup, child activation and recursive teardown, refresh/navigation that preserves intended UI, and an integrated public application proof. This closes the proposed implementation without claiming those underlying problems are resolved. Related issues #190 and #191 are not closed by this decision.

PR #533, opened 2026-09-16
feat(application): separate preparation from lifecycle notifications

paulfalgout:
## What

- Separate Application readiness into optional `prepareStart`, `prepareStop`, and `prepareDestroy` methods. Each receives `(options, { signal })` with the Application as `this`.
- Keep `onBefore*` / `before:*` and completion callbacks as synchronous notifications. Forward the resolved startup result unchanged as one third argument to `onStart` / `start`; operation promises still resolve booleans.
- Migrate lifecycle tests, executable guides, type consumers, benchmark workloads/reference examples, and the public contract inventory.

```js
const SessionApplication = Application.extend({
  onBeforeStart(app, options) {
    showLoadingIndicator();
  },

  prepareStart(options, { signal }) {
    return loadSession({ signal });
  },

  onStart(app, options, session) {
    this.showView(new SessionView({ model: session }));
  }
});

const application = new SessionApplication({ region: '#app' });
await application.start(); // true when running; false when superseded
```

## Why

Through beta.3, returning a Promise from `onBeforeStart` delayed readiness, while returning the same Promise from a `before:start` listener did nothing. The resolved startup value was also discarded. Separate preparation methods make awaited work explicit and let ordinary Marionette notifications follow one consistent return-value contract.

This is a breaking API change for the next prerelease after the already published beta.3. Move asynchronous `onBeforeStart`, `onBeforeStop`, and `onBeforeDestroy` work into the corresponding preparation method; before notifications no longer receive the abort context. Existing synchronous notification work stays in place. Toolkit's `beforeStart` maps to `prepareStart`; use explicit `Promise.all` when collecting multiple results, which arrive as one array value.

## Scope

Application lifecycle dispatch and public types, associated documentation, tests, and development examples. Existing cancellation, adopted stop readiness, child ownership, restart, failure states, and terminal destruction remain in force. A before notification that supersedes its operation prevents that operation's preparation from starting.

Stop and destroy preparation values are awaited but otherwise ignored. There are no compatibility aliases, new per-instance resources, or global registries. View, Region, and MnObject lifecycle behavior, app-frontend adoption, and website example expansion are outside this PR.

## Deployment Impact

No forms or bundles need redeployment for this library PR. No package publication or site deployment is included. Consumers adopting the next package version must migrate asynchronous Application notification hooks as described above. Published beta.3 artifacts and release authorization are unchanged.

## Testing

- Review follow-up: `npm run test:dist` checks 18 notification/preparation rejection scenarios in isolated Node subprocesses across CommonJS and ESM. Rejected notifications reach the host handler; preparation rejections reject the lifecycle operation.
- Review follow-up: the `docs-application-state` installed fixture and all 47 focused Application preparation/child-lifecycle tests pass. Migration guidance now explicitly states error ownership and skipped preparation for already-stopped owners; the example trace runs in `onBeforeStart`.

- `npm run coverage`: 2,043 tests pass; existing per-file coverage thresholds pass, with 100% function coverage.
- `npm run build`: source types, generated ESM/CommonJS declarations, and type consumers pass. Result inference, inherited preparation, native subclasses, and invalid result types are covered.
- `npm run test:source`, `npm run test:dist`, `npm run check:public-tests`, `npm run check:api-contracts`, `npm run lint`, and `npm run docs:check` pass.
- Four installed consumer fixtures pass against the same five locally packed tarballs: `core-types`, `docs-application-guides`, `docs-routing`, and `docs-package` (`npm run test:fixtures -- --fixture <name>` with all five artifact flags).
- `npm run test:browser -- test/browser/application-effects.spec.mjs test/browser/application-child-activation.spec.mjs test/browser/application-refresh.spec.mjs test/browser/application-prepared-view.spec.mjs`: 15 cases pass across Chromium, Firefox, and WebKit.
- `node --test test/browser/docs-routing.test.mjs`: all six native/Backbone routing cases pass across the three engines.
- `npm run agent:reference -- --task borrowed-workspace-state` and `--task owned-workspace-state`: both reference controls pass; no paid agent evaluation was run.
- `npm run test:browser-performance` and `node scripts/performance/retention.mjs --batches 2 --cycles 10 --output /tmp/mn-preparation-retention.json` pass. This is workload/retention validation, not a new performance baseline.

New public tests cover notification/preparation order and receiver, ignored notification promises, synchronous notification failures, synchronous/async startup results, array identity, cancellation before preparation, and stale result suppression after restart.

Claude reviewed the complete authored diff and lifecycle source before publication. No core correctness defects were found. Its runtime-check conditions are satisfied; the suggested explicit two-argument stop/destroy completion tests were added.

A follow-up Claude review and a public reproduction against both beta.3 source and this PR confirmed that synchronous restart/destroy from a before:stop notification begins a fresh stop attempt in both versions. Added tests for notification counts, supersession results, and replacement-only preparation; no new reentrancy machinery was added.

## Summary by CodeRabbit

- **New Features**
  - Added awaited lifecycle preparation methods for starting, stopping, and destroying applications.
  - Startup preparation results are passed to the start completion callback.
  - Lifecycle preparation receives operation options and cancellation context.

- **Breaking Changes**
  - Lifecycle notifications are synchronous; returned Promises no longer delay operations.
  - Asynchronous readiness work must use the corresponding preparation method.
  - Preparation callbacks no longer receive the application instance as an argument.

- **Documentation**
  - Updated lifecycle, migration, event, and cancellation guidance with the new contract.


src/modules/application.ts
```diff
@@ -444,24 +448,30 @@ async function startApplication(application: ApplicationInternals, operation: Op
...
 }
 
 async function stopApplication(application: ApplicationInternals, operation: Operation, options: unknown, notify = true) {
   try {
     if (!operation.stopReadiness) {
       const readiness = beginReadiness(operation, options, async context => {
-        if (notify) { await application.triggerMethod('before:stop', application, options, context); }
+        if (notify) {
+          application.triggerMethod('before:stop', application, options);
+          if (!isCurrentOperation(application, operation)) { return false; }
```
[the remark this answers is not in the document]
paulfalgout: Preserving the existing behavior here. I reproduced this against both the beta.3/master Application source and this PR: synchronous restart/destroy from a once-bound before:stop listener emits two before:stop notifications, resolves the original stop to false, and completes the replacement successfully.

The notification runs before stop preparation begins and before stopReadiness is assigned. The replacement therefore begins its own stop attempt; the superseded attempt never invokes prepareStop. Adopting that unstarted preparation would conflict with the new skip-preparation rule. Existing in-flight async stop-readiness adoption remains covered separately.

48b3534 adds public tests for both replacement kinds, including notification arguments/counts, replacement-only preparation, operation results, and a single before:destroy notification. The guide now distinguishes synchronous replacement from adoption of preparation already in flight. Deduplicating these synchronous notification attempts would introduce a separate reentrancy guarantee, so no runtime machinery is added in this PR.

src/modules/application.ts
```diff
@@ -444,24 +448,30 @@ async function startApplication(application: ApplicationInternals, operation: Op
   // Restart has finished deactivation; explicit child starts are now allowed.
   application._lifecycleState = STARTING;
   const readiness = beginReadiness(operation, options, context => {
-    return application.triggerMethod('before:start', application, options, context);
+    application.triggerMethod('before:start', application, options);
```
paulfalgout, line 452: **Rejected notification Promises become unhandled rejections.**

The upgrade guide says returned Promises are "ignored", but discarding a *rejected* Promise is not ignoring it. `packages/utils/src/trigger-method.ts:37` returns the handler result without awaiting it or attaching a handler, so nothing ever settles it.

Verified against master today, with `onBeforeStart() { return Promise.reject(new Error('boom')) }`:

    await app.start()   // rejects with 'boom'

After this change the value is discarded here, so `start()` resolves `true`, the application reaches RUNNING despite failed startup work, and the rejection surfaces as a process-level `unhandledRejection` — which terminates the process by default on Node 15+. That is a sharp edge for exactly the beta.3 consumers the guide is addressed to: their app used to fail startup cleanly and now may take the process down.

Same at `:471` for `before:stop` and `:619` for `before:destroy`.

If the intent is genuinely "ignored", attaching a no-op catch would make it literally true:

    void Promise.resolve(application.triggerMethod('before:start', application, options)).catch(() => {});

Flagging confidence: I verified the current behaviour and the `triggerMethod` mechanism directly, but the PR branch needs a build to run its tests, so the post-change half is read from the diff rather than executed. Worth confirming against a built branch before acting.

test/unit/application-preparation.spec.js
```diff
@@ -0,0 +1,197 @@
...
+    ]);
+    expect(calls[3][1][2]).toBe(result);
+  });
+
+  for (const [operation, before, prepare] of [
+    ['start', 'onBeforeStart', 'prepareStart'],
+    ['stop', 'onBeforeStop', 'prepareStop'],
+    ['destroy', 'onBeforeDestroy', 'prepareDestroy']
+  ]) {
+    it(`${operation} ignores notification Promises and awaits only its preparation method`, async() => {
```
paulfalgout, line 57: **This test does not cover the rejected case.**

It returns `notification.promise` from both the `onBefore*` method and the listener, then calls `notification.resolve()` on its last line — so it exercises pending and resolved notification Promises only.

The rejected notification is the case with a harmful failure mode (see the comment on `src/modules/application.ts:451`), and nothing in this file covers it. A future change that started awaiting notifications again would also still pass this test.

The rest of the file is notably thorough — supersession during notification, synchronous notification failure, result forwarding, and the no-preparation-method case are all covered — which is why this gap stands out.

paulfalgout (commented):
Review of the preparation/notification split. The design is coherent and the new spec is unusually thorough — supersession, synchronous failures, result forwarding and the absent-method case are all covered. Three findings below, one of which is the PR's own thesis not holding in one branch.

src/modules/application.ts
```diff
@@ -444,24 +448,30 @@ async function startApplication(application: ApplicationInternals, operation: Op
...
-  application.triggerMethod('start', application, options);
+  application.triggerMethod('start', application, options, result);
 }
 
 async function stopApplication(application: ApplicationInternals, operation: Operation, options: unknown, notify = true) {
   try {
     if (!operation.stopReadiness) {
       const readiness = beginReadiness(operation, options, async context => {
-        if (notify) { await application.triggerMethod('before:stop', application, options, context); }
+        if (notify) {
```
paulfalgout, line 471: **`prepareStop` is skipped whenever the notification is suppressed**, which couples the two concerns this PR exists to separate.

`prepareStop` sits inside `if (notify)`, and `stopApplication` is called with `notify = !wasStopped` from both `stop()` (`:577`) and `restart()` (`:594`).

So restarting an application that is already STOPPED but still owns children takes the `shouldStop` path with `notify === false`: `stopChildApps` runs, `prepareStop` never does. A consumer using `prepareStop` to release resources or wire the abort signal for teardown silently gets nothing, while `prepareStart` still runs on the subsequent startup.

Neither `upgradeGuide.md` nor `docs/marionette.application.md` mentions this — the docs present notification and preparation as separate phases, and the table at `docs/marionette.application.md:118` lists them side by side with no note that one gates the other.

If it is deliberate, it is worth a line in the docs; if not, moving the `prepareStop` call outside the `if (notify)` block matches the stated model.

docs/marionette.application.md
```diff
@@ -315,7 +338,7 @@ const SearchApplication = Application.extend({
 });
 
 const RootApplication = Application.extend({
-  async onBeforeStart(app, options) {
+  async prepareStart(options) {
```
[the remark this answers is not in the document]
paulfalgout: Fixed in 3b68f50: the root:before:start trace now runs in onBeforeStart, while prepareStart only awaits the required child. The installed docs-application-state fixture passes with the existing expected notification order.

src/modules/application.ts
paulfalgout: Confirmed the migration hazard and clarified it in the Application guide, upgrade guide, and semantic contract in 3b68f50. Returned notification Promises are neither awaited nor given rejection handlers; required work must move to prepare*, and background work owns its errors. We are intentionally not adding a no-op catch: it would suppress method errors while leaving event-listener rejections untreated. New isolated subprocess checks exercise rejected before methods and listeners for start/stop/destroy in both CJS and ESM. They verify successful lifecycle completion plus the host unhandled-rejection observation, and contrast that with rejected preparation, which rejects the operation without an unhandled rejection.

test/unit/application-preparation.spec.js
paulfalgout: Added rejected-notification coverage in test/dist/application-notifications.mjs, invoked by npm run test:dist in isolated Node processes for both CJS and ESM. All 18 method/listener/preparation rejection scenarios pass. Isolation keeps deliberate host unhandled rejections outside Vitest. One correction to the existing-test assessment: notification.resolve() occurs only after await pending, so awaiting notification returns again would fail or stall that test, not pass. The new checks additionally detect rejection swallowing and notification failures incorrectly becoming readiness failures.

src/modules/application.ts
paulfalgout: This is deliberate idempotence for the owner, now explicit in the Application guide, upgrade guide, and semantic contract in 3b68f50. An already-stopped owner has no local stop preparation to repeat, while its active descendants still run their own stop lifecycle. Existing public tests cover stop/restart/destroy draining a running grandchild through stopped owners and restart adopting descendant readiness without repeating the owner preparation; all 47 focused preparation/child-lifecycle tests pass. Moving prepareStop outside the condition would change that accepted behavior, so the runtime remains unchanged.

docs/events.class.md
```diff
@@ -46,52 +46,49 @@ proxied view events receive the host view.
 
 ## Application Events
 
-Application events describe its asynchronous lifecycle. Use a readiness method
-when completion must wait for work; event-listener return values are not awaited.
+Application lifecycle notifications are synchronous, like other Marionette events.
+Their corresponding `onBeforeStart`, `onStart`, `onBeforeStop`, `onStop`,
+`onBeforeDestroy`, and `onDestroy` methods are also synchronous notifications.
```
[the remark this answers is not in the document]
paulfalgout: Checked the current PR head and the reviewed 1a5368ea revision: this paragraph lists the six notification methods but does not say "seven" or give a numeric count. No documentation change is needed; resolving this as a false positive.

docs/marionette.application.md
```diff
@@ -95,54 +95,82 @@ transitions, after stop, and after destroy.
...
+are not consumed: returned Promises are neither awaited nor given rejection
+handlers. Use `prepareStart`, `prepareStop`, and `prepareDestroy` for work the operation must await. They are optional instance
+methods, called with `this` as the Application and `(options, context)` arguments.
+A synchronous return also completes preparation; a throw or rejected Promise
+rejects the operation. Notification callbacks must handle any asynchronous work
+and its errors themselves. An unhandled rejected notification Promise can surface
+as a host-level unhandled rejection even when the lifecycle operation succeeds;
+returning it does not make its failure a readiness failure.
+
+`prepareStart`'s resolved value is passed unchanged as one third argument to
```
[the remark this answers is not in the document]
paulfalgout: Fixed in 1b06c71f: the result is passed unchanged "as the third argument". npm run docs:check passes.

src/modules/application.ts
```diff
@@ -79,8 +82,9 @@ export interface ApplicationInstance<Options extends object = object, State = ob
   getView(): SupportedView | undefined;
 }
 
+type StartResultFor<Props> = Props extends { prepareStart: (...args: never[]) => infer Result } ? Awaited<Result> : unknown;
```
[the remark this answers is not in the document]
paulfalgout: Fixed in 1b06c71f. Required preparation methods keep their precise awaited result; optional methods infer the awaited result union undefined because the hook may be absent. Missing hooks retain unknown. Public package consumer tests cover valid/invalid results, required versus optional undefined, and inherited optional hooks. npm run build (including test:types), npm run lint, npm run docs:check, and npm run check:api-contracts pass. This changes declarations only, with no runtime changes.

PR #534, opened 2026-09-16
docs(skill): clarify packaged docs helper location

paulfalgout:
## What
- Identify the helper relative to `SKILL.md`, distinguishing the npm package's `dist/agent-skill/` from a copied skill directory.
- Use an absolute skill-directory placeholder so commands work from the application directory or any other working directory.
- Execute the documented commands from both packaged and copied skills; verify distinct list/page modes, the page list, and the complete requested page body.

## Why
The old `/path/to/marionette/scripts/docs.mjs` placeholder could be mistaken for a path relative to the npm package root. The published helper lives under `dist/agent-skill/scripts/docs.mjs`; a copied skill has its own directory.

## Scope
Consumer skill instructions and the existing docs-package fixture. No helper implementation or library runtime changes. The fixture accepts quoted or unquoted placeholders without introducing a general shell parser.

## Deployment Impact
No forms or application redeployment is required. Consumers receive the corrected skill in the next package release or by recopying it from that release. This PR does not publish packages.

## Testing
- `npm run test:fixtures -- --fixture docs-package` — passes, including the package build and both documented commands in both skill locations, executed from the application directory. Existing explicit `--package-root` coverage remains.
- `npm run docs:check` — passes.
- `npm run lint` and `npm run check:api-contracts` — pass.
- Skill Creator's `quick_validate.py skills/marionette` — passes using `uv run --with pyyaml python`.
- Claude review was attempted on the initial change but authentication failed before review began; no Claude feedback was obtained.

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation lookup instructions to work from any working directory.
  * Clarified paths for installed and copied skill locations.
  * Added absolute, quoted path examples for lookup commands.

* **Tests**
  * Added validation for documented lookup commands across both supported skill locations.
  * Verified list and page results against the application’s documentation and manifest data.


test/fixtures/docs-package/validate.mjs
```diff
@@ -58,6 +58,21 @@ const installedSkill = resolve(packageRoot, 'dist/agent-skill');
...
+      .flatMap(token => token.text.split('\n'));
+    assert.equal(commands.length, 2, 'Exercise both documented lookup commands');
+    for (const command of commands) {
+      const [executable, ...args] = command.split(/\s+/);
+      assert.equal(executable, 'node');
+      const output = execFileSync(process.execPath,
+        args.map(arg => arg === '"/path/to/application"' ? process.cwd() : arg),
+        { cwd: skillRoot, encoding: 'utf8' });
+      assert.ok(output.includes(manifest.sourceRevision));
+      assert.ok(output.includes('docs/agents.md'));
```
paulfalgout: **Both commands get the same two assertions, and both pass on the provenance header alone.**

For `--page docs/agents.md`, `docs.mjs` prints a JSON provenance line first — containing `sourceRevision` *and* `"source":"docs/agents.md"` — and only then the page body. So both asserts are satisfied before the body is printed at all.

Concretely: drop the second `console.log` in `docs.mjs`, or have it print a different page's bytes, and this loop still passes. It proves each command exits 0 and emits provenance, which is the same thing for both — so the assertion labelled *"Exercise both documented lookup commands"* does not actually distinguish them.

The sha256 check inside `docs.mjs` limits the blast radius for wrong content, so this is a coverage gap rather than a live bug. Asserting a known substring of the agents page body in the `--page` case would close it.

test/fixtures/docs-package/validate.mjs
```diff
@@ -58,6 +58,21 @@ const installedSkill = resolve(packageRoot, 'dist/agent-skill');
...
+  for (const skillRoot of [installedSkill, directory]) {
+    const skill = await readFile(resolve(skillRoot, 'SKILL.md'), 'utf8');
+    const commands = parser.lexer(skill).filter(token => token.type === 'code' && token.lang === 'sh')
+      .flatMap(token => token.text.split('\n'));
+    assert.equal(commands.length, 2, 'Exercise both documented lookup commands');
+    for (const command of commands) {
+      const [executable, ...args] = command.split(/\s+/);
+      assert.equal(executable, 'node');
+      const output = execFileSync(process.execPath,
+        args.map(arg => arg === '"/path/to/application"' ? process.cwd() : arg),
```
paulfalgout: **The substitution depends on the placeholder keeping its double quotes.**

`arg === '"/path/to/application"'` matches only the quoted form. `SKILL.md:21` tells the reader to replace `/path/to/application`, written unquoted in the prose, so dropping the quotes in the code block is a natural edit for someone touching the docs.

If that happens, the substitution silently stops matching, the command runs with the literal path `/path/to/application`, and the failure surfaces as:

    Marionette docs: No installed marionette found. Use the application workspace or --package-root for its physical package directory.

— which points at package resolution rather than at the doc edit that caused it. Stripping surrounding quotes before comparing would make this tolerant of either form.

skills/marionette/SKILL.md
```diff
@@ -11,11 +11,19 @@ choices. This skill does not authorize dependency upgrades.
...
 Resolve the package from the application workspace, not the copied skill directory
-or a neighboring monorepo package. The read-only helper requires Node 24 or later:
+or a neighboring monorepo package. The read-only helper requires Node 24 or later.
+Its path is `scripts/docs.mjs` relative to the directory containing this `SKILL.md`:
+
+- In the npm package, that directory is `<package-root>/dist/agent-skill/`.
+- After copying the skill, it is the copied directory, such as
+  `/path/to/application/.agents/skills/marionette/`.
+
+Run these commands from that skill directory, replacing `/path/to/application`
```
paulfalgout: **These now carry a working-directory precondition that the old form did not.**

The examples changed from an absolute script path to `node scripts/docs.mjs`, which resolves relative to cwd. An agent consuming this skill is usually working in the application root, not the skill directory — so copying the snippet verbatim from there fails with a module-not-found error. The previous absolute form worked from anywhere.

The text does say to run from the skill directory, so this is stated rather than hidden. But it puts a cwd requirement on a snippet whose entire purpose is to be copied and run, in a file read by agents that mostly cannot assume their cwd.

Keeping the command absolute while adding the new explanation of *where* that directory is would give the clarification this PR is after without the precondition — e.g. `node <package-root>/dist/agent-skill/scripts/docs.mjs`, with the two bullets above already telling the reader how to resolve that.

paulfalgout (commented):
The clarification is accurate — I confirmed `dist/agent-skill/scripts/docs.mjs` exists and that the two documented commands parse into exactly the two invocations the new check runs. Executing the documented commands in the fixture is a good addition. Three notes, none blocking.

test/fixtures/docs-package/validate.mjs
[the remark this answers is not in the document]
paulfalgout: Fixed in f84ce4d9. The fixture now requires distinct list and page modes, validates the list against the packaged manifest, and compares page output with the complete packaged agents page. Two list commands cannot satisfy the mode assertion. The installed docs-package fixture passes.

paulfalgout: Fixed in f84ce4d9. The page case parses the provenance header separately and compares all remaining output against the packaged docs/agents.md contents plus the printed newline. A missing or different page body now fails. The installed docs-package fixture passes.

test/fixtures/docs-package/validate.mjs
paulfalgout: Fixed in f84ce4d9. Matching surrounding single or double quotes are stripped from tokens before placeholder substitution, so an unquoted placeholder also resolves. This remains a narrow parser for these documented commands.

skills/marionette/SKILL.md
paulfalgout: Fixed in f84ce4d9. Both examples now use an absolute /path/to/skill-directory/scripts/docs.mjs placeholder, with the package and copied-skill locations explained above. The fixture executes the shipped commands from the application directory for both skill locations. npm run docs:check and the installed docs-package fixture pass.

PR #540, opened 2026-09-17
docs: clarify cancellation and active resource lifetimes

### LINKED ISSUE #538: v5: add executable contrasts for cancellation and active-resource lifetimes
Add concise, executable contrasts for the lifecycle mistakes identified by exploratory beta.4 probes, while reusing existing correct examples and checks. This is a documentation/test subtask of #140 and #147, not a report of a reproduced library bug.

## Existing evidence to preserve

At upstream `f4165f14198da115cc998c842fbf4b6a4886fe45`, [application-effects.md](https://github.com/marionettejs/marionette/blob/f4165f14198da115cc998c842fbf4b6a4886fe45/docs/application-effects.md) and its [installed fixture](https://github.com/marionettejs/marionette/blob/f4165f14198da115cc998c842fbf4b6a4886fe45/test/fixtures/docs-application-guides/effects.mjs) already cover rejected stop permission, successful timer cleanup, cancellation, stale results after a new start, and repeated restarts. Do not duplicate this work or claim those contracts lack tests.

Exploratory checks used published `marionette@5.0.0-beta.4`, jsdom 30.0.1, Node 24.19.0 and Jev 1.13.0. Sixteen new snippets (eight correct/buggy pairs) had public-API runtime observations fixed before model review. Model errors motivated this issue; they are not evidence of runtime defects or human usability failures. In a separate fixed-input diagnostic, Jev falsely flagged valid rejected-stop handling 10/10 times and missed an active timer surviving stop 10/10 times. Do not use a Jev accuracy threshold as acceptance.

## Scope and acceptance

- [ ] Audit the canonical examples and current checks first; link already-covered cases and add only missing explanatory contrasts or sequence assertions.
- [ ] Explain why a cancellation flag captured before an await becomes stale; check the relevant signal immediately before committing after the final asynchronous boundary, including a second await and a non-cooperative loader.
- [ ] Contrast an operation-specific cancellation check with `isRunning()` after a newer start: a currently running feature does not make an older result current.
- [ ] Explain object lifetime versus active lifetime: configured subscriptions survive stop; an `isRunning()` guard can suppress delivery after stop, but is not resource disposal and does not imply activity during pending stop permission. For effects that must remain active until stop permission succeeds, use the existing explicit scope pattern.
- [ ] Link rejected-stop preservation and timer cleanup to their existing executable examples; clarify that a preparation signal is not the signal for the subsequent active run.
- [ ] Explain that returning a shared external source from `createState()` assigns ownership; passing `state` borrows it. Cross-link the state guide and #191.
- [ ] Keep one canonical implementation of reusable example helpers. Teach `prepareStart` and synchronous notifications as documented; do not restore older lifecycle aliases.
- [ ] Make added examples executable against installed public packages. Assert observed state, lifecycle results, subscription/timer effects and identity; pair legitimate examples with narrowly broken controls where that proves discrimination.
- [ ] Check links, executable snippets, and the generated/packaged compact reference for drift.

Runtime cost: documentation and tests only. No core helper, private framework probes, automatic rollback, attempt-all cleanup, or new synchronous failure recovery. Any necessary production change needs its own public reproduction, proposed contract and cost review. Candidate resource APIs remain under #136, not prerequisites for this issue.

paulfalgout:
Startup cancellation can occur between loading and validation, and a newer running instance does not make an older result current. Add an executable preparation example that checks the originating signal after both awaits, alongside concise guidance on delivery gating versus resource disposal and borrowed versus factory-owned state.

The new installed-package fixture checks successful commit, canceled loading, and old validation completing after a newer successful start. Export the fixture with the docs and refresh the generated documentation hash. This changes documentation and tests only; the existing lifecycle and synchronous-failure contracts remain unchanged.

Fixes #538.

## Validation

- New fixture against published `marionette@5.0.0-beta.4`: all three scenarios passed.
- Removing either cancellation guard made its corresponding acceptance test fail; restored both and reran successfully.
- Focused ESLint passed with zero warnings.
- Executable-example validation: seven tests passed; 34 examples mapped to 17 validators.
- Docs build/link checks: 116 pages built; 117 HTML files and 1,772 links checked.
- Documentation export: six tests passed, including fixture dependencies.
- Public-contract drift and whitespace checks passed. Fixture and drift checks rerun after rebasing onto the current master.

The complete library and installed-guide suites were not rerun locally; hosted checks remain required.

## Advisory Jev comparison

A frozen 96-call before/after probe used `jev-1.13.0`, three passes over 16 previously seen runtime-labeled cases, counterbalanced condition order, and unchanged-page controls. On changed-guidance cases, bug detection increased from 11/18 to 16/18 responses, while false alarms increased from 5/18 to 6/18; correct responses increased from 24/36 to 28/36. Unchanged-input controls also varied.

This is exploratory regression evidence, not independent validation, a release gate, or a general model-performance claim. Executable checks and contract review establish the change's correctness.

---
## Summary by cubic
Clarifies cancellation and active resource lifetimes in the `application-effects` and `state` docs so startup cancellation between loading and validation is explicitly covered. A `prepareStart` signal covers only pending startup; an adopted `prepareStop` phase keeps its original options and context. Fixes #538. Docs and tests only; the existing lifecycle and synchronous-failure contracts are unchanged.

- Documents rechecking `signal.aborted` after each awaited step before committing a result, since a newer running instance does not make an older result current.
- Adds an executable preparation example and an installed fixture covering successful commit, canceled loading, and old validation completing after a newer successful start.
- Distinguishes delivery gating from resource disposal and clarifies that `createState()` ownership is set by its return value even when returning a shared source.
- Registers the new fixture with the docs and refreshes the generated documentation hash.

<sup>Written for commit e78fa8a72b5be510fbc665d73924886c4efb0871. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/540?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Clarified that state created by a factory is owned and disposed by that state, while shared sources should be passed explicitly when ownership remains with the caller.
  - Added guidance for rechecking cancellation after asynchronous work to prevent stale operations from committing results.
  - Documented the distinction between suppressing effect delivery and cleaning up resources during stopping.

- **Tests**
  - Added executable coverage for preparation, validation, cancellation, and protection against stale results from earlier operations.


docs/application-effects.md
```diff
@@ -27,6 +27,62 @@ not a signal for the entire subsequent run. The scope below owns that longer
 lifetime and passes its signal to the loader. Every awaited continuation checks
 cancellation before committing its result, including providers that ignore abort.
 
+## Recheck cancellation after awaited work
+
+A preparation signal belongs to one pending lifecycle operation. Read
```
paulfalgout: This rule is stated for preparation generally, but it does not hold for `prepareStop`: an adopted stop readiness keeps its original context and is never aborted, so its signal outlives the operation that created it.

Probe against published `marionette@5.0.0-beta.4` (prepareStop awaiting a gate; `stop()` then `destroy()`):

```
prepareStop calls: 1  aborted after destroy supersedes stop: false
stop resolved: false  destroy resolved: true  prepareStop calls total: 1
```

A reader applying this sentence to `prepareStop` would conclude that work was invalidated when `stop()` resolved `false`, and skip or discard a result the adopted `destroy` operation is still awaiting and will commit. Suggest scoping the sentence to startup preparation (or noting the adopted-stop exception, which `docs/compact-reference.md` already records as "adopted stop retains its original context/options without abort").

paulfalgout: Fixed in e78fa8a7. Scoped all three signal-lifetime statements in this guide to prepareStart and added the adopted prepareStop exception with a link to the lifecycle contract. Reproduced the stop-then-destroy probe against published 5.0.0-beta.4: one prepareStop call, un-aborted signal, stop=false, destroy=true. Startup fixture tests, executable-example validation, docs build, and internal links all pass.

PR #546, opened 2026-09-19
feat(application): bind the host region at startup

paulfalgout:
Stacked on #545; merge #545 first.

## What
- Accept an existing same-runtime Region through typed `start({ region })` and `restart({ region })`, binding it before startup hooks. Startup neither constructs Regions nor rewrites constructor configuration; `getRegion()` returns the active host.
- Preserve lifecycle identity rules: compatible pending operations share a Promise; plain start rejects a different active/pending host with `MN0041`; restart can supersede startup or a restart requesting a different host.
- Add public runtime/type contracts and browser scenarios for reusable children following recreated layout Regions and a borrowed shell surviving canceled loading and controller restart.

## Why
Parents know their current layout slots at activation time. Passing the Region at startup lets a registered child reuse its lifecycle while moving between newly created layout hosts. Constructor Region definitions remain available for apps that own a host; startup needs only borrowed instances.

## Scope
Application startup/restart host binding, public types, diagnostics, documentation, and validation registration. Removes startup construction, argument-shape diagnostics, and duplicated constructor-state assignments. Retains same-runtime validation and current-operation checks around teardown callbacks because a newer async operation must remain authoritative.

The `region` startup key is reserved; rename domain options such as region codes. There is no running-app `setRegion()` API. General synchronous teardown recovery and automatic activation based on Region contents are unchanged.

## Deployment Impact
Library change for the beta.5 candidate. No forms or bundles require redeployment from this PR alone; no package version, publication, or documentation deployment is performed. Consumer adoption requires passing Region instances rather than selector/definition forms at startup.

## Testing
At combined commit `f9cc02cc`:
- `npm run coverage` — 2,072 tests passed in 134 files.
- `npm run build` — passed, including consumer type checks.
- `npm run lint:ci`, `npm run docs:check`, `npm run check:api-contracts`, `npm run check:diagnostics`, and `npm run check:public-tests` — passed.
- `npm run test:browser -- test/browser/application-prepared-view.spec.mjs test/browser/application-start-region.spec.mjs test/browser/application-child-activation.spec.mjs` — 18 passed across Chromium, Firefox, and WebKit.
- Runtime cases include same-host joins, conflicting plain starts, restart supersession, canceled/failed readiness, owned-host disposal, cross-runtime rejection, shared-host cleanup, and late loading completion after cancellation. ESM/CommonJS type cases reject selector, definition, class, number, and boolean startup hosts.
- Full app-frontend migration and package release certification are outside this PR's validation.

Latest master documentation (#544) is integrated in the final heads: #545 `6e06bea5`, #546 `7c9c84d3`. Documentation and API-inventory checks passed again; source and runtime/browser tests are unchanged from the tested revisions above.


src/modules/application.ts
```diff
@@ -463,6 +478,43 @@ function beginOperation(application: ApplicationInternals, kind: OperationKind,
...
+  if (definition === current || definition === application.region) { return; }
+  const defaults = {
+    [runtimeId]: application[runtimeId],
+    regionClass: application.regionClass
+  };
+  const next = buildRegion(definition, defaults);
+
+  const displayed = releaseDisplayedView(application);
+  if (displayed && current?.currentView === displayed) {
+    current.empty();
```
paulfalgout, line 486: `current.empty()` runs user code (view `before:destroy`/`destroy` handlers), and the commit below it is not re-checked against the current operation — which this file otherwise does after every await (`// Only the current operation may commit or restore Application state`, line ~426).

Concrete corruption: app bound to owned `R1` displaying `V`; a handler on `V` calls `app.start({ region: R3 })`.

1. `app.start({ region: R2 })` → `replaceStartRegion` builds `R2`, releases `V`, calls `current.empty()`.
2. `V`'s handler re-enters `start({ region: R3 })`. State is `STOPPED`, so `assertStartRegion` is skipped; the new operation supersedes, builds `R3`, and sets `_region = _ownedRegion = R3` (and returns synchronously into `beginReadiness`).
3. Control returns to the outer frame, which calls `application._ownedRegion?.destroy()` — **destroying `R3`** — then assigns `_region = R2`, `region = R2`, `_ownedRegion = R2`.

The newer `start` resolves `true` while the Application is bound to `R2` and its requested host is destroyed. Capturing `current`/`_ownedRegion` in locals before `empty()` and bailing out when `isCurrentOperation` no longer holds would close it.

src/modules/application.ts
```diff
@@ -612,8 +672,12 @@ export default /* @__PURE__ */ ((methods: object) => {
       return Promise.resolve(false);
     }
 
+    const definition = getStartRegion(options);
     const operation = this._lifecycleOperation;
-    if (operation?.kind === 'restart') { return operation.promise; }
+    if (operation?.kind === 'restart') {
+      try { assertStartRegion(this, definition, operation); } catch (error) { return Promise.reject(error); }
```
paulfalgout: This makes `restart` the *stricter* path, which is the opposite of the documented guidance ("use `restart({ region })` … to change a running app's host").

With a restart in flight in its stop phase (state `RESTARTING`):
- `app.restart({ region: third })` → rejects `MN0041` (covered by `rejects a different Region during restart without changing the pending host`).
- `app.start({ region: third })` → passes the guard at line 622 (`kind` is `'restart'`, state is neither `STARTING` nor `RUNNING`), supersedes the restart, and binds `third` successfully.

Same state, same intent, opposite outcomes — and the rejecting one is the API the docs point at. Everywhere else in this lifecycle a newer incompatible operation supersedes and resolves the old one `false` rather than rejecting. Is there a reason a second `restart({ region })` should not simply supersede the first?

src/modules/application.ts
```diff
@@ -34,6 +34,10 @@ export interface ApplicationOptions {
   stateEvents?: Bindings | (() => Bindings);
   state?: unknown;
 }
+export interface ApplicationStartOptions {
```
paulfalgout, line 37: `ApplicationStartOptions` is exported on the public surface but no public signature references it — `start`/`restart` stay `(options?: unknown)`. So it buys consumers no checking (`app.start({ region: 42 })` and `app.start({ regoin: r })` both compile), and the new cases in `test/types/application.{cts,mts}` only assert assignability to an explicitly annotated local, so they would still pass if the parameter type were wrong. Either widen the `start`/`restart` parameter to `ApplicationStartOptions` (the index signature keeps arbitrary options assignable) or drop the export.

src/modules/application.ts
```diff
@@ -463,6 +478,43 @@ function beginOperation(application: ApplicationInternals, kind: OperationKind,
   return deferred.promise;
 }
 
+function getStartRegion(options: unknown) {
+  return (options as ApplicationStartOptions | undefined)?.region;
```
paulfalgout: This reserves the `region` key of the forwarded options object. The breaking-changes section notes it, but the collision is silent at the call site: an app passing a domain value today — `app.start({ region: 'us-east-1' })` for `prepareStart` — now has `buildRegion` construct a Region with `el: 'us-east-1'`, silently rebinding the host and failing later at show time (`MN0004`), or rejecting with `MN0041` if it is already running. Worth considering a distinctly named key (`startRegion`) or at least calling out the silent-rebind case in the upgrade guide alongside the `MN0041` case.

paulfalgout (commented):
Review of startup Region binding. Ordering is careful in the places that matter — `buildRegion` before any mutation so a bad definition leaves the old host intact, `releaseDisplayedView` before the swap so the `empty` subscription comes off the *old* Region, and the bind after adopted stop readiness. Four comments: one re-entrancy hole, one API-consistency question, and two smaller ones.

upgradeGuide.md
```diff
@@ -570,6 +570,32 @@ canceled teardown can retain partial progress. Child start/restart
...
+
+```javascript
+const ChildApplication = Application.extend({
+  prepareStart(options) {
+    return this.loadData(options.source);
+  }
+});
+
+parent.prepareStart = () => child.start({
+  region: parent.getRegion(),
```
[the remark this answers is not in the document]
paulfalgout: Updated in 64ca3e5d (including 0b45c39e). The upgrade example now creates a Layout View on each parent start and passes its named content Region to the same registered child. The browser contract verifies child identity and the new layout host across restart.

src/modules/application.ts
```diff
@@ -612,8 +672,12 @@ export default /* @__PURE__ */ ((methods: object) => {
       return Promise.resolve(false);
     }
 
+    const definition = getStartRegion(options);
     const operation = this._lifecycleOperation;
-    if (operation?.kind === 'restart') { return operation.promise; }
+    if (operation?.kind === 'restart') {
```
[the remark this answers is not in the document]
paulfalgout: Updated in 64ca3e5d (including 0b45c39e). The requested restriction is intentionally not applied: restart deactivates before binding its requested host, and may cancel an unfinished start. This is distinct from plain start, which rejects a different host while starting/running. The contract is now explicit and a public test verifies the old start resolves false, its signal aborts, and the new restart succeeds.

docs/marionette.application.md
```diff
@@ -187,8 +187,25 @@ also supersede its requested stop. Inspect the result and handle rejection; a
...
+methods and events. Its `region` property can bind the Application to a
+Region instance, selector, Region class, or Region definition before
+`before:start` and `prepareStart` run:
+
+```javascript
+const childRegion = layout.getRegion('content');
+await child.start({ region: childRegion, source: 'layout' });
+```
+
+An omitted or `undefined` `region` keeps the current host. Region instances are
```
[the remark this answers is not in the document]
paulfalgout: Updated in 64ca3e5d (including 0b45c39e). Documented the distinction explicitly: Region.empty clears displayed presentation, not the feature lifecycle or its separately prepared replacement. The parent/router coordinates feature stop when appropriate; emptying a host does not implicitly stop its borrowers.

test/unit/application-start-region.spec.js
```diff
@@ -0,0 +1,313 @@
...
+    const second = makeRegion();
+    const ready = Promise.withResolvers();
+    const application = makeApplication({ prepareStart: () => ready.promise });
+    const started = application.start({ region: first, source: 'first' });
+    const joined = application.start({ region: first, source: 'same-region' });
+
+    expect(joined).toBe(started);
+    await expect(application.start({ region: second })).rejects.toMatchObject({ code: 'MN0041' });
+    ready.resolve();
+    await started;
```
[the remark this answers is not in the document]
paulfalgout: Updated in 64ca3e5d (including 0b45c39e). The join test now asserts the shared start Promise resolves true, in addition to Promise identity and conflicting-host rejection.

src/modules/application.ts
paulfalgout: Updated in 64ca3e5d (including 0b45c39e). Host replacement now captures the old owned host and checks that the operation remains current after teardown callbacks before disposing/binding further state. Public tests cover a newer restart during View or Region teardown and destruction during teardown. The exact nested plain-start scenario is already rejected with MN0041; a regression makes that explicit. Startup now accepts only borrowed instances, so temporary host construction/disposal is removed.

src/modules/application.ts
paulfalgout: Updated in 64ca3e5d (including 0b45c39e). A newer restart requesting a different host now supersedes the pending restart and adopts its stop readiness; the older operation resolves false and only the latest host is bound. Compatible pending restarts still share the same Promise.

src/modules/application.ts
paulfalgout: Updated in 64ca3e5d (including 0b45c39e). Both public and implementation start/restart signatures now accept ApplicationStartOptions. Its region accepts only RegionInstance. ESM and CommonJS direct-call type tests reject numbers, booleans, selectors, definitions, and Region classes.

src/modules/application.ts
paulfalgout: Updated in 64ca3e5d (including 0b45c39e). The upgrade guide explicitly reserves region and shows renaming domain values to regionCode. Following the maintainer decision, start/restart now accept only an existing Region instance, eliminating silent string-to-selector rebinding. Constructor options retain Region definitions.

src/modules/application.ts
```diff
@@ -450,6 +465,34 @@ function beginOperation(application: ApplicationInternals, kind: OperationKind,
...
+
+function replaceStartRegion(application: ApplicationInternals, operation: Operation, region: RegionInstance) {
+  const current = application._region;
+  if (region === current) { return; }
+  if ((region as RegionInternals)[runtimeId] !== application[runtimeId]) {
+    throw new MarionetteError({
+      code: 'MN0030',
+      name: 'RegionError',
+      message: 'A Region instance must belong to the same Marionette runtime as its owner.'
+    });
```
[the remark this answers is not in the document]
paulfalgout: Intentionally not changed, per the maintainer decision. start/restart accept an existing same-runtime Region instance; selectors, definitions, classes and null are unsupported startup inputs. Their diagnostics are not a supported contract. The runtime marker check enforces runtime isolation for valid Region instances, while public types reject unsupported forms. Restoring shape checks would reverse the explicitly approved simplification.

test/types/application.cts
```diff
@@ -68,6 +68,20 @@ const count: number = stateOwner.getState().count;
 stateOwner.getState().ready;
 
 async function lifecycle() {
+  const dynamicRegionStart: ApplicationStartOptions = { region: borrowedRegion, source: 'dynamic-region' };
+  const dynamicRegionStarted: Promise<boolean> = root.start(dynamicRegionStart);
```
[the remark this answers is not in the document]
paulfalgout: Intentional compile-only type assertion: assigning root.start(dynamicRegionStart) to Promise<boolean> verifies the public startup return contract. This fixture is compiled, not executed; the value need not be consumed. Keeping the assertion, which passes the public type build.

test/types/application.mts
```diff
@@ -68,6 +68,20 @@ const count: number = stateOwner.getState().count;
 stateOwner.getState().ready;
 
 async function lifecycle() {
+  const dynamicRegionStart: ApplicationStartOptions = { region: borrowedRegion, source: 'dynamic-region' };
+  const dynamicRegionStarted: Promise<boolean> = root.start(dynamicRegionStart);
```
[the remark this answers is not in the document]
paulfalgout: Intentional compile-only type assertion: assigning root.start(dynamicRegionStart) to Promise<boolean> verifies the public startup return contract. This fixture is compiled, not executed; the value need not be consumed. Keeping the assertion, which passes the public type build.

PR #548, opened 2026-09-19
feat(application): declare static child applications

paulfalgout:
## What
- Add `childApps: { editor: EditorApplication }` or `childApps()` returning that map to Application declarations and constructor options. Each parent constructs fresh children without arguments and registers them before `initialize()` through `addChildApp()`.
- Inherit the declaration independently of `initialize()` overrides; subclass and option maps replace the inherited map. Start, restart, remove, and destroy retain their existing child lifecycle behavior.
- Add public runtime/type tests, lifecycle documentation, upgrade guidance, and contract inventory evidence.

## Why
Static child registration in `initialize()` is repetitive and can be skipped when a subclass overrides initialization. Representative app-frontend router and detail controllers already express fixed child constructor maps. Native support removes the need for a consumer-side registration abstraction without adding another lifecycle mechanism.

## Scope
Application construction only: a constructor-time declaration resolution and loop using the existing getOption/addChildApp helpers. An undeclared Application adds no registry, listener, or child allocation. A declared map is read once, enumerated once, and each child uses existing ownership registration. Constructors receive no parent options. Declaration does not activate children, construct Regions, or reprocess the map on restart.

Per-child factories, shared instances, per-child option descriptors, lazy construction, implicit inheritance merging, and synchronous construction rollback are not added. Dynamic children and required constructor arguments use explicit `addChildApp()`. Native JavaScript classes can use a prototype method or getter; TypeScript native classes use a getter or `preinitialize()` because instance fields are assigned after `super()`.

Targets master after #545 and #546 merged. The declaration change preserves their startup/presentation contracts and does not update app-frontend dependencies.

## Deployment Impact
No forms or bundles require redeployment from this library PR alone. Consumers gain the API after adopting a package containing this change through their normal app deployment. No package publication, version bump, or documentation deployment is performed by this PR.

## Testing
Combined validation at `32ab51fa2`; runtime source and tests are identical at final PR head `dfe31e320`:
- `npm run coverage`: 2,081 tests passed and coverage thresholds passed.
- `npm run build`: passed, including ESM/CommonJS public consumer type checks.
- `npm run lint:ci`, `npm run docs:check`, `npm run check:api-contracts`, `npm run check:public-tests`, and `git diff --check`: passed.
- `npm run test:browser -- test/browser/application-prepared-view.spec.mjs test/browser/application-start-region.spec.mjs test/browser/application-child-activation.spec.mjs`: 18 checks passed across Chromium, Firefox, and WebKit.
- Cases cover per-parent identities, initialization order, option precedence, inherited-map replacement, explicit activation, restart identity, removal, native getters, nested destruction, and runtime ownership rejection.
- The subsequent Region documentation clarification passed docs and API contract checks. No app-frontend E2E was run as part of this integration.

---
## Summary by cubic
Adds `childApps` declarations to `Application`, so fixed children no longer need manual registration in `initialize()` and remain registered when subclasses override it. Maps or one-time parent-bound functions now construct fresh no-argument children before `initialize()`; declarations register children but do not start them.

- Subclass and constructor-option declarations replace inherited maps; non-undefined options override getters without reading or assigning them, and `{}` omits inherited children.
- Parent restart retains child instances without reconstructing or activating them; removing a declared child destroys it and does not recreate it.
- Child constructors receive no parent options, and declared lookups retain the general optional `ApplicationInstance` type.
- Documents initialization and name timing, native class declaration forms, migration guidance, and lifecycle boundaries.
- Adds runtime, type, and API contract coverage; coverage, build/type, lint, docs, contract, public-test, and browser checks pass.

<sup>Written for commit dfe31e3202cdbefa8bf97f68bbda9312bf4630df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marionettejs/marionette/pull/548?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Applications can declare named child applications using a constructor map or a function returning one.
  - Children are created and registered before initialization, support nesting and inheritance, and remain available across parent restarts.
  - Child applications require explicit activation and are cleaned up with their parent.

- **Documentation**
  - Updated lifecycle references, migration guidance, and the upgrade guide with declaration and activation behavior.

- **Tests**
  - Added runtime and type coverage for declarations, precedence, nesting, lifecycle behavior, and invalid configurations.


paulfalgout (commented):
Nine runtime lines for a feature with this much surface is the right shape, and reusing `addChildApp` for the ownership rules keeps the contract single-sourced. Five comments: one silent-failure inconsistency, one construction leak the tests work around, and three documentation/typing gaps.

src/modules/application.ts
```diff
@@ -179,6 +182,12 @@ const Application = function(this: ApplicationInternals, options?: ApplicationOp
   this._initRegion();
   this._initRadio();
   this._initState(options);
+  const childApps = this.childApps;
```
paulfalgout: `childApps` is the only declarative map in the framework read without `getValue`. Every comparable hash resolves a function form:

| hash | resolution |
| --- | --- |
| `regions` | `getValue(this, 'regions')` — src/modules/view.ts:286 |
| `ui` | `getValue(this, 'ui')` — src/mixins/ui.ts:98 |
| `events` / `triggers` | src/mixins/view-events.ts:83,95 |
| `channelName` / `radioEvents` / `radioRequests` | src/mixins/radio.ts:25,33,36 |
| `stateEvents` | src/mixins/state.ts:44 |

So this fails silently rather than loudly:

```js
Application.extend({
  childApps() { return { editor: EditorApplication }; }   // registers nothing
});
```

`if (childApps)` is truthy for the function, `Object.entries(fn)` is `[]`, the loop body never runs, and the parent constructs with no children and no error — surfacing only at the first `getChildApp('editor')`. Given the sibling APIs and Toolkit's own `childApps`, a function is the form people will reach for first.

`getValue(this, 'childApps')` would also be the natural answer to the native-class problem the docs currently work around with a prototype getter. If a function form is deliberately excluded, the loop should reject a non-object instead of no-oping.

src/modules/application.ts
```diff
@@ -179,6 +182,12 @@ const Application = function(this: ApplicationInternals, options?: ApplicationOp
...
+  if (childApps) {
+    for (const [name, ChildApp] of Object.entries(childApps)) {
+      this.addChildApp(name, new ChildApp());
```
paulfalgout, line 195: `new ChildApp()` completes before `addChildApp` can reject the registration, so a rejected child is fully constructed — Radio channel, Region, State wiring — and then dropped with nothing holding a reference to destroy it.

The new test acknowledges exactly this: the MN0031 case has to register the orphan for cleanup *from inside the child itself*.

```js
const Foreign = Application.extend({ initialize() { own(this); } });
expect(() => new runtime.Application({ childApps: { foreign: Foreign } }))
  .toThrow(expect.objectContaining({ code: 'MN0031' }));
```

If a test needs that hook to avoid leaking, consumers will leak. This one is cheap to close without adding the general rollback the PR scopes out — destroy the just-constructed child before the error propagates:

```js
for (const [name, ChildApp] of Object.entries(childApps)) {
  const child = new ChildApp();
  try { this.addChildApp(name, child); }
  catch (error) { child.destroy(); throw error; }
}
```

Earlier siblings registered before a later entry throws are the broader partial-construction case the PR body already scopes out; the orphan is different, because the framework constructed it in the statement that failed.

docs/marionette.application.md
```diff
@@ -258,6 +258,47 @@ Application locates and controls its children, while children receive required
...
+    editor: EditorApplication
+  },
+  initialize() {
+    // Both children are already registered, but neither has been started.
+    this.getChildApp('search');
+  }
+});
+```
+
+Each parent instance constructs fresh children once, after Region, Radio, and
```
paulfalgout, line 311: One ordering constraint missing here: a declared child's own `initialize()` runs inside `new ChildApp()`, which is *before* `addChildApp` assigns `_parentApp` and `_name`.

```js
const Child = Application.extend({
  initialize() { this.getName(); }   // undefined — not registered yet
});
Application.extend({ childApps: { editor: Child } });
```

The spec covers the parent side (`getChildApp('editor').getName()` from the parent's `initialize`), which reads as though the name is available throughout construction. Worth a sentence stating that a declared child is not yet owned during its own `initialize`, so name- or owner-dependent setup belongs in a start hook.

docs/marionette.application.md
```diff
@@ -258,6 +258,47 @@ Application locates and controls its children, while children receive required
...
+Children must belong to the parent's Marionette runtime.
+
+The declaration is inherited even when a subclass overrides `initialize()`.
+A subclass declaration or constructor `childApps` option replaces the entire
+inherited map; maps are not merged. Use `{}` to omit inherited children.
+`preinitialize()` may configure the declaration. With native classes, use a
+prototype getter or `preinitialize()`, not an instance field assigned after
+`super()` has completed construction.
+
+Declaration only constructs and registers. Start chosen children explicitly with
```
paulfalgout, line 337: `docs/migration-from-v4.md` line 87 was not updated and now contradicts this section. Its Application ownership row still reads "Keep Toolkit's established `addChildApp` … but **construct the child explicitly**" and "Remove `AppClass` configs" — which is what a declarative class map is, narrowed to no-argument constructors. A reader arriving from the migration table concludes the form is unsupported. Both documents are digested by the api-contracts inventory, so the published contract currently disagrees with itself.

test/types/application.mts
```diff
@@ -160,3 +160,20 @@ replacementPreparation.onStart?.(replacementPreparation, undefined, {name: 'Edit
 declare const notifications: ApplicationInstance;
 // @ts-expect-error Before notifications do not receive the preparation context.
 notifications.onBeforeStart?.(notifications, undefined, {signal: new AbortController().signal});
+
+const StaticChild = Application.extend({ label() { return 'static'; } });
+const declaredParent = new Application({ childApps: { editor: StaticChild } });
+declaredParent.getChildApp('editor') satisfies ApplicationInstance<object, unknown> | undefined;
```
paulfalgout, line 180: This asserts the weak type, which is the honest state of things, but it also shows the declaration's type information being thrown away:

```ts
const Workspace = Application.extend({ childApps: { editor: EditorApplication } });
new Workspace().getChildApp('editor').openDocument();  // does not type-check
```

The map is statically known, so keying `getChildApp` on `keyof this['childApps']` with `InstanceType<…>` would deliver the main ergonomic payoff of declaring children statically rather than calling `addChildApp`. If that is deliberately deferred, a line in the docs section saying declared children are still read as `ApplicationInstance` would set expectations.

src/modules/application.ts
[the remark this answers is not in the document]
paulfalgout: Intentionally not changed, following the maintainer decision and the documented synchronous failure contract. Constructor/registration failures abort without rollback; AGENTS.md explicitly excludes constructor and partial-registration recovery. An unawaited Application.destroy() cannot guarantee cleanup before a synchronous throw and can itself reject. The declaration remains shorthand for construction followed by addChildApp.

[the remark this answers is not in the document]
paulfalgout: Intentionally not changed, following the maintainer decision. Recursively constructing new instances is invalid recursive configuration, distinct from an ownership cycle between existing instances checked by MN0031. We are not adding construction-chain state or a new guaranteed diagnostic for this case.

src/modules/application.ts
paulfalgout: Fixed in 15a745d3. childApps now resolves through getValue, with the parent as this, once before initialize. Public runtime and ESM/CommonJS type tests cover function maps. Per-child factories remain unsupported.

src/modules/application.ts
paulfalgout: Intentionally not changed, following the maintainer decision and the documented synchronous failure contract. Constructor/registration failures abort without rollback; AGENTS.md explicitly excludes constructor and partial-registration recovery. An unawaited Application.destroy() cannot guarantee cleanup before a synchronous throw and can itself reject. The declaration remains shorthand for construction followed by addChildApp.

docs/marionette.application.md
paulfalgout: Fixed in 15a745d3. Documented that child initialize precedes registration/name assignment, while parent initialize follows it. The public test now asserts getName() is undefined during child initialization.

docs/marionette.application.md
paulfalgout: Fixed in 15a745d3. The migration table now recommends constructor maps or functions returning maps for static children and explicit construction for dynamic children or constructor arguments; Toolkit AppClass descriptors remain distinct.

test/types/application.mts
paulfalgout: Documented in 15a745d3. getChildApp remains ApplicationInstance | undefined; key-specific inferred lookup is deferred from this focused declaration change. Removal remains possible, so lookup remains optional.

src/modules/application.ts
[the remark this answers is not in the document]
paulfalgout: Intentionally not changed, following the maintainer decision and the documented synchronous failure contract. Constructor/registration failures abort without rollback; AGENTS.md explicitly excludes constructor and partial-registration recovery. An unawaited Application.destroy() cannot guarantee cleanup before a synchronous throw and can itself reject. The declaration remains shorthand for construction followed by addChildApp.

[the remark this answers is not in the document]
paulfalgout: Intentionally not changed, following the maintainer decision. Recursively constructing new instances is invalid recursive configuration, distinct from an ownership cycle between existing instances checked by MN0031. We are not adding construction-chain state or a new guaranteed diagnostic for this case.

docs/compact-reference.md
```diff
@@ -21,7 +21,7 @@ A synchronous callback/adapter exception escapes; no successful result or rollba
...
+Construction, addChildApp, setView and showView(view) return their public instance; showView() returns the prepared or current View, or undefined; removeChildApp returns a Promise of the removed child or undefined; queries return the declared state or optional owner member. Construction, registration, root preparation/display and queries are synchronous; removeChildApp awaits child destruction before releasing ownership. childApps constructs and registers fresh no-argument children once before initialize; declarations do not activate children. Stop destroys a prepared View and empties the host. Destroy also tears down owned children and a constructed root Region; it preserves a borrowed Region. Direct destruction releases preparation; displayed Views follow Region cleanup and detachment. [Application Lifecycle](./marionette.application.md).
```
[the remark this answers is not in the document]
paulfalgout: Fixed in 15a745d3. The compact reference and semantic contract explicitly state that restart retains registered children without reconstructing or automatically reactivating them; selected children must be started explicitly.

src/modules/application.ts
```diff
@@ -179,6 +183,12 @@ const Application = function(this: ApplicationInternals, options?: ApplicationOp
   this._initRegion();
   this._initRadio();
   this._initState(options);
+  const childApps = getValue(this, 'childApps') as ChildApplications | undefined;
```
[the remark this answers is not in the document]
paulfalgout: Fixed in e58b81e5. childApps is no longer copied onto the instance. getOption selects the constructor option or declaration property, then function maps are invoked with the parent context. Public tests cover map/function/empty overrides without reading or assigning a getter, plus undefined fallback; ESM/CommonJS tests cover native getter overrides.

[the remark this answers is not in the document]
paulfalgout: Fixed in e58b81e5 through option resolution rather than property assignment. No backing state or shared option-handling changes. Runtime and consumer-type regressions cover the documented combination.

config/api-contracts/semantics.json
```diff
@@ -2341,10 +2341,10 @@
         "Application"
       ],
       "result": "Construction, addChildApp, setView and showView(view) return their public instance; showView() returns the prepared or current View, or undefined; removeChildApp returns a Promise of the removed child or undefined; queries return the declared state or optional owner member.",
-      "timing": "Construction, registration, root preparation/display and queries are synchronous; removeChildApp awaits child destruction before releasing ownership.",
+      "timing": "Construction, registration, root preparation/display and queries are synchronous; removeChildApp awaits child destruction before releasing ownership. childApps resolves a map or function once and registers fresh no-argument children before initialize; declarations do not activate children. Parent restart retains registered instances without reconstructing or automatically reactivating them; start selected children explicitly.",
```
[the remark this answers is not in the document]
paulfalgout: Fixed in e58b81e5: MN0031 is now explicitly associated with application-composition, and the inventory was regenerated and checked.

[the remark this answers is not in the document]
paulfalgout: The declaration lifecycle evidence was already present in application-composition, and child lifecycle evidence exists under its lifecycle contract. e58b81e5 also adds explicit anchors for function resolution and getter-option overrides, making those boundaries directly auditable.

src/modules/application.ts
```diff
@@ -179,6 +182,13 @@ const Application = function(this: ApplicationInternals, options?: ApplicationOp
   this._initRegion();
   this._initRadio();
   this._initState(options);
+  const declaration = this.getOption('childApps') as ApplicationOptions['childApps'];
+  const childApps = typeof declaration === 'function' ? declaration.call(this) : declaration;
+  if (childApps) {
+    for (const [name, ChildApp] of Object.entries(childApps)) {
+      this.addChildApp(name, new ChildApp());
```
paulfalgout, line 195: Still open from the previous pass, now with direct evidence. Running the branch source:

```
ORPHAN: {"constructed":true,"destroyed":false,"cid":"mna2"}
```

— `new runtime.Application({ childApps: { foreign: ForeignApp } })` throws MN0031, and the constructed `ForeignApp` (Radio channel, Region, State wiring) is left alive with nothing holding a reference to destroy it. `test/unit/application-child-declarations.spec.js:144` still carries the workaround that makes this visible:

```js
const Foreign = Application.extend({ initialize() { own(this); } });
```

The narrow fix does not require the rollback the PR scopes out — only the child constructed by the statement that failed:

```js
for (const [name, ChildApp] of Object.entries(childApps)) {
  const child = new ChildApp();
  try { this.addChildApp(name, child); }
  catch (error) { child.destroy(); throw error; }
}
```

For completeness, the wider partial-construction case behaves as the docs say (no rollback), but it is worth knowing what it looks like: with `{ a: A, b: B }` where `B`'s constructor throws, `A` ends up `{aBuilt: true, aDestroyed: false, aName: 'a'}` — constructed, *named and registered*, on a parent whose construction was abandoned. That matches "there is no partial-construction rollback", so no change requested there; the orphan above is the one the framework creates and drops in the same step.

src/modules/application.ts
```diff
@@ -179,6 +182,13 @@ const Application = function(this: ApplicationInternals, options?: ApplicationOp
   this._initRegion();
   this._initRadio();
   this._initState(options);
+  const declaration = this.getOption('childApps') as ApplicationOptions['childApps'];
+  const childApps = typeof declaration === 'function' ? declaration.call(this) : declaration;
```
paulfalgout, line 192: The resolved declaration is never shape-checked, so a malformed one fails deep in construction with no mention of `childApps`. Observed on the branch source:

| declaration | result |
| --- | --- |
| `childApps: EditorApplication` (constructor where the map belongs, `Application.extend` form) | `TypeError: undefined is not a function` |
| same, native `class` child | `TypeError: Class constructor NativeEditor cannot be invoked without 'new'` |
| `P.prototype.childApps = { self: P }` | `RangeError: Maximum call stack size exceeded` |

The first two now reach `declaration.call(this)` because `typeof` is `'function'` for a constructor as well as for a declaration function. TypeScript rejects all three, so this is a JavaScript-consumer concern — but it is the same audience the function form was added for.

The cycle case is the one I would at least document: `addChildApp` deliberately detects it (`assertChildAppCanRegister` → MN0031 "A child Application cannot be an ancestor of its owner"), and the declarative path silently loses that diagnostic to a stack overflow. A sentence in the declaration section saying cycles are not detected — or a shape check on the resolved value — would keep the two registration paths telling the same story.

docs/marionette.application.md
```diff
@@ -258,6 +258,60 @@ Application locates and controls its children, while children receive required
...
+
+The declaration is inherited even when a subclass overrides `initialize()`.
+A subclass declaration or constructor `childApps` option replaces the entire
+inherited map; maps are not merged. A non-undefined `childApps` option takes
+precedence through `getOption()` without assigning to the declaration property,
+including a getter-only property. Use `{}` to omit inherited children.
+`preinitialize()` may configure the declaration property or `this.options.childApps`;
+a non-undefined option takes precedence over the property. With native classes, use a
+prototype `childApps()` method, getter, or `preinitialize()`, not an instance field assigned after
+`super()` has completed construction. TypeScript native subclasses should use a
```
paulfalgout, line 332: The advice is right, the reason is not. Verified with `tsc` against the branch source: the getter compiles and the method form fails with

```
TS2425: Class 'ApplicationInstance<…>' defines instance member property 'childApps',
        but extended class 'MethodParent' defines it as instance member function.
```

That is property-vs-method declaration form, not the union — narrowing the type to `() => ChildApplications` would still be rejected. As written ("the public property type supports both map and function values, so TypeScript rejects overriding it with a method declaration") a reader could reasonably conclude the opposite.

The new type tests cover the getter and the `Application.extend({ childApps() {…} })` method form, but nothing pins the native-method rejection — a `// @ts-expect-error` case would lock the behavior the docs are prescribing around.

paulfalgout (commented):
Second pass over `15a745d3`/`e58b81e5`. The `getOption` + function-resolution rewrite is a better fix than the `getValue` I suggested — it covers the option form and sidesteps assigning to a getter-only property, and the new getter/option-precedence test is the right regression to have. The migration table, the child-`initialize` ordering note and the lookup-type caveat all landed.

I re-ran everything below against the branch source in a worktree, so these are observed, not inferred. One finding is unaddressed from the first pass; two are new.

src/modules/application.ts
paulfalgout: Reviewed the direct reproduction. The maintainer decision remains unchanged: synchronous registration failure has no cleanup/rollback guarantee, including a freshly constructed child that registration rejects. This is the partial-registration try/catch recovery specifically excluded by AGENTS.md, not an overlooked supported cleanup path. Application.destroy() is asynchronous and can reject or be canceled; calling it without awaiting cannot establish the proposed synchronous cleanup guarantee. Keeping the accepted failure boundary rather than adding a special recovery path.

src/modules/application.ts
paulfalgout: Documented in 89beea9fa: declarations must form a finite construction tree; recursive declarations can exhaust the stack and are not covered by the MN0031 existing-instance ownership-cycle diagnostic. Unsupported declaration shapes have no guaranteed diagnostic. No constructor-shape guards or construction-chain state added, per the agreed scope.

docs/marionette.application.md
paulfalgout: Fixed in 89beea9fa. The docs now identify property-versus-method declaration form as the reason for TS2425, rather than the union. Added negative native-method cases to both ESM and CommonJS consumer type tests; npm run test:types passed.

test/types/application.cts
```diff
@@ -139,3 +139,40 @@ replacementPreparation.onStart?.(replacementPreparation, undefined, {name: 'Edit
...
+}
+new NativeDeclaredParent();
+new FunctionDeclaredParent().getChildApp('editor') satisfies ApplicationInstance<object, unknown> | undefined;
+// @ts-expect-error Declaration functions return constructors, not instances.
+new Application({ childApps: () => ({ editor: new Application() }) });
+
+new NativeDeclaredParent({ childApps: { replacement: StaticChild } });
+new NativeDeclaredParent({ childApps: () => ({ replacement: StaticChild }) });
+
+class NativeMethodDeclaredParent extends Application {
```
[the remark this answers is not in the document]
paulfalgout: Intentional compile-time regression fixture. Declaring this native subclass exercises TS2425 at the annotated method; @ts-expect-error makes the consumer type check fail if the expected diagnostic disappears. The ESM/CommonJS fixtures are compiled, not executed, so the class does not need to be instantiated or referenced. Keeping the negative contract test unchanged.

test/types/application.mts
```diff
@@ -160,3 +160,40 @@ replacementPreparation.onStart?.(replacementPreparation, undefined, {name: 'Edit
...
+}
+new NativeDeclaredParent();
+new FunctionDeclaredParent().getChildApp('editor') satisfies ApplicationInstance<object, unknown> | undefined;
+// @ts-expect-error Declaration functions return constructors, not instances.
+new Application({ childApps: () => ({ editor: new Application() }) });
+
+new NativeDeclaredParent({ childApps: { replacement: StaticChild } });
+new NativeDeclaredParent({ childApps: () => ({ replacement: StaticChild }) });
+
+class NativeMethodDeclaredParent extends Application {
```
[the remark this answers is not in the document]
paulfalgout: Intentional compile-time regression fixture. Declaring this native subclass exercises TS2425 at the annotated method; @ts-expect-error makes the consumer type check fail if the expected diagnostic disappears. The ESM/CommonJS fixtures are compiled, not executed, so the class does not need to be instantiated or referenced. Keeping the negative contract test unchanged.

config/api-contracts/inventory.json
```diff
@@ -8071,7 +8074,7 @@
     "docs/dom.interactions.md": "81bd678c7d498e08c08c84cfc3bbe1d27451ef46ed521e67fff4f72450bdbadb",
     "docs/events.class.md": "faed5ac4f6eca3ec447ae7cd44b6efd8f039e935b38a4c191076d485bd57a35f",
     "docs/events.md": "8225ceda0d100b6bf466ab779cacc1001b7eec16b8dfd09d8c43d97924228e50",
-    "docs/marionette.application.md": "b9cf3c3f47425d8de305cfa43a686f8eca1cd178d7e1b6bdcb19aa0f1b4413c2",
+    "docs/marionette.application.md": "5df2ffd3b4f6d18ec944cea91ce4b4d311c350c773265e3eff932f007f3bd652",
```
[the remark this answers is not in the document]
paulfalgout: The maintainer explicitly approved this bounded API for beta.5. childApps is declaration shorthand for constructing a child and calling the existing addChildApp API. It adds no topology inspection, parent/root traversal, second ownership registry, or automatic activation. Existing ownership and lifecycle contracts remain authoritative. The proposed v6 deferral does not apply to this approved static-composition convenience; retaining the implementation.

